PR #23, opened 2020-09-14
Subscriptions

teocomi:
Adds subscriptions for streams, branches and commits


IntegrationTests/Api.cs
```diff
@@ -20,260 +20,228 @@
...
+      Assert.NotNull(res);
+    }
+
+
+    [Test, Order(0)]
+    public async Task StreamCreate()
+    {
+      var res = await myClient.StreamCreate(new StreamCreateInput
       {
-        // First user - will own the client and streams
```
didimitrie, line 47: uhm... why is this gone? or is it github wrecking havoc with the diff display?

IntegrationTests/Api.cs
```diff
@@ -20,260 +20,228 @@
...
+    private string branchName = "";
+    private string commitId = "";
+    private string objectId = "";
+
+    [OneTimeSetUp]
+    public void Setup()
+    {
+      testServer = new ServerInfo { url = "http://127.0.0.1:3000", name = "TestServer" };
+
+      firstUserAccount = Utils.SeedUser(testServer);
```
teocomi, line 44: @didimitrie seed user creation has been moved to a utils class, should have broken it down in multiple PRs, sorrry!

PR #26, opened 2020-09-29
MongoDB transport base

clairekuang:
Closes specklesystems/Core#12

### DIFF
--- a/Core/Core.csproj
+++ b/Core/Core.csproj
@@ -13,6 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="MongoDB.Driver" Version="2.11.2" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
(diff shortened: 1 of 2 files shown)


didimitrie:
Hi @clairekuang! Thanks for this PR. One thing that I'm worried about is that I see no dependencies to the [mongodb driver](https://www.nuget.org/packages/mongocsharpdriver) added to the csproj file, maybe that was missed in the `git add` step?

clairekuang:
@didimitrie  forgot to commit .csproj updates! The mongo package reference should be included now

didimitrie:
Just passing by to say that we're not merging this *yet* as we've uncovered a little structural problem of having all the transports inside Core proper: we'll end up having a wee bit too many dependencies! 

We're thinking of a way to do organise this to be future proof, and hopefully we'll be able to proceed! (PS: i've started tracking this issue in #36).

izzylys:
Thanks a lot for this PR Claire ✨

As Dimitrie mentioned, we've decided to move transports out of Core to prevent bloating of Core itself as more transports are added over time. We've just merged this change (along with your MongoDB transport) in #38 

Closing this PR as the changes have already been merged into master 🚀

Issue #35, opened 2020-10-05
Add cancellation logic


didimitrie:
### Expected vs. Actual Behavior

Currently there's no way to exit form a `Send` or `Receive` operation before it's done. This is, obviously, no good! 

### Proposed Solution (if any)

Pass in an `CancellationToken` and honour the cancellation requests throughout the codebase. It can be optional to prevent client work disruption.

#### Optional: Affected Projects

It might affect existing 2.0 client work being done in Gh, Revit, Dynamo and Rhino. But not dramatically 😅

didimitrie:
closed in #49

PR #49, opened 2020-10-18
WIP: Dim/dev transport cancellation

### LINKED ISSUE #4: (Remote) Transport Error States
Remote transports should handle error states gracefully. Perhaps we need to introduce an onError event? 
PS: throwing errors from inside the `ConsumeQueue` function does not help, as it can't be caught anywhere.

didimitrie:
Changes in a nutshell: 
- transports can be cancelled (re #35)
- send can be cancelled (re #35)
- transports can bubble up errors (closes #4)
- gql api calls can be cancelled

- [x] receive should be cancellable (untested, needs connector work)
- [x] serialize should be cancellable (untested, needs connector work)
- [x] deserialize should be cancellable (untested, needs connector work)
- adds StreamWrapper class for ease of use of (stream, account) structs in dynamo/gh

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,18 @@
 version: 2.1
 jobs:
+
+  # The unit tests. 
+
   unittest:
     working_directory: /temp
+    
     docker: 
       - image: mcr.microsoft.com/dotnet/core/sdk:3.1-focal
+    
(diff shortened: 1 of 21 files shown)


didimitrie:
Okay I think this PR is done 😪

PR #98, opened 2020-11-04
👌 Show user if there are no local accounts

izzylys:
This adds an overlay to the DesktopUI home page if there are no local accounts on the machine. This is likely caused by not having installed manager, so the overlay directs the user to download the manager and try again.

no accounts found overlay

This includes a small change in core: `AccountManager` was using `CaptureAndThrow` when no default account was found. I assume the intended behaviour was to `CaptureException` but allow it to continue and return null, so I have made this change.

### DIFF
--- a/Core/Core/Credentials/AccountManager.cs
+++ b/Core/Core/Credentials/AccountManager.cs
@@ -106,7 +106,7 @@ namespace Speckle.Core.Credentials
         var firstAccount = GetAccounts().FirstOrDefault();
         if (firstAccount == null)
         {
-          Log.CaptureAndThrow(new SpeckleException("No Speckle accounts found. Visit the Speckle web app to create one."), level: Sentry.Protocol.SentryLevel.Info);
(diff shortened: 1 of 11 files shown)


Core/Core/Credentials/AccountManager.cs
```diff
@@ -106,7 +106,7 @@ public static Account GetDefaultAccount()
         var firstAccount = GetAccounts().FirstOrDefault();
         if (firstAccount == null)
         {
-          Log.CaptureAndThrow(new SpeckleException("No Speckle accounts found. Visit the Speckle web app to create one."), level: Sentry.Protocol.SentryLevel.Info);
+          Log.CaptureException(new SpeckleException("No Speckle accounts found. Visit the Speckle web app to create one."), level: Sentry.Protocol.SentryLevel.Info);
```
izzylys, line 109: @didimitrie tests pass, but just to confirm this is ok and that `GetDefaultAccount()` returning null doesn't break anything

izzylys:
it works!
reload-works

PR #100, opened 2020-11-05
fix(dynamo): removes objects.dll reference

didimitrie:
this was throwing off the GetConverterForApp func as it was looking inside the plugin folder rather than the kits folder.


izzylys:
ayyy PR#100 let's gooo 🎉

didimitrie:
@izzylys thanks for spotting! I was too busy thinking of how to punish @teocomi for adding this reference there... I'll go celebrate with... a `covfefe` :D

PR #123, opened 2020-12-01
WIP: Revit/sending receiving converting

didimitrie:
## Revit conversions: 
- moved conversion logic to/from revit from bulk list to per item
- receiving: previously received objects, that are not part of the current "bake", are removed (e.g., send 10 columns, then only 5). 
- Host elements and children (**tested only w/h walls**): 
  - when sending, hosts convert their children IF the children are part of the selection/filter. 
  - when receiving, host elements convert and bake their children
- Simplified inheritance chains and removed interfaces that were a bit superfluous 
- Moved revit classes next to the "clean" classes for easier cross-referencing between the two "paradigms" 
- Simplified `AddCommonRevitProps`

## Schema builder:
- make smart application Ids per component✅
- prevent seed refresh on file open/close✅
- refresh seed on copy paste ✅
- faster casting 

## Other changes in core: 
- `units` are now part of the `Base` object class, to simplify access. 
- cleaned up HasMemberSafe etc. methods for dynamic props; now going forward use runtime casting with as/is, e.g. `myObject["test"] as YourType` (as the `[]` operator returns null on undefined key names rather than throwing errors).


didimitrie:
Tests that do not pass: 

image

The rest is green, besides breps (i didn't dare to touch them, as I know alan's working on it on a diff branch). Will merge now, so we can unblock things.

PR #150, opened 2020-12-17
🌀 Refresh stream cards when the stream is modified elsewhere

### LINKED ISSUE #124: Stream Refreshes
When a stream gets updates while it's not active (either a different document is open in revit, or the ui is closed bc host app is closed), we should update it after it's deserialised and refresh the branches/commits/etc. 

PS: This might've been something I borked during the last round of changes on it; not entirely sure; apols if that's the case 😬

izzylys:
This resolves the issue of streams not being refreshed when a document with a saved stream is reopened and the stream is deserialised. The stream is now refreshed upon initialisation after being deserialised.

An issue regarding deserialisation of selection filters has also been resolved with a json converter.

This also resolves other issues with the branches / commits:
- commits and branches now update consistently including when they have been edited from the web, from another connector, or from the document itself (previously branches and commits would not update if you were modifying the stream from the document itself - they would only update if the stream was modified elsewhere)
- a bug with the commit list compiling commits from different branches has been squashed
- branch subscriptions are now handled so the branch / commit lists will update consistently 

This closes #124

### DIFF
--- a/DesktopUI/DesktopUI/DesktopUI.csproj
+++ b/DesktopUI/DesktopUI/DesktopUI.csproj
@@ -115,6 +115,7 @@
     <Compile Include="Utils\Helpers.cs" />
     <Compile Include="Utils\StreamDialogBase.cs" />
     <Compile Include="Utils\StreamState.cs" />
+    <Compile Include="Utils\StreamState.Events.cs" />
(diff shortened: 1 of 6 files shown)


didimitrie:
**This** looks really good. Moreover, **this** PR should be approved and merged at **this** user's (@izzylys) discretion.

PR #165, opened 2021-01-04
Integration tests

didimitrie:
theoretically updates to respect new ~~server~~ speckle-web repo structure


didimitrie:
@AlanRynne seems like the PR build ain't triggering, as you said during standup :D I'll merge this for now to see if it actually works 🤞

Issue #246, opened 2021-02-18
Acad: show warning when element conversion fails or is not supported


teocomi:
We should show warnings when element conversion fails or is not supported, in the same way the Revit one does it, via the expandable warnings panel.

PR #248, opened 2021-02-18
fix(autocad): failed conversion handling

### LINKED ISSUE #245: Acad: circle to native fails
Sending a circle from Rhino to Acutocad fails and crashes Autocad. We should:

- convert the circle properly
- catch and handle exceptions in the conversion properly

Sample stream: https://staging.speckle.dev/streams/a06dd8e92e/commits/4355e15c74

clairekuang:
-converter using circle.plane.origin instead of circle.center
-deprecated objects circle.center and ellipse.center properties
-added try catch for failed conversions
-added try catch for add-in initialize()

closes #245 #246

### DIFF
--- a/ConnectorAutocadCivil/ConnectorAutocadCivil/Entry/App.cs
+++ b/ConnectorAutocadCivil/ConnectorAutocadCivil/Entry/App.cs
@@ -4,6 +4,7 @@ using System.Reflection;
 using System.IO;
 using System.Windows.Media.Imaging;
 using System.Linq;
+using Forms = System.Windows.Forms;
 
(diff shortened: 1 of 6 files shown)


teocomi (changes requested):
I'm still having issues sending basic geometry from Rhino. It seems that as an error is encountered the entire conversion stops, while it should keep converting the remaining elements.

With the "acad test" stream I'm getting: 

```
Recieving Objects.Other.RenderMaterial objects is not supported. Object 08829e9f38ee2e25f9776b3473b990e6 not baked.
Failed to convert object f55771d7a7928503ed385c6576435a49 of type Objects.Geometry.Circle.

```

PR #264, opened 2021-02-23
GH: Misc fixes in component behaviour

AlanRynne:
## Description

First of all, sorry for the miscellaneous PR... I started pulling a thread and it seemed to have no ending! This PR fixes a collection of small issues I discovered today, mostly related to `Stream` handling in Grasshopper.

- Added `Equals` method to `StreamWrapper` to unify equality checks across components.
- Modified `ToString` method in `StreamWrapper` to reflect branch/commit/object urls.
- `CreateStream` now assigning the new `account.userInfo.id` instead of `account.id` (old format).
- `StreamGet` now passes along the `Commit/Branch/ObjectId` when assigning a new account.
- The `ErrorAction` in both Sender and Receiver was borked.
  - Added a missing `Done()` call before cancelling any other sources except the current one.
  - Added a quick hack to show an appropriate message when dealing with the known issue with `link sharing` server side. Included a `TODO` comment to remove them once the issue is fixed.
- Modified `HandleInputType` in receiver to use the new `StreamWrapper.Equals` method to determine if the client needs to be reset.
  - This change may fix an issue we encountered with the receiver, where changing from different branch url's in the same stream did not update properly, causing to receive the wrong data!!

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How has this been tested?

Please describe the tests that you ran to verify your changes. 

Tested all different type's of streams (stream, branch, commit, object) including:

- Accounts that had access as collaborator
- Accounts with no access but with link sharing on
- Accounts with no access and link sharing off
- Stream with wrong ID (non existent)
- Branch stream with wrong name (non existent)
- Object with wrong ID (non existent)

In all cases, all modified components seem to work as expected, reporting errors when any operation fails.


Core/Core/Credentials/StreamWrapper.cs
```diff
@@ -234,8 +235,20 @@ public async Task<Account> GetAccount()
 
       throw err;
     }
-
-
+    
+    public bool Equals(StreamWrapper wrapper)
```
teocomi, line 239: Maybe unnecessary, but could sanitize some of these fields (eg to lowercase) to be 100% sure.

PR #278, opened 2021-03-01
🌋 Refactor exception handling and pass GraphQL Errors from Client Operations

### LINKED ISSUE #32: Pass GraphQL Errors From Client Operations
### Expected vs. Actual Behavior

At the moment, [GraphQL Client Operations](https://github.com/specklesystems/Core/blob/master/Core/Api/Client.GraphqlClientOperations.cs) all return standard string errors (eg "Could not create stream"). It would be more helpful if the specific GraphQL errors could be passed on in the message so you don't have to dig all the way into core to find out what the actual error is.

### Reproduction Steps & System Config (win, osx, web, etc.)

Make any bad GraphQL request using `Client.GraphqlClientOperations`.

### Proposed Solution (if any)

Pass the actual GraphQL errors into the `GraphQLException`.

izzylys:
## Description

Previously, exceptions were thrown by calling `CaptureAndThrow( new SpeckleException( ... ) )` or `CaptureAndThrow( new GraphQLException( ... ) )`. Additionally, GraphQL Errors were not passed on to the dev when a `GraphQLException` was thrown.

This refactor changes the behaviour so exceptions are captured on construction and should be thrown like any other exception.
```cs
throw new SpeckleException("Something went wrong 😢", log: true, level: SentryLevel.VerySad)
```

- closes #32 

--- 

## Key Changes

1. Exceptions are captured in the constructors. Default is `log = true` and default log level is `Info`:
```cs
    public SpeckleException(string message, bool log = true, SentryLevel level = SentryLevel.Info ) : base(message)
    {
      if (log)
        Log.CaptureException(this, level);
    }
```

2. the GraphQL errors are now passed along to the dev (via a `SpeckleException` - the `GraphQLException` has been removed):
```cs
    public SpeckleException(string message, GraphQLError[ ] errors, bool log = true,
      SentryLevel level = SentryLevel.Info) : base(message)
    {
      GraphQLErrors = errors.Select(error => new KeyValuePair<string, object>("error", error.Message)).ToList();
      if (log)
        Log.CaptureException(this, extra: GraphQLErrors);
    }
```

3. `TaskCancelledException`s are never captured
```cs
    public static void CaptureAndThrow(Exception e)
    {
      if ( !( e is TaskCanceledException ) )
        CaptureException(e);
      throw e;
    }
```

4. `AggregateException`s are unpacked and the inner exceptions are captured individually
unpacked aggregate exceptions in the sentry dashboard

5. `CaptureAndThrow` has been removed 
5. `SpeckleException` with `InnerException` will be ignored and not captured

```cs
    public SpeckleException(string message, Exception inner, bool log = true, SentryLevel level = SentryLevel.Info)
         : base(message, inner)
    {
      if (inner is SpeckleException)
        return;
      if (log)
        Log.CaptureException(this);
    }
```

### Additional Details

- GraphQL exceptions are no longer being double reported

- All generic `System.Exception`s that we were throwing have been converted to `SpeckleException`s and are being logged. We can revisit this and weed out / adjust log levels if we find this is too much and we don't care about some of these. 

- `ReflectionTypeLoadException`s when loading kits are ignored.

### Log Levels

Regarding log levels, I'm not too clear on the strategy here. At the moment, basically everything is info. The things I have elevated are:

1. `🔴 Error`: critical core stuff (exceptions in `Operations.Send`/`Receive` and the `BaseObjectSerialiser`) 
2. `🔴 Error`: unable to generate account id (`Account.id`)
3. `🟡 Warning`: inheriting from `SelectKitComponentBase` in grasshopper
4. `🟡 Warning`: failure to load kits

I'm not sure about a lot of other stuff tho. eg exceptions from GraphQL queries are handled properly everywhere they are used, so I have this as `Info`. However, if a dev is using it from Core and not handling possible exceptions then for them this is an `Error`. Should all GraphQL errors then be `Error` level? 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

This changes the behaviour of the `SpeckleException` but the rest of `speckle-sharp` has also been refactored to reflect this.

## How has this been tested?

triggered errors manually and checked sentry dashboard to ensure errors are being reported as expected

### DIFF
--- a/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
+++ b/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
@@ -33,7 +33,7 @@ namespace Speckle.Core.Api
         CommitCreatedSubscription = res.Subscribe(response =>
         {
           if (response.Errors != null)
-            Log.CaptureAndThrow(new GraphQLException("Could not subscribe to commitCreated"), response.Errors);
+            throw new GraphQLException("Could not subscribe to commitCreated", response.Errors);
 
           if (response.Data != null)
             OnCommitCreated?.Invoke(this, response.Data.commitCreated);
@@ -78,7 +78,7 @@ namespace Speckle.Core.Api
         CommitUpdatedSubscription = res.Subscribe(response =>
         {
           if (response.Errors != null)
-            Log.CaptureAndThrow(new GraphQLException("Could not subscribe to commitUpdated"), response.Errors);
+            throw new GraphQLException("Could not subscribe to commitUpdated", response.Errors);
 
           if (response.Data != null)
             OnCommitUpdated(this, response.Data.commitUpdated);
@@ -123,7 +123,7 @@ namespace Speckle.Core.Api
         CommitDeletedSubscription = res.Subscribe(response =>
         {
           if (response.Errors != null)
-            Log.CaptureAndThrow(new GraphQLException("Could not subscribe to commitDeleted"), response.Errors);
+            throw new GraphQLException("Could not subscribe to commitDeleted", response.Errors);
 
           if (response.Data != null)
             OnCommitDeleted(this, response.Data.commitDeleted);
--- a/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
+++ b/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
@@ -42,7 +42,7 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        Log.CaptureAndThrow(e);
+        throw new SpeckleException(e.Message, e);
       }
     }
 
@@ -86,7 +86,7 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        Log.CaptureAndThrow(e);
+        throw new SpeckleException(e.Message, e);
       }
     }
 
@@ -130,7 +130,7 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        Log.CaptureAndThrow(e);
+        throw new SpeckleException(e.Message, e);
       }
     }
 
--- a/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
+++ b/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
@@ -42,7 +42,8 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        Log.CaptureAndThrow(e);
+        if (!(e is SpeckleException))
+          Log.CaptureAndThrow(e);
       }
     }
 
@@ -86,7 +87,8 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        Log.CaptureAndThrow(e);
+        if (!(e is SpeckleException))
+          Log.CaptureAndThrow(e);
       }
     }
 
@@ -130,7 +132,8 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        Log.CaptureAndThrow(e);
+        if (!(e is SpeckleException))
+          Log.CaptureAndThrow(e);
       }
     }
 
--- a/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
+++ b/Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
@@ -33,7 +33,7 @@ namespace Speckle.Core.Api
         CommitCreatedSubscription = res.Subscribe(response =>
         {
           if (response.Errors != null)
-            throw new SpeckleException("Could not subscribe to commitCreated", response.Errors, log: true);
+            throw new SpeckleException("Could not subscribe to commitCreated", response.Errors);
 
           if (response.Data != null)
             OnCommitCreated?.Invoke(this, response.Data.commitCreated);
@@ -42,8 +42,7 @@ namespace Speckle.Core.Api
       }
       catch (Exception e)
       {
-        if (!(e is SpeckleException))
-          Log.CaptureAndThrow(e);
+        Log.CaptureAndThrow(e);
       }
     }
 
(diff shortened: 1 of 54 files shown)


izzylys:
- [ ] this terrible overlay that isn't changing for light mode (i'm sorryyyyy)
image
- [ ] button to remove streams you don't have a local account for / don't have access to

Core/Core/Logging/SpeckleException.cs
```diff
@@ -1,4 +1,9 @@
 ﻿using System;
```
didimitrie, line 1: What i was thunking here was that we don't really need two exception classes, and could just have one! Perhaps it's not worth it though. Here's my 0.002c re what I had in mind: 

```cs

public SpeckleException( string message, bool log = false) { ... } 

public SpeckleException( GraphQLError[ ] gqlErrors, bool log = false) { ... } 

```

Not sure the `GraphQLException` does anything special in the way we handle it, but IIRC, it doesn't?

didimitrie: Ofc, internally: 
```cs
public SpeckleException( string message, bool log = false) 
{ 
   // ...
   if(log) 
   { 
     Log.Capture() etc.   
   }
} 
```

I'm not sure if we catch globally unhandled exceptions in .NET?

teocomi: > I'm not sure if we catch globally unhandled exceptions in .NET?

They are usually swallowed by the host applications.

izzylys:
PR description is being updated as I go so have a read to get an idea of where this is at. suggestions welcome!

AlanRynne:
Regarding the `CaptureAndThrow` magic and the handling of `SpeckleExceptions`. My only concern with that is that we have no way of knowing, as a developer, if this exact exception has already been captured at a lower level.

I would propose the `CaptureAndThrow` behaviour to be:

1. Capture the original exception for Sentry, whatever that is, so that it is properly logged with its name 👍🏻 
2. Create a new **unlogged** `SpeckleException` to which we can assign the "original" exception as the `InnerException`.
3. Throw the `SpeckleException` instead.

This way, you could assume that if you encounter a `SpeckleException` you can already assume that it has been handled in Sentry (either logged or ignored...whatever), and we also have access to the original Exception further on if necessary.

teocomi:
Agree with what you describe @AlanRynne ! And I think @didimitrie 's suggestion was to have 1 & 2 in the constructor of the `SpeckleException`.

Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
```diff
@@ -42,8 +42,7 @@ public void SubscribeCommitCreated(string streamId)
       }
       catch (Exception e)
       {
-        Log.CaptureException(e);
-        throw e;
+        Log.CaptureAndThrow(e);
```
teocomi, line 45: Was this one missed? There's a few more below...

izzylys:
cool yes I've incorporated Alan's suggestion! see the notes section of the PR

can you clarify what you mean? 1. and 2. are the constructors

Core/Core/Logging/SpeckleException.cs
```diff
@@ -1,36 +1,37 @@
...
-  {
-    public GraphQLException()
+    public SpeckleException(string message, bool log = false, SentryLevel level = SentryLevel.Info ) : base(message)
     {
+      if (log)
+        Log.CaptureException(this, level);
     }
 
-    public GraphQLException(string message) : base(message)
+    public SpeckleException(string message, GraphQLError[ ] errors, bool log = false,
```
teocomi, line 22: Minor comment: maybe log could default to true, since that's the majority of the cases. But it's probably too much hassle to update it, so no biggie!

Core/Core/Logging/SpeckleException.cs
```diff
@@ -1,36 +1,37 @@
...
+      SentryLevel level = SentryLevel.Info) : base(message)
     {
+      GraphQLErrors = errors.Select(error => new KeyValuePair<string, object>("error", error.Message)).ToList();
+      if (log)
+        Log.CaptureException(this, extra: GraphQLErrors);
     }
 
-    public GraphQLException(string message, Exception inner)
+    public SpeckleException(string message, Exception inner, bool log = false, SentryLevel level = SentryLevel.Info)
```
teocomi, line 30: We should probably check if `inner is SpeckleException` and in that case ignore it as it's already been reported before.

Core/Core/Logging/SpeckleException.cs
didimitrie: i'd agree re too much hassle :D

Objects/Objects/ObjectsKit.cs
```diff
@@ -84,7 +84,7 @@ public ISpeckleConverter LoadConverter(string app)
         }
         else
         {
-          throw new SpeckleException($"Converter for {app} was not found in kit {basePath}");
+          throw new SpeckleException($"Converter for {app} was not found in kit {basePath}", level: Sentry.Protocol.SentryLevel.Warning);
```
teocomi, line 87: Should log be true?

didimitrie: yep agreed on this - if we end up here something seriously borked in the installation process...

Core/Core/Logging/SpeckleException.cs
teocomi: > cool yes I've incorporated Alan's suggestion! see the notes section of the PR
> 
> can you clarify what you mean? the code snippets in 1. and 2. are the constructors

Yup, soz I replied before looking at the code! I think the only bit missing then is https://github.com/specklesystems/speckle-sharp/pull/278#discussion_r586563243

didimitrie (commented):
beyond @teocomi 's last point, all good 2 go :)

Core/Core/Logging/SpeckleException.cs
izzylys: rip soz I assumed there was a reason for dim's suggestion in his comment to default `log = false` but I guess not 😅

Core/Core/Api/GraphQL/Client.Subscriptions.Commit.cs
izzylys: in what way? i'm using `CaptureAndThrow` here on purpose if that's what you mean

PR #284, opened 2021-03-03
fix(rhino): Model unit converter now checks for unit change.

### LINKED ISSUE #239: GH units not reset upon document change
Flagged by @AlanRynne. It seems the units being passed to the GH converter are not refreshed upon Rhino document change. 
The GH connector should always refresh the active document units before sending/receiving.

AlanRynne:
## Description

Grasshopper nodes were not properly updating the `ModelUnits` when the Document units in Rhino changed.

Added missing switch case for `km` and `kilometer` to be converted to Speckle Units.

Initially, @teocomi suggested to create a new `IConverter` every time we wanted to run a conversion, which would guarantee the proper units were set. This turned out to be a bigger change than intended, so I settled for a fast and simple check on the `ModelUnits` getter.

Previous behaviour was checking `_modelUnits` for `null` or `empty` values, we now also check if the saved value corresponds to the document's model units.

Added `Converter.Set` call on the `BeforeSolveInstance` method of `SelectKitComponentBase`, so that it would propagate to any components inheriting from it. This ensures correct units are set even when the Active RhinoDoc changes (new document is created or a different document is opened while grasshopper was open).

- Fixes #239 

## Type of change

## How has this been tested?

Please describe the tests that you ran to verify your changes. 

Tested with some basic geometric objects (points, curves, simple surfaces) and serialized the result to check the set units corresponded to the newly changed units in Rhino.

Also tested changing to "unsupported" units.

### DIFF
--- a/Objects/Converters/ConverterRhinoGh/ConverterRhinoGh/ConverterRhinoGh.Units.cs
+++ b/Objects/Converters/ConverterRhinoGh/ConverterRhinoGh/ConverterRhinoGh.Units.cs
@@ -13,13 +13,15 @@ namespace Objects.Converter.RhinoGh
   public partial class ConverterRhinoGh
   {
     private string _modelUnits;
+    
     public string ModelUnits
     {
       get
       {
-        if (string.IsNullOrEmpty(_modelUnits))
+        var unitToSpeckle = UnitToSpeckle(Doc.ModelUnitSystem);
+        if (string.IsNullOrEmpty(_modelUnits) || unitToSpeckle != _modelUnits)
         {
-          _modelUnits = UnitToSpeckle(Doc.ModelUnitSystem);
+          _modelUnits = unitToSpeckle;
         }
         return _modelUnits;
       }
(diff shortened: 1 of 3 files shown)


Objects/Converters/ConverterRhinoGh/ConverterRhinoGh/ConverterRhinoGh.Units.cs
```diff
@@ -13,13 +13,15 @@ namespace Objects.Converter.RhinoGh
   public partial class ConverterRhinoGh
   {
     private string _modelUnits;
+    
     public string ModelUnits
     {
       get
       {
-        if (string.IsNullOrEmpty(_modelUnits))
+        var unitToSpeckle = UnitToSpeckle(Doc.ModelUnitSystem);
```
teocomi, line 21: Hey Alan,
I have two comments on this approach :
- we don't need to cache the _modelUnits every time, so it could be simplified
- we still need to make sure an up-to-date Doc is passed to the converters every time before converting

Issue #310, opened 2021-03-10
Revit: error sending columns


teocomi:
When sending columns during our workshop, 5 were sent fine but others didn't go through (some of these columns were copy-pasted). To be debugged.

![image](https://user-images.githubusercontent.com/2679513/110645776-2d588280-81ae-11eb-8349-216128234c82.png)

PR #317, opened 2021-03-10
fix(revit): resolve issues with column conversions

### LINKED ISSUE #309: Revit: Column conversion ToSpeckle failing
The unit test passes somehow, but sending the BeamsCols.rvt test file to Speckle produces conversion errors and one of the columns is missing.

commit: https://speckle.xyz/streams/1a9c89d03a/commits/6a6c4b5982
image

the file:
image

This failure is occurring on a column that doesn't have a baseline (`baseGeometry` is a Point), so creating one is attempted.

```cs
      //make line from point and height
      if (baseLine == null && baseGeometry is Point basePoint)
      {
        var topLevelParam = revitColumn.get_Parameter(BuiltInParameter.FAMILY_TOP_LEVEL_PARAM);
        var elevation = (double)((RevitLevel)ParameterToSpeckle(topLevelParam).value).elevation;
        baseLine = new Line(basePoint, new Point(basePoint.x, basePoint.y, elevation + speckleColumn.topOffset, ModelUnits), ModelUnits);
      }
```

This fails when getting the `elevation` because the `topLevelParam` storage type is `ElementId` (weird?) which is not a handled case in the `ParameterToSpeckle` switch statement. This means `.value` is null and we get a null ref exception and this conversion fails.

izzylys:
## Description

This was a super ez fix! Just some small changes to fix column conversion issues
- uncomments handling of case `ElementId` in `ParameterToSpeckle()`
- adjusts line creation logic for vertical columns with the "top level" at the bottom of the column

before:
borked columns

after:
fixed columns

[the commit](https://speckle.xyz/streams/4a8efa56a8/commits/176028dfb7) if you're curious

- Fixes #309 and #310 

## Type of change

## How has this been tested?

- [x] Unit/Integration Tests
- [x] Manual Tests 

Ran all xUnitRevit tests and tested manually using the revit connector

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -253,16 +253,15 @@ namespace Objects.Converter.Revit
           if (sp.value == null)
             sp.value = rp.AsValueString();
           break;
-        case StorageType.ElementId:
-          // NOTE: if this collects too much garbage, maybe we can ignore it
-          var id = rp.AsElementId();
-          var e = Doc.GetElement(id);
-          if (e != null && CanConvertToSpeckle(e))
-            sp.value = ConvertToSpeckle(e);
-          break;
+          // case StorageType.ElementId:
+          //   // NOTE: if this collects too much garbage, maybe we can ignore it
+          //   var id = rp.AsElementId();
+          //   var e = Doc.GetElement(id);
+          //   if (e != null && CanConvertToSpeckle(e))
+          //     sp.value = ConvertToSpeckle(e);
+          //   break;
         default:
           return null;
-          break;
       }
       return sp;
     }
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -253,13 +253,13 @@ namespace Objects.Converter.Revit
           if (sp.value == null)
             sp.value = rp.AsValueString();
           break;
-          //case StorageType.ElementId:
-          //  // NOTE: if this collects too much garbage, maybe we can ignore it
-          //  var id = rp.AsElementId();
-          //  var e = Doc.GetElement(id);
-          //  if (e != null && CanConvertToSpeckle(e))
-          //    sp.value = ConvertToSpeckle(e);
-          //  break;
+        case StorageType.ElementId:
+          // NOTE: if this collects too much garbage, maybe we can ignore it
+          var id = rp.AsElementId();
+          var e = Doc.GetElement(id);
+          if (e != null && CanConvertToSpeckle(e))
+            sp.value = ConvertToSpeckle(e);
+          break;
         default:
           return null;
           break;
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
@@ -197,8 +197,7 @@ namespace Objects.Converter.Revit
       //make line from point and height
       if (baseLine == null && baseGeometry is Point basePoint)
       {
-        var topLevelParam = revitColumn.get_Parameter(BuiltInParameter.FAMILY_TOP_LEVEL_PARAM);
-        var elevation = (double)((RevitLevel)ParameterToSpeckle(topLevelParam).value).elevation;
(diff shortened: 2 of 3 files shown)


teocomi (changes requested):
Yo Izzy, thanks for looking into this!
The preferred methods for extracting parameters and levels from when converting to Speckle are actually: `ConvertAndCacheLevel` and ` GetParamValue` see example:

https://github.com/specklesystems/speckle-sharp/blob/master/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial%20Classes/ConvertWall.cs#L134-L139

I think the names of these functions became a bit messy over time, we should clean them up at some point 😅

izzylys:
no prob!

okie, i didn't actually change much, just expanded what was there so this was the method that was already being used here to get the levels and params. will change this to the updated preferred method!

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -253,13 +253,13 @@ private Parameter ParameterToSpeckle(DB.Parameter rp, bool isTypeParameter = fal
...
             sp.value = rp.AsValueString();
           break;
-          //case StorageType.ElementId:
-          //  // NOTE: if this collects too much garbage, maybe we can ignore it
-          //  var id = rp.AsElementId();
-          //  var e = Doc.GetElement(id);
-          //  if (e != null && CanConvertToSpeckle(e))
-          //    sp.value = ConvertToSpeckle(e);
-          //  break;
+        case StorageType.ElementId:
```
teocomi, line 256: We should keep this commented as otherwise a lot of garbage is collected too

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
```diff
@@ -197,8 +197,7 @@ public RevitColumn ColumnToSpeckle(DB.FamilyInstance revitColumn)
       //make line from point and height
       if (baseLine == null && baseGeometry is Point basePoint)
       {
-        var topLevelParam = revitColumn.get_Parameter(BuiltInParameter.FAMILY_TOP_LEVEL_PARAM);
-        var elevation = (double)((RevitLevel)ParameterToSpeckle(topLevelParam).value).elevation;
+        var elevation = ConvertAndCacheLevel(revitColumn, BuiltInParameter.FAMILY_TOP_LEVEL_PARAM).elevation;
```
teocomi, line 200: Will this throw if the top level is null? Or is it never null?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertLocation.cs
```diff
@@ -112,7 +114,7 @@ private object LocationToNative(Base elem)
 
       if (elem is Column)
       {
-        //revit verical columns can only be POINT based
```
teocomi, line 115: so many typos, lol

teocomi (commented):
Minor changes pls!

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertLocation.cs
izzylys: 😅

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
izzylys: I don't think this can be null. unlike walls, there doesn't seem to be an "unconnected height" concept for cols and 
I didn't manage to make a beam without a top level

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
izzylys: ah tru, now that we're no longer using this method in the column conversion, I can safely comment this out again!

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
izzylys: the issue here before wasn't that there wasn't a top level, it was that the top level param was of type `ElementId` which wasn't handled by `ParameterToSpeckle` (not an issue anymore since we're no longer using this)

teocomi: Gotcha, txs!

PR #396, opened 2021-03-25
fix(core): fixes #356 and prevents dupes dynamic properties (detached…

### LINKED ISSUE #356: Core: prevent misuse of detachable properties
Current problem: let's say you have 2 base objects, one has a **typed** detached prop `[detached] wall.elements` and another has a **dynamic** detached prop `floor[@elements]`, I cannot write a function that handles them the same way
as the first would need to be called `base["elements"]` and the other `base["@elements"]`. This hampers the flexibility given by having dynamic properties as I might want to reuse the same function with both typed and dynamic objects.

Proposed solution: edit `DynamicBase.cs` so that, after being set a first time, a detachable property can be retrieved either with  `base["elements"]` or `base["@elements"]`. This also means **not** allowing having two properties with the same name that are detached and non-detached.

teocomi:
## Description

-  fixes #356 and prevents dupes dynamic properties (detached and non detached) to exist

For dynamic props, now the logic is that a prop becomes `detached` or `non detached` the first time it's created. Afterwards, it can be either accessed with `base["@prop"]` or `base["prop"]` equally.

For typed props, they can now be accessed with `base["@prop"]` or `base["prop"]` equally whether they are detached or not.

## Type of change

## How has this been tested?

- New tests were made

## Docs

- Will be updated ASAP

### DIFF
--- a/Core/Core/Models/DynamicBase.cs
+++ b/Core/Core/Models/DynamicBase.cs
@@ -66,7 +66,7 @@ namespace Speckle.Core.Models
 
       // TODO: Check for detached/non-detached duplicate names? i.e: '@something' vs 'something'
       // TODO: Instance members will not be overwritten, this may cause issues.
-      var checks = new List < (bool, string) >
+      var checks = new List<(bool, string)>
       {
         (!(string.IsNullOrEmpty(name) || name == "@"), "Found empty prop name"),
         // Checks for multiple leading @
@@ -81,7 +81,7 @@ namespace Speckle.Core.Models
       // Prop name is valid if none of the checks are true
       var isValid = checks.TrueForAll(v =>
       {
-        if (!v.Item1)r = v.Item2;
+        if (!v.Item1) r = v.Item2;
         return v.Item1;
       });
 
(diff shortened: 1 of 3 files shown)


izzylys (approved):
just tried this out and the rework looks good to me! the new behaviour is quite convenient 👍 

only thing that comes to mind is that now there is ambiguity re whether a dynamic prop is detached or not. 

eg you might think you're adding a detached prop with `base["@prop"]` but you're actually setting `base["prop"]` behind the scenes. I'm not sure if that's ok or if it could be user unfriendly behaviour. maybe if someone is actively trying to set `base["@prop"]` when `base["prop"]` exists, you should actually be removing "prop" and adding "@prop"? tho I guess issues arise here if the prop you're trying to set is actually a typed one and not a dynamic one... 

perhaps we could at least add a helper method that you can run to check if a prop is detached or not so a user can be sure they're adding a detached prop?

teocomi:
I hear you izzy! Since it seems the decision is between: "not properly setting a prop as detached" and "not being able to access a prop because of `@` confusion" I think the first is preferable :) But let's let @didimitrie weigh on this...

didimitrie:
Ok thought on this more. I don't really like the backchannel extra magic, and it has the potential to add extra confusion. There's an extra layer of indirection that I feel is not needed...

IMHO, the way to fix #356 is with a change in code, in how those funcs deal with the `["@elements"]` prop (see comment in issue). 

One thing that is an improvement is being able to allow, if one uses the square bracket notation, to get/set typed detachable props with it. Ie, if i do `base["@test"]`, it can pull in/try to set the typed detachable prop called `test` on that class, rather than add an extra dynamic prop called `@test`.

PR #429, opened 2021-04-12
CI: Docker server for Integration tests

### LINKED ISSUE #374: Tests/Core: use docker image to spin up server for integration tests on CI
This will make testing locally more difficult, but it's getting annoying; or perhaps we can set an environment var to specify which test to use, with easier to handle default for local testing. 

Raising this as a placeholder for discussion and reminder to ourselves :)

AlanRynne:
## Description

- Fixes #374

## Type of change

## How has this been tested?

- Unit/Integration Tests for Core now passing.

## Docs

- No updates needed.


Core/IntegrationTests/Fixtures.cs
```diff
@@ -14,8 +14,8 @@ namespace TestsIntegration
 {
   public static class Fixtures
   {
-    public static readonly ServerInfo Server = new ServerInfo { url = "https://latest.speckle.dev", name = "Test Server" };
-    
+    public static readonly ServerInfo Server = new ServerInfo { url = "http://localhost:3000", name = "Docker Server" };
```
AlanRynne, line 17: Only thing remaining for this issue to be closed is to decide what to do with this hardcoded URL so local testing would still use `latest`.

PR #460, opened 2021-04-27
Update schemabuilder behavior for Rhino and GH

### LINKED ISSUE #366: Rhino: surfaces with floor schema attached come out as polylines
**Current behaviour:** 
- select a bunch of rectangular surfaces
- apply schema to them (auto)
- send 
- attempt to [view them](https://speckle.xyz/streams/df3565183f/commits/327aa0f909): they come out as polylines! 
- attempt to receive them in another rhino file: nada (layer structure is created, but nothing gets baked). 

**Expected behaviour**: get them back out as surfaces! 

Why are we actually converting rhino elements when sending, if they have a schema, to that specific schema? Can't this be done on the other end? With the current behaviour, it appears as we're "lossy-fying" the interop process in a pre-emptive way that can be avoided.

### LINKED ISSUE #441: Gh&Rhino Schema Builder Behaviour v2
To be scoped further, but in short: send the geometry as is with attached "schema prop" → move conversions in the receiving end.
Related to #366

clairekuang:
## Description

Updates schemabuilder behavior in Rhino and GH by attaching BuiltElement as a prop.

GH:
 - CreateSchemaObject output is now the defining param input converted to Base, with the generated BuiltElement attached as a prop.
 - The `defining param` is the first input geometry param. This will register as a null value in the detached BuiltElement prop.
 - If no `defining param` is found, the output will be the generated BuiltElement (old functionality)
 - When a CSO object is converted back to GH, only the `defining param` will be converted

Rhino:
 - SpeckleSchema user attribute string will attach generated BuiltElement as a prop, same as GH
 - When a schema object is converted back into Rhino, the original geometry will be received *with* the SpeckleSchema user attribute string

Revit:
 - Adds a check on ConvertToNative for an attached SpeckleSchema property. Handles case where when sent from GH, the `defining prop` value of the BuiltElement object is null.

Fixes #441 
Fixes #366 

## Type of change

## How has this been tested?

- Manual Tests: updated BuiltElements test files and tested Rhino -> Revit -> Rhino, GH -> Revit -> GH

## Docs

- No updates needed

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/SchemaBuilder/CreateSchemaObject.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/SchemaBuilder/CreateSchemaObject.cs
@@ -258,6 +258,8 @@ namespace ConnectorGrasshopper
         return;
       }
 
+      var units = Units.GetUnitsFromString(Rhino.RhinoDoc.ActiveDoc.GetUnitSystemName(true, false, false, false));
+
       List<object> cParamsValues = new List<object>();
       var cParams = SelectedConstructor.GetParameters();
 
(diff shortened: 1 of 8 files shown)


ConnectorGrasshopper/ConnectorGrasshopper/SchemaBuilder/CreateSchemaObject.cs
```diff
@@ -293,24 +296,35 @@ protected override void SolveInstance(IGH_DataAccess DA)
           var extractRealInputValue = ExtractRealInputValue(inputValue);
           var objectProp = GetObjectProp(param, extractRealInputValue, cParam.ParameterType);
           cParamsValues.Add(objectProp);
+
+          if (!param.Optional && mainGeometryParam == null)
+            mainGeometryParam = objectProp;
```
teocomi, line 301: If I'm reading this correctly, we assume the first non-optional parameter in a BuiltElem constructor to be the `mainGeometryparam`, but this would not work for most of the Revit specific classes, see: https://github.com/specklesystems/speckle-sharp/blob/master/Objects/Objects/BuiltElements/Wall.cs#L72

A better approach would be to add a custom attribute [like this](https://github.com/specklesystems/speckle-sharp/blob/master/Core/Core/Kits/Attributes.cs#L32) to define what is the main geometry.

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -190,6 +190,17 @@ private string GetElemInfo(object o)
 
     public object ConvertToNative(Base @object)
     {
+      // schema check
+      var speckleSchema = @object["@SpeckleSchema"] as Base;
```
teocomi, line 194: I think we'll also need to add similar logic in the `CanConvertToNative` method so that it returns true if the `speckleSchema` prop is not null

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -190,6 +190,17 @@ private string GetElemInfo(object o)
...
+      if (speckleSchema != null) 
+      {
+        // find self referential prop and set value to @object if it is null (happens when sent from gh)
+        var objProp = speckleSchema.GetInstanceMembers().Where(o => !o.GetType().IsEnum).Where(o => !Utilities.IsSimpleType(o.GetType())).First();
```
teocomi, line 198: Wouldn't it be safer to look for an InstanceMember that _has the same type as @object_ and that _is null?_

teocomi (commented):
Didn't try running this locally, but pls have a look at the comments and let me know if they make sense :)

ConnectorGrasshopper/ConnectorGrasshopper/SchemaBuilder/CreateSchemaObject.cs
clairekuang: Improved the logic for now to look for objects.geometry params, and it's working for all of the revit built elements that can be successfully sent. I'm mostly hesitant to add an attribute flag to a prop in Objects when this issue is specific to GH atm

clairekuang:
Sent builtelements from the test files for [Rhino](https://latest.speckle.dev/streams/da89628505/commits/92bf49c47a) and [GH](https://latest.speckle.dev/streams/da89628505/commits/dbf75249bd). Tried to cover all currently supported BE objects.

Download the new test files in google drive: ...\07 Sample Models\Testing\02 BuiltElements Geometry

Issue #547, opened 2021-06-07
GH: Stream receiver doesn't like a valid stream


iltabe:
I'm facing a problem with this stream: `https://speckle.xyz/streams/0355f60d50/branches/02_base_model`
The receiver stream doesn't like it.
I tried already with a new component, a new file... same error:
![image](https://user-images.githubusercontent.com/64929000/121000524-9c804b80-c78a-11eb-9570-5edf56d671f6.png)

If I stream the last commit url of 02_base_model stream, I actually get another commit.. of another branch.... and empty!
![image](https://user-images.githubusercontent.com/64929000/121000968-157fa300-c78b-11eb-812b-a3e25eb99ce4.png)

Currently have the master at the `2.1.17` release.

iltabe:
I tried to push again on the branch `test001` the same data, and it actually do the same...

AlanRynne:
Hi @iltabe!! As usual, thanks for reporting this (and for pushing the Speckle connectors to its limits!!). We've managed to track down the issue and we'll discuss this afternoon how best to fix this.

FYI, the error is not related with the `stream url` itself, but rather with the amount of objects it contains and how we're sending the request to fetch them :)

iltabe:
Did I pushed too many objects? ahah 🙈

iltabe:
Hey @AlanRynne,
This issue is quite essential for me to fix. Any workaround I can apply myself to be able proceed?

teocomi:
We've prioritized this and should get a fix in place today/tomorrow!

AlanRynne:
Hey @iltabe, as @teocomi very well pointed out... we're merging this right now and looking to release all affected connectors.

If you're building your own connectors, the latest master should do the trick

iltabe:
It worked! Nice job!

PR #552, opened 2021-06-09
Download objects in batches (of 5k objects)

cristi8:


didimitrie:
@AlanRynne since you were debugging this the other day, could you give it a spin pretty please?

AlanRynne (approved):
All is good! We settled for 1k objects as 5k seemed to still throw the error in GH

AlanRynne:
Fixes #547

PR #549, opened 2021-06-08
Sync components

psarras:
## Description

- Adds the Synchronous Feature described here: #199

## Type of change

- New feature (non-breaking change which adds functionality)

## How has this been tested?

- Manual Tests (please write what did you do?)
I compared the data from the existing receiver with the data from this receiver and made sure they were the same.
I've made sure that only when the data were received the canvas become unlocked (and triggered downstream components)

## Docs

- Will be updated ASAP

Issue #199, opened 2021-01-29
Synchronous Senders / Receivers


psarras:
In Generative studies / Rhino Compute studies you need to rely on the synchronous nature of the document in order to make your workflows work. For this reason the GH_AsyncComponent component would not work. So far as I could tell there is no way to easily revert back to a Synchronous experience, therefore, I recommend creating a new component for the Sender / Receiver that extends the TaskCapable class and manage the different calls.

I've made a POC for the Receiver class here: https://github.com/psarras/speckle-sharp/tree/syncComponents

didimitrie:
Nice, just managed to have a look! Do we really need the task capable component workflow in this specific scenario? Was thinking that it might be just easier to dunk everything in solve instance, and just pepper all the async stuff with `.Result` where needed. 

Understand where this is coming from, and happy to have it developed fully.

psarras:
Hey Dimitrie,

Cool nice to hear. It needs a bit of clean up but happy to do it if you are
keen to merge.

What I found is that if you cannot for some reason do .result (it happens I
don't know why!!!) Then you will need to use this. I think I managed to do
it everywhere here but not 100% sure. The other benefit of this is that it
is multi-threaded so if you give 3 streams will occupy 3 threads to solve
the problem. What is the improvement here over baseline? Not sure :)

didimitrie:
Hah, `async` is full of weird gotchas in .net when you don't control the full app. Looking forward to the `SyncSend` too then!

psarras:
Hey @AlanRynne I've updated this branch to fix the conflicts with current master and the changes to the APIs, [here](https://github.com/psarras/speckle-sharp/tree/syncComponents). Let me know what you think.

There are a few ugly things I am planning to fix before making the PR. However, do you think we can decouple a few functionalities from the ```Operations.ReceiveComponent``` so that I can reuse them on the ```Operations.ReceiveComponentSync```? I am looking specifically on the ```DoWork``` method. This is currently part of the ```WorkerInstance``` class that I do not need as I am not providing updates on every tick. Let me know what you think.

p.s. Could we also put classes on separate files? I realise that this needs to be done in coordination as it is always a pain to merge changes like that.

AlanRynne:
Hi there! First of all, thinks for this effort @psarras :) I had a quick look at your code to wrap my head around it 👍🏻  I see what you mean with the nasty `DoWork` task in your code.

[...]

PR #553, opened 2021-06-09
Sync sender

psarras:
## Description

- Related to #549 #199 

## Type of change

- New feature (non-breaking change which adds functionality). Adds the Sender to accompany the Synchronous Receiver.

## How has this been tested?

- Manual Tests (please write what did you do?). Sending is synchronous with GH and correctly triggers any downstream objects. 

## Docs

- Will be updated ASAP


psarras:
@AlanRynne this should be merged after the #549. 

Also there is a slightly off think I don't like on the way I've done this. In order for the Task-Capable component to work, I cannot have all inputs as trees! I need at least one to be something else. In this case I have the message set to Item access. As I understand it you are only using the first item there and nothing else. However this means that in this component if someone puts two messages then the component would send off everything twice. I think I can hack it further but maybe it is better to make so each thread would be in charge to send 1 package to each transport and with individual messages? So I am proposing to make the an overload method for ```Operations.Send``` with signature: 

```cs
Operations.Send(Base @object, CancellationToken cancellationToken, ITransport transports = null, bool useDefaultCache = true, Action<ConcurrentDictionary<string, int>> onProgressAction = null, Action<string, Exception> onErrorAction = null
```

That way I would be able to call this once per Transport / Message. And would allow me to set the transport input to item as well. 

I guess the point of this message is to understand if this is a good idea or not. Is there a hidden optimization going on? Why are you looping through all your transports and what would be the benefit to provide more than one transport on this component? I feel I am looking at a niche case but please enlighten me!

cc. @didimitrie

AlanRynne:
Hi Stam! I was reviewing your nodes and everything seems to work fine! Regarding your question about `item access` in the messages, we could just prevent this from happening by forcing the component to only compute in the first iteration. This should work fairly well to keep things consistent (and prevent double sends) while we come up with a more consistent strategy.

I'm going to quickly test if this is feasible with the `TaskCapableComponents` and i'll get back to you :)

AlanRynne:
Hey again @psarras, this 2-liner fix should prevent the component from executing twice, while also warning the user. 

`DisableGapLogic()` prevents the output from generating extra empty branches, as there should only be 1 item in the output.

Not sure if I should commit it to your repo, merge and commit it myself, or let you apply it and check it before merging this in. Do you have any preference?  Pinging @didimitrie too!

Screenshot 2021-06-11 at 09 22 02

psarras:
Hey Alan, this is great! I didn't know you could that, that surely
simplifies things. Let me try it this morning and let you know!

I wonder though if there is a benefit to have it multi threaded and take
advantage the Task Capable fully?

AlanRynne:
Great, let us know how it goes 😄 

As for your question regarding multi threaded components, I'm really not sure as I never got to play around with them. It may be tricky as, as far as I understand, the task capable component was designed to execute many unrelated tasks while the send/receive nodes should actually perform a single task (send one commit) that can potentially be split into many "sub-tasks" (converting every object... detaching...) Which is taken care of by the Decomposition API in `Core` and the `Kits`.

But I'd be happy to hear any ideas if you have them! There are also some optimisations we're currently working on that will speed up things on our end too 😉 🚀

psarras:
@AlanRynne I mean here you must be doing one await per transport, but there is no reason to do those sequentially we can do them in parallel. 

image

We can avoid re-writting this method if on the component we do something like this:

image

BTW, I realised what you did, sorry I was not familiar with the ```DisableGapLogic``` and thought it was doing something special on the TaskCapable component that allowed you to set messages back to Tree. But yeah it makes sense, all working fine on my end. I will re-push this with your fixes.

AlanRynne:
Interesting! You're right, although not sure if this will step on anything else (like progress reporting). I'll defer to @didimitrie for this since he's the expert on that side of things 👍🏻 

Could be worth a look to see if it speeds up things when sending to multiple transports.

I'll wait for your changes and most likely both PRs will go in today 😁

psarras:
OK, I hope I am not pushing for too much, but I made the stream input an item so now you can have individual messages per transport and you also shave off a few seconds of computation. With my tests I pushed 8 streams, in 6.8 seconds using the original component, 6 without multithreaded and 4.5 with. We can probably improve this further but maybe not important for this PR. Let me know what you think. 

Uncommenting lines 206-207 + 217 replace line 216 and that allows you to go back to the previous behaviour.

image

AlanRynne:
I'll merge it in and we can discuss further improvements down the line! :)

AlanRynne (approved):

PR #560, opened 2021-06-11
fix(revit): better base point transformation for XYZ vectors, improved views

### LINKED ISSUE #559: Revit: vector conversions are failing with moved base point
Observed from bad view conversions: base points that have an angle form north will result in XYZ vectors being improperly transformed (compared to XYZ points).

clairekuang:
## Description

Revit XYZ class can be points or vectors: in the case of a transformed base point (translation, angle from north), the ToInternalCoordinates() and ToExternalCoordinates() was working properly but not for vectors. This was resulting in faulty view conversions where camera location was correct but direction was incorrect.

1.  Added TotalTransform property to BetterBasePoint class
2.  Modified Point and Vector conversions to use TotalTransform
3.  Fixed View conversions to use Point and Vector conversions for all props

- Fixes #559

## Type of change

## How has this been tested?

- Manual Tests: sent sample arch and structure file from Revit with views, some light additional testing with current test streams 

## Docs

- No updates needed


teocomi (approved):
🤯

PR #575, opened 2021-06-16
Dim/hotfix

didimitrie:
## Description

This is an emergency fix for https://speckle.community/t/grasshopper-send-issue/1452. It probably affects all currently released connectors, so we will need to do a fast hot fix release tomorrow. 

tenor

## Type of change

## How has this been tested?

I sent [64k points from Gh](https://speckle.xyz/streams/1ce562e99a/commits/5b913e131f).


cristi8 (approved):
yep, my bad, the unwritten "when sending, we batch and then ask about each small batch and should not exceed 1k objects" assumption failed. thx for the quick pr

PR #612, opened 2021-07-13
basic synchronous Account from token and url

psarras:
## Description

Adds a new component that creates a local Account from a Token and Url.
The goal is to have a synchronous way to inject the Account in Rhino Compute. This should be able to serve different users, so we cannot have something on the Rhino Compute servers baked in. The way it is setup right now it would play nice with they way Rhino Compute works.

## Type of change

- New feature (non-breaking change which adds functionality). 

## How has this been tested?

Generated a token from my account, input this plus the address of the server. Dropped the ManualAccount component and setup inputs. Waited for 1second (canvas blocked) got an id and a remark with my new account. 

## Docs

- Will be updated ASAP


didimitrie:
Welcome back from holidays @psarras! I'm curious about this PR, as it's opening up a pretty big and bad security vulnerability - if that gh definition gets saved it's like... storing your password in plaintext, on a post-it note, potentially on a network drive. 

What are the use-cases that need it?

psarras:
Hey @didimitrie, valid concern, not sure how to overcome it. Open to alternatives as always. 

The goal is to have a synchronous way to inject the Account in Rhino Compute. This should be able to serve different users, so we cannot have something on the Rhino Compute servers baked in. The way it is setup right now it would play nice with they way Rhino Compute works. Maybe we can have a very special input that gets erased when you save the gh file? Not sure if it is possible.

BTW I will update the description.

didimitrie:
Hi Stam, sorry for the really late reply. What's the architecture planned around the Rhino Compute part? If my assumptions are correct, the end user will need to somehow upload the grasshopper definition to a rhino compute application. This is the magic moment where you can associate credentials with the gh definition safely; the application can be a registered third party app for a given speckle server, and thus hold a user's api token that it can pass on to the compute workflow.

Is this something feasible? I'll discuss this with @cristi8 tomorrow, but I'm really hesitant to merge something that allows for the equivalent of plain text token storage inside potentially portable grasshopper files by everyone...

psarras:
Hey Dimitrie. Yeah something like that. However, the user will not be the one to upload the GH definition necessarily. The idea is you might have generic definitions that do X, Y, Z and that do not include any credentials, only placeholders (empty strings). The a front end app will fill those in for you after you have "logged" in with your Speckle account. 

I guess the awkwardness is that RhinoCompute can edit existing params from the gh canvas (AFAIK). 

Another option for this piece of work is to put it in a different library and only include it in our workflows, while also taking the risk! Let me know how the discussion with @cristi8 goes and if you prefer this separate and we can manage it! Cheers!

teocomi:
Would it be possible to, somehow, hardcode in the node a variable name that is then set by your frontend? Something like an environment variable, but maybe specific to Rhino Compute?

psarras:
Hey @teocomi !I will check with the Rhino Compute team and get back to you. Thanks for the feedback guys

CapnK:
@didimitrie - As you've closed this issue, where would you like to continue this discussion?
There are some clarifications to make, and I'm not sure the environment variable would work as the webserver may receive concurrent requests, with the potential for two GH solve's to be running at the same time, in the same process space.

didimitrie:
Hi @CapnK! I've closed this as we're certain we do not want to officially allow for plain text tokens stored in grasshopper definitions: this has the potential to undermine the trust in the whole Speckle ecosystem. Speckle Core is explicitly and carefully architected to prevent any kind of security critical information to be serialised in any files in which Speckle might be used, so it's a definitive no.  

Ideally this would've been discussed first in an _actual_ issue rather than a surprise PR 😎 IMHO, the best place would be the [forum](speckle.community/) as it's not something related to the code in this repo but rather to the way you will architect your rhino compute server and its integration with Speckle. We can help you on this front and explore options together if you want to.  

As a post mortem, you can much easier do what this PR intends via a few lines of C# inside a scripting component: 

image

Agreed that env variables would not work. If your server application is registered as a third party application with your Speckle Server, it will be granted a token which it can use to act on behalf of the user. You can store this token somehow in your backend and inject it programmatically in the gh definition before it is run by the rhino compute engine.

psarras:
Hey @didimitrie, I guess I was going the long way around to solve this issue. I see this as a simpler and neater solution. But you can see how I was lead in that direction if the Account class doesn't expose a minimal constructor with token / api and instead there are a bunch of public properties I cannot be certain are unnecessary?

psarras:
I was a bit quick to praise this solution as it seems that most components require this account to be added / be visible on the Speckle Manager: 
image
The proposed solution was adding it on the local accounts. Any thoughts?

teocomi:
Hey Stam,
Would you be able to use local accounts instead? 
https://speckle.guide/user/manager.html#adding-accounts-manually
https://github.com/specklesystems/speckle-sharp/blob/main/Core/Core/Credentials/AccountManager.cs#L126

On a similar note, would RhinoCompute have access to Speckle's kit?

daviddekoning:
Let me take a step back and reframe our aims. Our high level goal is to introduce accounts at a grasshopper canvas level of scope, in a secure manner. (This is Arup's goal, I'm not suggesting that this is necessarily a Speckle-wide goals).

Question 1: do we want to assume grasshopper canvas scope is the same as computer scope? This is a Rhino.Compute set up question and I'll follow up with @CapnK. It is much easier to deal with this if each canvas is run in a separate worker with a clean environment, because then you can just write an account to the SQLite DB as @teocomi suggests.

However, for this discussion, the interesting question is: can we securely define an account on the GH canvas level?

Here's one idea: when searching for accounts, look for a SpeckleAccounts.json in the same directory as the grasshopper file.

The "specify account" component then just takes username and server URL as input and searches the json file for the rest of the info (and then the DB if it doesn't find it). The advantage to this method is that it doesn't change how all the other components deal with accounts, so we maintain the token protections in place, and the GH file can be  shared without including the token.

The Rhino.Compute environment can then authenticate the user, save the token to a separate file, run the script then delete the token file. 

Thoughts?

daviddekoning:
p.s. my concern with directly adding accounts to the DB is that if multiple canvases are running in a single Rhino.Compute runtime, we are potentially exposing tokens to multiple users who shouldn't have them.

didimitrie:
I feel this is the right direction. Ideally, the most secure way to go would be to inject somehow the token at runtime in the gh canvas through the backend service that will most likely wrap around the compute engine. E.g, user goes to web ui, logs in, uploads gh definition and sets params (which stream to post data to, which branch). The backend service then knows the identity of the user, and can inject the token in the gh canvas when it's run. I'm assuming here that the backend+frontend service are a registered third party app on the speckle server you're running this against!

daviddekoning:
yup, that's the exact approach we use. The question is how to inject them.

Rhino.Compute has a mechanism to inject data directly into the canvas ([using RH_IN tags](https://discourse.mcneel.com/t/grasshopper-within-compute-rhino3d/77656/2?u=david26)), thus the component in this PR that takes username, server and token as input. The intention is for this component to only ever be used with injected data, but the question is how to do it without creating something easily abused...

psarras:
@teocomi Yes this is what this PR was doing. Anyway, the idea has been moved here: https://github.com/arup-group/SpeckleExtensions

@daviddekoning we can maybe add something that removes it once the canvas finishes?

daviddekoning:
Honestly, in this situation, GH is a programming environment, and we can't stop people from hardwiring their tokens into their Python or dotNet code and committing it to GitHub. My gut feeling is to make it easy to do this securely, but making it "impossible" for programmers to make mistakes will be counter productive.

teocomi:
Hey @psarras thanks!
I think it makes a lot of sense to have these components in a separate repo, given their use. In case you think they should instead be part of the core Speckle ones, maybe start a discussion on the forum, so other Speckle+RhinoCompute users can chime in.

@daviddekoning I would argue that developers have a better understanding of tokens and basic security concepts that average computational designers usually don't have ;) we don't want to make it impossible to use tokens from GH, we simply don't want to encourage it!

PR #633, opened 2021-07-22
Option convert to native

psarras:
## Description

- Added option on the context menu to not convert to native for the Sync Receive Component

## Type of change

## How has this been tested?

## Docs

- Will be updated ASAP


AlanRynne:
Hey @psarras!

Thanks for this! We've gone over your changes and it's a great (and necessary) addition.

There's one change i'd like to suggest to keep things aligned with the Object Management nodes.

The way other nodes work is that the "Do not convert" option is part of the KitSelection process. When this option is selected, the `Converter` instance in that component is set to `null`, and, whenever it's used, we just perform a null check and pass-through the value instead of converting.

In your node, both options can be active at the same time (i.e. you can have a Kit selected AND the "Do not convert" option active). Is there a specific reason you'd need both active? It would be preferable that all nodes that use this option have a consistent behaviour, but we can organise a video call if you prefer to discuss this things in detail if you prefer.

Just for guidance:


Here's how we're adding the do not convert option.
https://github.com/specklesystems/speckle-sharp/blob/56c0b700d953020ef6c8ad9e06d97dfb22e17b3b/ConnectorGrasshopper/ConnectorGrasshopper/Objects/SelectKitTaskCapableComponentBase.cs#L87-L92

And here's the Kit selection logic.
https://github.com/specklesystems/speckle-sharp/blob/56c0b700d953020ef6c8ad9e06d97dfb22e17b3b/ConnectorGrasshopper/ConnectorGrasshopper/Objects/SelectKitTaskCapableComponentBase.cs#L46-L63

And then when using the converter:
https://github.com/specklesystems/speckle-sharp/blob/56c0b700d953020ef6c8ad9e06d97dfb22e17b3b/ConnectorGrasshopper/ConnectorGrasshopper/Objects/CreateSpeckleObjectTaskComponent.cs#L160

I also noticed that I didn't add this feature on the original Receive node. Would you be willing to add this functionality there too?

PS. Sorry for delay in getting back to you, I was certain I had answered last week... not sure what happened with that message! 😓

PR #653, opened 2021-08-02
fix(gh): Empty optional arguments no longer block sender

### LINKED ISSUE #580: Create Speckle Object Optional Input Parameter Error
Discovered an issue with the **CreateSpeckleObject** node potentially not using the `Optional` input parameter correctly. When I passed a `Base`  object created from that node that contained an empty `Optional` parameter, the conversion process that happens in the **Sender** node gets stuck on conversion process. 

To replicate:
- Use **CreateSpeckleObject** and set an input parameter as `Optional` 
- Leave optional parameter empty 
- Pass created object into sender node and try sending to stream, it should throw an error saying that there is empty data trying to be converted

AlanRynne:
Fixes #580

This fixes an error when optional arguments existed but had no data (or were null) were making the sender get stuck and throw a 'Null Exception'. Right now, it will ignore any non-set properties (just like the 'Expand Speckle Object'  works, but I'm aware this may not sit well with some users.

We've had other petitions to keep the null items when being sent to the server, so I'm not sure if we should tackle this now, or in a separate issue.

Bonus: Better error handling when sender performs conversions.

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/Extras/Utilities.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/Extras/Utilities.cs
@@ -187,10 +187,14 @@ namespace ConnectorGrasshopper.Extras
       if (subclass) 
         return @base;
       var copy = @base.ShallowCopy();
-      copy.GetMembers().ToList().ForEach(keyval =>
+      var keyValuePairs = copy.GetMembers().ToList();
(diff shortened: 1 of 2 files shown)


AlanRynne:
Added a try/catch clause on the first part of the sender logic to prevent conversion errors from being mishandled.

Issue #394, opened 2021-03-25
Feat Revit: Additional conversions


teocomi:
Some of the additional direct conversions we should support ToSpeckle and if possible ToNative include:

- [x] Grid
- [x] Area
- [ ] Sloping Floors
- [ ] HVAC zones
- [ ] Spaces
- [ ] Analytical objects
- [x] Rooms to native https://speckle.community/t/sending-walls-and-rooms-from-revit-2020-to-2019/1288/5
- [x] Ceiling creation in revit 2022 https://speckle.community/t/create-ceiling-using-speckle-schema/1471/2
- [x] Direct shape to Speckle
- [ ] Element parts https://knowledge.autodesk.com/support/revit-products/learn-explore/caas/CloudHelp/cloudhelp/2019/ENU/Revit-Model/files/GUID-DA150C6B-996C-4C70-9E8C-3C536C232851-htm.html
- [x] Rebar https://www.revitapidocs.com/2021.1/70fd7426-f4a4-591c-8c06-3c18dda45e7d.htm
- [x] Wall by profile (look into it first, there should be a dynamo component) https://speckle.community/t/revitwall-by-profile/1559/4

teocomi:
Added element parts
https://knowledge.autodesk.com/support/revit-products/learn-explore/caas/CloudHelp/cloudhelp/2019/ENU/Revit-Model/files/GUID-DA150C6B-996C-4C70-9E8C-3C536C232851-htm.html
https://speckle.community/t/revit-unit-issue-with-reinforcement-family/1508/5

teocomi:
Closing this for now until ppl scream for stuff :)

PR #695, opened 2021-08-27
Jm/revit spaces conversion

jenessaman:
## Description

- Addresses to-do item for Spaces in https://github.com/specklesystems/speckle-sharp/issues/394 (the ToSpeckle and ToNative methods for Space objects and SpaceSeparationLine objects added)
- Fixes ToNative method for Room item in https://github.com/specklesystems/speckle-sharp/issues/394 (use scaled center point instead of speckle center point for placement)

## Type of change

## How has this been tested?

- Manual Tests, sent space separation lines, spaces and rooms to/from Rhino/GH and Revit

## Docs

- Will be updated ASAP

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
@@ -99,7 +99,8 @@ namespace Objects.Converter.Revit
           if ((BuiltInCategory)o.Category.Id.IntegerValue == BuiltInCategory.OST_RoomSeparationLines)
           {
             returnObject = RoomBoundaryLineToSpeckle(o);
-          } else if ((BuiltInCategory)o.Category.Id.IntegerValue == BuiltInCategory.OST_MEPSpaceSeparationLines)
+          } 
+          else if ((BuiltInCategory)o.Category.Id.IntegerValue == BuiltInCategory.OST_MEPSpaceSeparationLines)
           {
             returnObject = SpaceSeparationLineToSpeckle(o);
           }
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
@@ -99,6 +99,9 @@ namespace Objects.Converter.Revit
           if ((BuiltInCategory)o.Category.Id.IntegerValue == BuiltInCategory.OST_RoomSeparationLines)
           {
             returnObject = RoomBoundaryLineToSpeckle(o);
+          } else if ((BuiltInCategory)o.Category.Id.IntegerValue == BuiltInCategory.OST_MEPSpaceSeparationLines)
+          {
+            returnObject = SpaceSeparationLineToSpeckle(o);
           }
           else
           {
@@ -126,6 +129,9 @@ namespace Objects.Converter.Revit
           break;
         case DB.Mechanical.Duct o:
           returnObject = DuctToSpeckle(o);
+          break;
+        case DB.Mechanical.Space o:
+          returnObject = SpaceToSpeckle(o);
           break;
         case DB.Plumbing.Pipe o:
           returnObject = PipeToSpeckle(o);
@@ -303,6 +309,9 @@ namespace Objects.Converter.Revit
         case BERC.RoomBoundaryLine o:
           return RoomBoundaryLineToNative(o);
 
+        case BERC.SpaceSeparationLine o:
+          return SpaceSeparationLineToNative(o);
+
         case BE.Roof o:
           return RoofToNative(o);
 
@@ -343,6 +352,9 @@ namespace Objects.Converter.Revit
         case BE.GridLine o:
           return GridLineToNative(o);
 
+        case BE.Space o:
+            return SpaceToNative(o);
+
         // other
         case Other.BlockInstance o:
           return BlockInstanceToNative(o);
@@ -375,6 +387,7 @@ namespace Objects.Converter.Revit
         DB.Architecture.TopographySurface _ => true,
         DB.Wall _ => true,
         DB.Mechanical.Duct _ => true,
+        DB.Mechanical.Space _ => true,
         DB.Plumbing.Pipe _ => true,
         DB.Electrical.Wire _ => true,
         DB.CurtainGridLine _ => true, //these should be handled by curtain walls
@@ -455,6 +468,7 @@ namespace Objects.Converter.Revit
         BE.View3D _ => true,
         BE.Room _ => true,
         BE.GridLine _ => true,
+        BE.Space _ => true,
         Other.BlockInstance _ => true,
         _ => false
 
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
@@ -455,6 +455,7 @@ namespace Objects.Converter.Revit
         BERC.ModelCurve _ => true,
         BE.Opening _ => true,
         BERC.RoomBoundaryLine _ => true,
+        BERC.SpaceSeparationLine _ => true,
         BE.Roof _ => true,
         BE.Topography _ => true,
         BER.RevitFaceWall _ => true,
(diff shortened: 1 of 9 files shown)


teocomi:
Thanks @jenessaman , I'll review this week!

Objects/Objects/BuiltElements/Space.cs
```diff
@@ -0,0 +1,54 @@
...
+        public double area { get; set; }
+        public double volume { get; set; }
+        public Point basePoint { get; set; }
+        public Level level { get; set; }
+        public double baseOffset { get; set; } = 0;
+        public Level upperLimit { get; set; }
+        public double limitOffset { get; set; } = 0; 
+        public ICurve boundary { get; set; }
+        public string spaceType { get; set; }
+        public string zoneId { get; set; }
```
teocomi, line 24: What is the ZoneId? In general, we want to avoid sending Elemenet Ids since they will vary from revit document to document unless they are some sort of fixed ids. 
A better way to support zones could be to pass their elmenent name and then try to match it when receiving.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
```diff
@@ -0,0 +1,117 @@
...
+            var basePoint = PointToNative(speckleSpace.basePoint);
+            var upperLimit = LevelToNative(speckleSpace.upperLimit);
+
+            // create new space if none existing, include zone information if available
+            if (revitSpace == null)
+            {                
+                revitSpace = Doc.Create.NewSpace(level, new UV(basePoint.X, basePoint.Y));
+                if (speckleSpace.zoneId != null)
+                {
+                    var speckleZone = Doc.GetElement(speckleSpace.zoneId) as DB.Zone;
```
teocomi, line 28: (see commend below on element ids)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
```diff
@@ -0,0 +1,117 @@
...
+                        var spaceSet = new DB.SpaceSet();
+                        spaceSet.Insert(revitSpace);
+                        speckleZone.AddSpaces(spaceSet);
+                    }
+                }
+            }
+            else
+            {                
+                // unplaced space, so just delete and recreate from scratch (but make sure to copy host zone info)
+                // ideally we just place/move the existing space instead, but this seems like a real pain to do with the revit api? should update in future!
```
teocomi, line 40: I feel your pain

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+
+            revitSpace.Name = speckleSpace.name;
+            revitSpace.Number = speckleSpace.number;
+
+            // if user does not specify an UpperLimit level, assume use of default UpperLimit (current level) and default offset values (base offset of 0, limit offset is distance to next level above)
+            if(upperLimit != null)
+            {
+                revitSpace.get_Parameter(BuiltInParameter.ROOM_UPPER_LEVEL).Set(upperLimit.Id); // not sure why, don't actually seem to be able to set the UpperLimit property when UpperLimit is the same as Level - it stays null after assignment!
```
teocomi, line 61: Not sure why it doesn't work, I checked and it's the correct param name!
Btw, you can use this to set it directly:

`TrySetParam(revitSpace, BuiltInParameter.ROOM_UPPER_LEVEL, upperLimit);`

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -375,6 +387,7 @@ public bool CanConvertToSpeckle(object @object)
         DB.Architecture.TopographySurface _ => true,
         DB.Wall _ => true,
         DB.Mechanical.Duct _ => true,
+        DB.Mechanical.Space _ => true,
```
teocomi, line 390: I think we're missing a `CanConvertToSpeckle` for the `SpaceSeparationLines`.

teocomi (commented):
All looking good, just a few minor remarks, lemme know if you wanna chat about them 👍

Objects/Objects/BuiltElements/Space.cs
```diff
@@ -0,0 +1,54 @@
...
+using Speckle.Core.Kits;
+using Speckle.Core.Models;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Objects.BuiltElements.Revit.Curve;
+
+namespace Objects.BuiltElements
+{
+    public class Space : Base, IHasArea, IHasVolume, IDisplayMesh
```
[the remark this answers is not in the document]
jenessaman: Ahh yea sorry, good point! Have updated boundary, upperLimit and limitOffset to outline, topLevel and topOffset respectively.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            var speckleSpace = new Space();
+            speckleSpace.name = revitSpace.Name;
+            speckleSpace.number = revitSpace.Number;
+            speckleSpace.basePoint = (Point)LocationToSpeckle(revitSpace);
+            speckleSpace.level = ConvertAndCacheLevel(revitSpace.LevelId);
+            speckleSpace.upperLimit = ConvertAndCacheLevel(revitSpace.get_Parameter(BuiltInParameter.ROOM_UPPER_LEVEL).AsElementId());
+            speckleSpace.baseOffset = GetParamValue<double>(revitSpace, BuiltInParameter.ROOM_LOWER_OFFSET); 
+            speckleSpace.limitOffset = GetParamValue<double>(revitSpace, BuiltInParameter.ROOM_UPPER_OFFSET);
+            speckleSpace.boundary = profiles[0];
```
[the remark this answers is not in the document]
jenessaman: Yes! Have updated to include "voids" to match Rooms.

Objects/Objects/BuiltElements/Space.cs
jenessaman: ZoneId is meant for the element id of the Zone which contains the current Space - have now updated to use that element name rather than id to keep track of this info.

A related item - do you want to just add a `Zone` object to the BuiltElements? Would you prefer to just have this object included as a detachable property in the `Space`?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
jenessaman: Noted, updated to using element name of zone, as opposed to element id, as recommended.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
jenessaman: Hahah 😢

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
jenessaman: Thanks, will use that going forward! Still not sure that param is behaving as expected unfortunately? Doesn't seem to be possible to set the value when UpperLimit is null (ie. as with a newly created Space) and the UpperLimit level is to be the same as the Level level: 
image

What do you think - shall we post on Revit forums?

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
jenessaman: They come through as `ModelLines`, as the `RoomBoundaryLines` do, so they are captured in the `DB.ModelCurve` case.

Objects/Objects/BuiltElements/Space.cs
teocomi: I think it's fine like this for now! We might decide to do what you suggest later if needed and deprecate this prop.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertSpace.cs
teocomi: Well, Revit can be v picky when setting parameters. Eg. for columns it wouldn't take offsets unless they are set in a specific order. I guess something similar is at play here!
I'm happy to merge this PR despite this, and I'll leave it up to you to decide whether to post on forums or try to find what makes it work (if anything!)

PR #723, opened 2021-09-14
removed system.text.json dependency

cristi8:


teocomi (approved):
Testes in Revit 2019 and 2021, LGTM!

PR #727, opened 2021-09-15
GH/Rhino: Adds multi platform "postBuild" command again

### LINKED ISSUE #724: GH: Split converter broke the Mac build
The recent split between Rhino 6 and 7 Converters broke the GH build :(

Reinstating the missing mac postbuild event should do the trick

AlanRynne:
When we did the split of Rhino converters per version, we lost the `PostBuild` commands that are necessary to run GH on Mac (plus some CI magic to prevent them).

Now they are back in all their glory.

Fixes #724


AlanRynne:
@teocomi and @clairekuang, added you as reviewers for this. I have already tested it on my side on both Windows and Mac, but double checking never hurts!

- [ ] Mac: 
    - [ ] Check that the GH Connector project builds properly
- [ ] Windows:
    - [ ] Check GH Connector project builds properly
    - [ ] Check Rhino connectors for proper build too

PR #739, opened 2021-09-27
Compute id for all base objects

### LINKED ISSUE #738: Revert: all speckle base objects should have ids, regardless of detachment or not

cristi8:
Closes #738


didimitrie (approved):
approved, without building/testing - relying on cristi's magic touch here.

PR #762, opened 2021-10-05
bug(rhino): handle block transforms with sketchup conventions

### LINKED ISSUE #758: Rhino: Various blocks issues
Trying to receive a stream with blocks throws an error in Rhino after the first time. User needs to manually open the Rhino `BlockManager` to delete the block definitions to receive updates on the stream

clairekuang:
## Description
-No longer deletes commit layer before receiving commit.
-Discovered that sketchup sends block instance transforms where the translation vector is not necessarily [x,y,z,1]: the last value can be a non-1 divisor value. Rhino is now handling division in ToNative conversions.
-Discovered that sketchup sends block instance transforms where the last row is not necessarily the identity row: it's sometimes [0,0,0,0] instead of [0,0,0,1]. This usually indicates vector transform (not a point transform), **need to figure out why sketchup is sending these**. For now, rhino is changing the last row from [x,y,z,0] to [x,y,z,1] which seems to fix it for now.

**NOTE**
When testing, Sketchup > Rhino and Sketchup > Acad work well.
Sketchup > Acad > Rhino works well.
Sketchup > Rhino > Acad has a few blocks with nested block defs that are not translated correctly. Needs further debugging in acad.

- Fixes #758 

## Type of change

## How has this been tested?

## Docs

- No updates needed

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
@@ -256,10 +256,6 @@ namespace SpeckleRhino
 
       // give converter a way to access the base commit layer name
       RhinoDoc.ActiveDoc.Notes += "%%%" + commitLayerName;
-
-      var existingLayer = Doc.Layers.FindName(commitLayerName);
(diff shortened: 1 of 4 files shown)


izzylys (approved):
done some tests between sketchup and acad and it all seems to be working perfectly now! ✨

PR #851, opened 2021-11-15
Gh/Dyn: `null` value alignment

### LINKED ISSUE #491: Support sending and receiving null values and properties
How we handle `null` values varies between connectors. Our most strict implementation is GH, which directly prevents pushing null objects on a _data tree_.

Dynamo apparently has no inconvenient on sending `null` values.

The main reason behind disallowing `null` values in GH was to prevent _index shifting_ in lists of objects, as the `null` values where never serialized: `[0,null,1]` became `[0,1]` after serialization/deserialization.

There's quite a debate in the team on wether they should be allowed or not, but we should handle all cases in a consistent manner.

AlanRynne:
## Description

- Fixes #491 🥳  

Null values are now allowed everywhere in GH and Dyn.

## Type of change

## How has this been tested?

Tested sending nulls along with other objects to/from Grasshopper/Dynamo in all directions. Null values got properly serialised and deserialised on each program.

Also ensured that receiving in DUI connectors like Rhino/Revit did not break anything.

## Docs

- No updates needed as there is no reference of this anywhere in the docs.

### DIFF
--- a/ConnectorDynamo/ConnectorDynamo/SendNode/Send.cs
+++ b/ConnectorDynamo/ConnectorDynamo/SendNode/Send.cs
@@ -239,16 +239,14 @@ namespace Speckle.ConnectorDynamo.SendNode
       {
         if (_transports == null)
           throw new SpeckleException("The stream provided is invalid");
-        if (_data == null)
-          throw new SpeckleException("The data provided is invalid");
-
+        
         long totalCount = 0;
         Base @base = null;
         var converter = new BatchConverter();
         try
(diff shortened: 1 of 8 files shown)


teocomi:
Not sure if it should be part of this PR or not, but nulls were preventing revit params to go through, so we could do some testing there too :)

AlanRynne:
Closing! As discussed with Matteo, I'll open a new issue to track the "missing revit params" and fix it in the near future.

Issue #630, opened 2021-07-21
GH: Query data on the Receive Component


psarras:
I was planning to work on adding some functionality for querying objects before fetching them from the Receive Component (to start with the Sync version but we can also add it on the original if you want). See this for Speckle v1 if you haven't already. 

![image](https://user-images.githubusercontent.com/3820417/126522066-866d86c3-c8df-46f1-a675-acb914056e0e.png)

First of all let me know if you are thinking of adding this so I will not collide. Happy to look at it of course if you are not actively developing it.

I think this should be the place to add further arguments. Let me know if you think otherwise please.

![image](https://user-images.githubusercontent.com/3820417/126521323-eaaa5697-e89e-4e8c-9e0c-17b864e2a5d1.png)

Lastly if you have any tips and ideas that might help me implemented let me know!

teocomi:
Hey Stam! Thanks for this and the [forum thread](https://speckle.community/t/query-data-on-receivers/1622) (and sorry for the slow reply!), I think it'd be very cool to have! 
I believe there are changes in the backend we are planning to do that would make querying even more powerful, so we'd suggest holding off a bit on this. @didimitrie is on holiday atm but will reply with more details once he gets the time!

psarras:
Cheers @teocomi! No problem let us know! I've posted a forum post as well, in case we can get a bit more feedback on what is best from a user's perspective.

teocomi:
Correct, and we don't have an ETA on that yet, but it's quite high priority as it's blocking a few things. I hope we can tackle it before the new year.

In the meantime, not ideal, but a workaround is to do some querying on the client side...

psarras:
Hey Matteo, thanks for letting us know! On client, side, I guess you mean receiving everything as usually and then filtering the objects appropriately? We were thinking about about that, we might give it a go! Cheers.

teocomi:
Yep exactly, but I guess that could be done with native GH components... So if don't want to develop something that might become deprecated soon, you might just do so :)

AlanRynne:
Closing this as stale! If it needs revisiting, we can re-open it in the future ;)

PR #857, opened 2021-11-17
added basic code for query component

psarras:
## Description

This is related to #630. It offers an offline way to filter your data. I think it can continue being used even if the backend is fixed as in the discussion on #630, suggested from @teocomi . The benefit is that it will do the filtering offline rather than having multiple queries going through the server. Having said that fixing the backend to support queries would also be beneficial on other use cases

## Type of change

## How has this been tested?

## Docs

- Will be updated ASAP

image


psarras:
Hey @teocomi not a problem for us. Just to be clear this is the final implementation for this functionality. I don't think a new component would replace the need for this specific functionality.

PR #860, opened 2021-11-18
Revit/export material pull

thoffeller:
## Description
material handling for Revit. I'm sure there was a issue in the board, but it seems to be closed?
The pull request contains enhancements for Revit material handling. Currently there is the first material picked for export, while, f. ex. windows contain more than one material for the frame and glas. This ends up in wrong displayment in viewer and receiving applications.

## Type of change

- Bug fix

## How has this been tested?

- Manual Tests 
compiled the addin locally and tested transfers from Revit 2022 to speckle. It is not tested for Revit 2019 - 2021, I don't have this on my machine.

## Docs

- No updates needed
- Have already been updated
- Will be updated ASAP

### DIFF
--- a/All.sln
+++ b/All.sln
@@ -109,13 +109,17 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ConnectorETABS", "Connector
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ConnectorETABSv18", "ConnectorETABS\ConnectorETABSv18\ConnectorETABSv18.csproj", "{CC1F6133-9DD5-437C-AB23-380AD00267A2}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PolygonMesher", "Objects\Converters\StructuralUtilities\PolygonMesher\PolygonMesher.csproj", "{6185B1A1-0A12-44B1-8AC5-0ED48147FB21}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "PolygonMesher", "Objects\Converters\StructuralUtilities\PolygonMesher\PolygonMesher.csproj", "{6185B1A1-0A12-44B1-8AC5-0ED48147FB21}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "StructuralUtilities", "StructuralUtilities", "{85C30B55-E486-49AE-B3F5-6A196D14CCB8}"
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ConnectorETABSShared", "ConnectorETABS\ConnectorETABS\ConnectorETABSShared.shproj", "{1C661822-4524-4FAA-97F2-B5AAB9F9ECC4}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DesktopUI2", "DesktopUI2\DesktopUI2\DesktopUI2.csproj", "{AC0613B5-BD0E-4DA3-B704-7BE70D17AFFE}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ConverterRevit2022", "Objects\Converters\ConverterRevit\ConverterRevit2022\ConverterRevit2022.csproj", "{E08EC798-D1B9-4F72-A7CF-8611B5B9FF4C}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ConnectorRevit2022", "ConnectorRevit\ConnectorRevit2022\ConnectorRevit2022.csproj", "{DFDFDBB8-018B-4DCB-A012-54227ABF53A7}"
 EndProject
 Global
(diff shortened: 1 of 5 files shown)


teocomi:
Thanks Tim!!
We've set it for review in the next sprint, feel free to update the docs if needed in the menatime ;)

thoffeller:
Hi @teocomi , cool, thanks!
Are there any docs I need to update? ;-)

teocomi:
It might not be necessary actually, I'll let @JR-Morgan decide if a section inside [here](https://speckle.guide/user/revit.html) is worth ;)

barocuda:
Hi @JR-Morgan, is there any update on this?:)

JR-Morgan:
Hi @thoffeller, 

Sorry for taking a while to review this, It's now at the top of my priority list.
Thank you for your patience.

I just want to double-check we are on the same page about the motivation behind this PR.

As I understand it, the Revit converter currently exports the Revit material of objects using a [`RenderMaterial`](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Objects/Other/RenderMaterial.cs). (this is done [here](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs#L830-L846))
And the current structure/usage of this means that there is one `RenderMaterial` per object.
However, this leads to issues with objects like windows, where the glass and the frames should be a separate material.

So it sounds to me like this is the motivation behind this PR is to fix this issue of Revit elements with multiple materials getting converted to a speckle object with only a single material. Does that sound correct?

I'm a bit sceptical of using `Mesh.colors` vertex colors for this, for a few reasons that I will try and explain.
1. It leads to a lot of redundant duplicate data being stored. Which (although streams use compression) seems like a bit of a waste.
2. It makes interoperability with other connectors a bit of an issue. Many connectors don't support vertex colors, or don't support opacity with vertex colors.

For this reason, my preference would be to use `RenderMaterial`s if possible.

One alternate approach could be for us to use multiple `Mesh` objects in the displayMesh/displayValue of Revit elements, one for each material.

I'm curious as to your thoughts about this.
Thanks again for your patience, and for being a stellar community member and collaborator!

Jedd.

barocuda:
> Hi @thoffeller,
> 
> Sorry for taking a while to review this, It's now at the top of my priority list. Thank you for your patience.
> 
> I just want to double-check we are on the same page about the motivation behind this PR.
> 
> As I understand it, the Revit converter currently exports the Revit material of objects using a [`RenderMaterial`](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Objects/Other/RenderMaterial.cs). (this is done [here](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs#L830-L846)) And the current structure/usage of this means that there is one `RenderMaterial` per object. However, this leads to issues with objects like windows, where the glass and the frames should be a separate material.
> 
> So, it sounds to me, that the motivation behind this PR is to fix this issue of Revit elements with multiple materials getting converted to a speckle object with only a single material. Does that sound correct?
> 
> I'm a bit sceptical of using `Mesh.colors` vertex colors for this, for a few reasons that I will try and explain.
> 
> 1. It leads to a lot of redundant duplicate data being stored. Which (although streams use compression) seems like a bit of a waste.
> 2. It makes interoperability with other connectors a bit of an issue. Many connectors don't support vertex colors, or don't support opacity with vertex colors.
> 
> For this reason, my preference would be to use `RenderMaterial`s if possible.
> 
> One alternate approach could be for us to use multiple `Mesh` objects in the displayMesh/displayValue of Revit elements, one for each material.
> 
> I'm curious as to your thoughts about this. Thanks again for your patience, and for being a stellar community member and collaborator!
> 
> Jedd.

Hi @JR-Morgan,

Yes, the motivation of the PR is the supporting materials for Revit elements with multiple materials(Any family instance, walls with multiple layers, etc...). Currently first material from Revit element is retrieved and stored in one single ` RenderMaterial`  object.

We have used vertex colors to export materials, because currently the geometry of Revit elements which consist of nested meshes are merged into one mesh and stored in ` displayMesh` .

 We think also the best way to retrieve materials from Revit elements correctly would be exporting multiple meshes and mesh materials and storing multiples `displayMesh` and ` RenderMaterial`  objects and storing the RenderMaterial identifier for the meshes in displayMesh object.
 
 However, this would mean a breaking change for the frontend/viewer and for the converters of the all other connectors to receive objects and this must be also handled by other community members who currently consume the data from Speckle in current format.
 
Therefore, we have exported vertex colors and send ` RenderMaterial`  as before for the connectors which don't support vertex colors. I have also made some small changes locally to test this on speckle viewer. We could also create an another PR for the changes if you decide for this way.

We are looking forward to hear your opinion/decision about this.

didimitrie:
Hi Baris! I think we already support multiple meshes in the `displayValue` of an object, at least inside the viewer. I think other connectors would also need to handle this, if they don't already - we need to double check! 

Nevertheless, it seems like we all agree that going down the route of multiple meshes + each its own render material is the way to go 😎  - and we'll make other connectors follow suit, if need be. 

If you want to go ahead and implement this like that -  please do!

JR-Morgan:
Just an update, I've created an issue #964. And will attempt to implement the solution we discussed over the next few weeks.

Closing this PR for now, open to further discussion under #964.

PR #866, opened 2021-11-19
fix(revit): receiving blocks

### LINKED ISSUE #608: Revit: issue with displaymesh conversions inside generic family document
Noticed when trying to convert blocks with breps from rhino into revit, which does so in a generic model family document. These brep meshes convert fine when not sent as a block.

Breps that correctly fall back to display mesh conversion fail when being converted in the family context, where the tesselated shape builder fails to build the mesh:

[https://speckle.xyz/streams/5dfbeb49c9/commits/22e9c409cc](https://speckle.xyz/streams/5dfbeb49c9/commits/22e9c409cc)

Stack:
TessellatedShapeBuilder failed to build the requested geometry.
   at Autodesk.Revit.DB.TessellatedShapeBuilder.Build()
   at Objects.Converter.Revit.ConverterRevit.MeshToNative(Mesh mesh, TessellatedShapeBuilderTarget target, TessellatedShapeBuilderFallback fallback)

izzylys:
## Description

This will fix the mesh/brep block receiving issue in Revit to the best of our ability. Receiving block definitions now creates new Model Groups. New groups are placed for each instance of the block. If the block instance is scaled, the scaling is appended to the end of the definition name

#### issues
- curve blocks are not scaled! this is a bigger thing to tackle as curves all have diff defs and will require different procedures for scaling. maybe we'll want to extract this scaling logic separately into utility methods? not sure if it's useful for anything else tho 

Closes #608 

## Type of change

- Breaking change: blocks now come in as groups instead of families

## How has this been tested?

- Manual Tests with blocks from sketchup

## Docs

- Will be updated ASAP

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertBlock.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertBlock.cs
@@ -18,13 +18,14 @@ namespace Objects.Converter.Revit
     // Creates a generic model instance in a project or family doc
     public Group BlockInstanceToNative(BlockInstance instance)
     {
-
-      string result = null;
-
       // Base point
-      var basePoint = PointToNative(instance.GetInsertionPoint());
+      var basePoint = PointToNative(Point.FromList(new [] {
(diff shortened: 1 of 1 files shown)


izzylys:
closing this bc we don't care anymore lol

PR #880, opened 2021-11-23
Make streams private on their creation in Dynamo and Gh connectors

jenessaman:
## Description

- Default to making streams private when they are created with the Dynamo and Gh connectors (since all other connectors, through DUI/DUI2, currently default to making streams private on creation)

## Type of change

## How has this been tested?

- Manual Tests (created new streams with Dynamo and Gh connectors)

## Docs

- No updates needed


teocomi (approved):
Makes sense, thanks!

PR #890, opened 2021-11-26
GH: Reduce calls to `Tracker.TrackPageView`

### LINKED ISSUE #743: GH: Rethink tracking strategy to prevent over reporting
This is a bug we did not realise was happening:

Currently, the Tracker.TrackPageView method is being called on `BeforeSolveInstance`. On normal GH components this will result in double reporting of page views (and even more...)

I propose we move the tracking to the SolveInstance method, right before the component work is executed to ensure it only tracks actual runs performed.

AlanRynne:
## Description

- Fixes #743

Moved tracking from "BeforeSolveInstance" to the actual function that does the work (depending on the component, that may be `SolveInstance` or the `DoWork` in the worker.

Tracker will only log when the node performs some operation, as opposed to the previous behaviour, that allowed for pages to be tracked when `F5` was pressed, even if the node had no inputs connected.

## Type of change

## How has this been tested?

Set breakpoints on tracker lines and ensured they were:
- Never called twice in the same operation
- Never get called if the node does not perform an operation (invalid inputs, etc).

## Docs

- No updates needed

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/Accounts/Accounts.AccountDetails.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/Accounts/Accounts.AccountDetails.cs
@@ -48,7 +48,10 @@ namespace ConnectorGrasshopper.Streams
     {
       string userId = null;
       if (!DA.GetData(0, ref userId)) return;
-    
+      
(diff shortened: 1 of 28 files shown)


AlanRynne:
This PR is mostly harmless, so I'm just going ahead and merging it in.

PR #905, opened 2021-12-01
`Block` and `Transform` reworks

### LINKED ISSUE #608: Revit: issue with displaymesh conversions inside generic family document

izzylys:
## Description

This is a rework of Blocks to open the door for support in applications that don't properly support them (eg Revit).

TLDR: `Transform` is now a speckle object rather than a list of doubles representing the transform matrix. It allows you to apply the transform to points and vectors. Speckle objects can now implement `ITransformable` which involves writing a `TransformTo()` method. Meshes and some `ICurves` have been implemented as `ITransformable` and the Revit connector now supports them. The Revit connector does not truly support the concept of blocks, so new Groups are created for each `BlockInstance` with the transform baked into the source geometry. Note that this is p slow and could use optimisation.

- Fixes #608 indirectly

image

---

A new class `Objects.Other.Transform` as well as the interfaces `ITransformable` and `ITransformable<T>` have been added. The `Transform` class holds the matrix at `Transform.value` and contains helper methods for transforming points and vectors. The `ITransformable` interfaces can be implemented for any object that is transformable. You'll need to write a `TransformTo` method that returns a *new* object with the transformed values. `ITransformable<T>` is preferred as the return type of the `TransformTo` will match the current type, but `ITransformable` should be used in cases like curves where you need to check if an object implements `ITransformable` without knowing the type argument `<T>`. 

The only change to existing/future connectors is that the flat list of doubles representing the transform is accessed from `BlockInstance.transform.value` rather than `BlockInstance.transform`. This is a **⚠ BREAKING CHANGE** as the type of `BlockInstance.transform` is no longer a `double [ ]` but a `Transform`.

AutoCAD & Rhino have been updated to use the new transform:

image

image

## Type of change

## How has this been tested?

- Manual Tests with the usual test streams

## Docs

- Will be updated ASAP

### DIFF
--- a/Objects/Objects/Geometry/Brep.cs
+++ b/Objects/Objects/Geometry/Brep.cs
@@ -316,12 +316,18 @@ namespace Objects.Geometry
     public bool TransformTo(Transform transform, out Brep brep)
     {
       displayMesh.TransformTo(transform, out var mesh);
+      var surfaces = new List<Surface>(Surfaces.Count);
+      foreach ( var srf in Surfaces )
+      {
+        srf.TransformTo(transform, out var surface);
+        surfaces.Add(surface);
+      }
       brep = new Brep
       {
         provenance = provenance,
         units = units,
         displayMesh = mesh,
-        Surfaces = Surfaces,
+        Surfaces = surfaces,
         Curve3D = transform.ApplyToCurves(Curve3D, out bool success),
         Curve2D = Curve2D,
         Vertices = transform.ApplyToPoints(Vertices),
--- a/Objects/Objects/Geometry/Brep.cs
+++ b/Objects/Objects/Geometry/Brep.cs
@@ -309,18 +309,18 @@ namespace Objects.Geometry
     }
 
     /// <summary>
-    /// Returns a new transformed Brep. Note that `displayMesh` is *NOT* transformed. If you fall back on
-    /// the `displayMesh`, make sure to call `TransformTo` on the mesh object.
+    /// Returns a new transformed Brep. Note that `displayMesh` is *NOT* transformed.
     /// </summary>
     /// <param name="transform"></param>
     /// <returns></returns>
     public bool TransformTo(Transform transform, out Brep brep)
     {
+      displayMesh.TransformTo(transform, out var mesh);
       brep = new Brep
       {
         provenance = provenance,
         units = units,
-        displayMesh = displayMesh,
+        displayMesh = mesh,
         Surfaces = Surfaces,
         Curve3D = transform.ApplyToCurves(Curve3D, out bool success),
         Curve2D = Curve2D,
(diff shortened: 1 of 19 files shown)


Objects/Objects/Geometry/Brep.cs
```diff
@@ -306,6 +307,33 @@ internal void OnDeserialized(StreamingContext context)
...
+    /// </summary>
+    /// <param name="transform"></param>
+    /// <returns></returns>
+    public Brep TransformTo(Transform transform)
+    {
+      return new Brep
+      {
+        provenance = provenance,
+        units = units,
+        displayMesh = displayMesh,
```
clairekuang, line 323: Is there a reason why the displayMesh isn't transformed here as well?

Objects/Objects/Geometry/Brep.cs
```diff
@@ -306,6 +307,33 @@ internal void OnDeserialized(StreamingContext context)
...
       }
     }
+
+    /// <summary>
+    /// Returns a new transformed Brep. Note that `displayMesh` is *NOT* transformed. If you fall back on
+    /// the `displayMesh`, make sure to call `TransformTo` on the mesh object.
+    /// </summary>
+    /// <param name="transform"></param>
+    /// <returns></returns>
+    public Brep TransformTo(Transform transform)
```
clairekuang, line 317: Should consider redesigning these methods to return bool (or some other way of indicating a successful operation), since anything that implements `ApplyToCurves` can result in invalid geo which will be difficult to troubleshoot later.

Objects/Objects/Geometry/Brep.cs
izzylys: performance 🙃
it's really slow to transform large meshes

Objects/Objects/Geometry/Brep.cs
izzylys: that's a fair point! could we make an `out bool result` value? i think it's quite friendly for the method itself to return the transformed obj

izzylys: good point - refactor pushed! have a looksie ^^

clairekuang (approved):
Looks good to me!

PR #938, opened 2021-12-13
Support send/receive of `null` values in properties

### LINKED ISSUE #491: Support sending and receiving null values and properties

AlanRynne:
## Description

- Fixes #491

## Type of change

## How has this been tested?

Received [this commit](https://speckle.xyz/streams/cd004dafbc/commits/a1b7d00287) successfully in:

- [x] Grasshopper
- [x] Rhino
- [x] Revit
- [x] Dynamo
- [x] Autocad/Civil (@clairekuang could you check this?)
- [x] ~GSA (@Reynold-Chan could you check this and the next?)~
- [x] ETABS

## Docs

- Will be updated ASAP


Reynold-Chan:
that stream won't be able to be received in any structural connectors, since they're not structural classes... we don't have direct conversions implemented from non-structural kits yet. Sending seems fine though ? 

 Also, i don't think we're responsible for GSA connectors, since Arup is the one developing them? @jenessaman let me know how you want to approach this whether you want me to test this ?

AlanRynne:
Merging this in for `beta` testing prior to our last release of 2021! 🎄⭐️

If we regret this decision while testing we can always revert ♥️

Issue #780, opened 2021-10-14
Objects: Support for n-gons


izzylys:
Support n-gons or triangulation of them in mesh conversions on receive using the `n` as the first item in the polygon index array.

Once we have more complete support across connectors, we can consider using `3`/`4` indicators for outgoing faces in ToSpeckle conversion.
Currently, we are still using `0` and `1` for backwards/cross compatibility.

### Support for 3/4 faces
 - [X] Revit
 - [x] Viewer - see https://github.com/specklesystems/speckle-server/pull/486
 - [x] Rhino
 - [x] Blender
 - [x] Unreal
 - [x] Unity 
 - [x] Sketchup
 - [x] AutoCad/Civil3D
 - [x] Bentley
    
### Support n-gon ToNative conversion
 - [x] Revit - see https://github.com/specklesystems/speckle-sharp/pull/831
 - [x] Viewer - see https://github.com/specklesystems/speckle-server/pull/508
 - [x] Rhino - see https://github.com/specklesystems/speckle-sharp/pull/818
 - [x] Blender - see https://github.com/specklesystems/speckle-blender/pull/66
 - [x] Unreal - see https://github.com/specklesystems/speckle-unreal/pull/24
 - [x] Unity - see https://github.com/specklesystems/speckle-unity/pull/55
 - [x] Sketchup
 - [x] AutoCad/Civil3D - see https://github.com/specklesystems/speckle-sharp/pull/949
 - [x] [Bentley](https://github.com/specklesystems/speckle-sharp/issues/1061)
 
### Documentation
- [ ] Document n-gon support across connectors, and detail limitations/lossy conversions.

JR-Morgan:
Faces supported Pre PR:
0 (tris) and 1 (quads)
![image](https://user-images.githubusercontent.com/45512892/141192377-e35ad5a2-0fb3-4966-8f43-e4a12f3d03b9.png)

Post PR:
0 - 1, n
![image](https://user-images.githubusercontent.com/45512892/141192397-2cf2bd0f-3627-4f63-b899-40b8cb61d840.png)

JR-Morgan:
The Viewer, Unity, and AutoCad don't support N-gons/face triangulation natively, and so we would have to triangulate them during ToNative conversion.

I propose creating a triangulation function to be added to the `Objects` namespace. Then the AutoCad and Unity connectors can both use this function.
This function (or a similar one) can be created in JS to be used by the viewer.

[...]

PR #949, opened 2021-12-20
N-gon Triangulation

JR-Morgan:
## Description

Related to:  #780

Objects:
- Adds Triangulation functions for triangulating n-gon faces in connectors with no native support for n-gons or n-gon triangulation (functions are encapsulated in [`Objects.Utils.MeshTriangulationHelper`](https://github.com/specklesystems/speckle-sharp/blob/jrm/triangulation/Objects/Objects/Utils/MeshTriangulationHelper.cs))

AutoCAD:
- Fixes issue with receiving n-gons causing stream to fail
- Adds support for receiving n-gon faces. (using aforementioned triangulation)
- Adds support for receiving faces with 3/4 cardinality indicators. 

Rhino:
 - Switched N-gon triangulation to now use `MeshTriangulationHelper` as it is computationally faster and more consistent than the previous method of using `RH.Mesh.CreateFromClosedPolyline`

Unity:
 - Not part of this PR (or this repo) but Unity will eventually use this triangulation function as well.

## Type of change

- New feature, no breaking changes.

## How has this been tested?

Triangulation: Unit tests
- See [`MeshTriangulationHelperTests.cs`](https://github.com/specklesystems/speckle-sharp/blob/jrm/triangulation/Objects/Tests/Utils/MeshTriangulationHelperTests.cs)

AutoCAD & Rhino: Manual test
- Current method of testing is to receive N-gon test stream, send back to speckle, and observe JSON.
- Also tested with Rhino-test stream, and no regressions.
- Note that the "Torture cube" test mesh fails in AutoCAD due to PolyfaceMesh using `int16`s in `FaceRecord`s. Attempting to convert meshes with more than 32767 vertices/faces will lead to a highly corrupted native mesh.

Civil3D:
 - [x] TODO, I have done no tests in civil 3D, in theory shouldn't be any problems, but this should be tested.

## Docs

- N-gon support is currently undocumented. Documentation is part of #780.

### DIFF
--- a/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
+++ b/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
@@ -6,6 +6,7 @@ using System.Drawing;
 using Autodesk.AutoCAD.Geometry;
 using AcadGeo = Autodesk.AutoCAD.Geometry;
 using Autodesk.AutoCAD.DatabaseServices;
+using Objects.Utils;
 using AcadBRep = Autodesk.AutoCAD.BoundaryRepresentation;
 using AcadDB = Autodesk.AutoCAD.DatabaseServices;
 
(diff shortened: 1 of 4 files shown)


clairekuang (approved):
Skimmed through the triangulate methods & they are a bit beyond me! I trust you on these - tested in civil3d by receiving the beta ngon test stream and everything came in fine :)

PR #993, opened 2022-01-26
fix(dui2): added progress dialog window to sending and receiving from saved streams

### LINKED ISSUE #991: DUI2: Add cancellation button to send/receive process.
Pretty much self explanatory:

You can currently press the blue `Send/Receive` button, but there's no way to cancel the process once it's started.

clairekuang:
## Description

Adds a dialog window with progress bar and cancel button (same as used for quick send and quick receive) to sending and receiving from saved streams.

- Fixes #991

## Type of change

- 
## How has this been tested?

- Manual Tests: cancelled a few sends and receives

## Docs

- No updates needed

### DIFF
--- a/DesktopUI2/DesktopUI2/ViewModels/SavedStreamViewModel.cs
+++ b/DesktopUI2/DesktopUI2/ViewModels/SavedStreamViewModel.cs
@@ -38,7 +38,7 @@ namespace DesktopUI2.ViewModels
         this.RaiseAndSetIfChanged(ref _menuItems, value);
       }
     }
-
+    
     public string LastUpdated
     {
       get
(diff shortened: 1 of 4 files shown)


DesktopUI2/DesktopUI2/ViewModels/SavedStreamViewModel.cs
```diff
@@ -172,36 +171,52 @@ public void ViewOnlineSavedStreamCommand()
...
       Progress.IsProgressing = true;
+
+      var dialog = Dialogs.SendReceiveDialog("Sending...", this);
+
+      _ = dialog.ShowDialog(MainWindow.Instance).ContinueWith(x =>
+      {
+        if (x.Result.GetResult == "cancel")
+          Progress.CancellationTokenSource.Cancel();
+      });
+
```
teocomi, line 194: Hey @clairekuang I was thinking we could add a X button to cancel the operation inside the circular progress bar instead of showing a dialog. WDYT?

clairekuang: Hmm I think that would be better, but then we should try to align the quick send/receive button behavior as well

Issue #986, opened 2022-01-24
archicad: add a few additional conversions


izzylys:
refactoring the conversion logic to work with our current connectors was a lot bigger of a task than anticipated. as such, additional conversions were put on hold to prioritise receiving objects from different speckle connectors. this work is almost complete and further connectors will be tackled next sprint

Issue #985, opened 2022-01-24
archicad: debug receive of walls


izzylys:

izzylys:
fixed it! was a weird edge case w verts not being passed

Issue #984, opened 2022-01-24
archicad: improve build / debug / install flow


izzylys:
debug flow is a lot better - c# copy is automatic on build and there are vs code tasks for building and copying the c++ side. however, this only works if you've done it once manually before or if you run w elevated permissions bc this goes in the program files dir rather than the user dir.

- [ ] verify addon w key so it can be used in regular archicad mode
- [ ] identify any alternative addon paths
- [ ] provide an accessible user install for alpha (can simply be an exe that we post on the forum)

PR #1006, opened 2022-02-01
feat(archicad): merge into sharp

### LINKED ISSUE #1003: archicad: fix mesh structure so objs play nice in the viewer
this is fixed in the to-be merged archicad PR

izzylys:
big pr to integrate archicad repo into sharp

- restructures archicad repo putting the C++ addon and C# connector together within sharp
- updates the converter to properly set display value as list of meshes
- initial refactor of `Wall` model for simplification (removes the `WallData` to create a flatter obj similar to the other built elements)
- changes the way the avalonia application is being launched to align with the other connectors (need to test on mac if this still works ok)
- adds `task.json` files (vs code) to the projects to allow for simpler building, moving files, and launching archicad in demo mode 

closes #1003, #985, #984, #986


Objects/Objects/BuiltElements/Archicad/ModelData/Element/CeilingData.cs
```diff
@@ -0,0 +1,25 @@
...
+		public string edgeAngleType { get; set; }
+
+		public double? edgeAngle { get; set; }
+
+		public string referencePlaneLocation { get; set; }
+
+		public int? compositeIndex { get; set; }
+
+		public int? buildingMaterialIndex { get; set; }
+
```
Reynold-Chan, line 22: think we're missing a constructor here ?

Reynold-Chan (approved):

Objects/Objects/BuiltElements/Archicad/ModelData/Element/CeilingData.cs
Reynold-Chan: is it supposed to be 4 spaces or two spaces for indent ?

PR #1029, opened 2022-02-17
(core) fix operations send being stuck in an infinite loop if sending fails

gjedlicska:
## Description

- Fixes # (issue number or description)

## Type of change

## How has this been tested?

- Tests not required
- Unit/Integration Tests (which?)

## Docs

- No updates needed
- Have already been updated
- Will be updated ASAP


cristi8:
hmm, i think it gets a little more complicated, cuz when `SaveObject` is called, the transport is marked as `IsWriteComplete = false;` again.. and then it won't be set to true again, cuz the bg thread is stopped.

PR #1077, opened 2022-03-07
Adds inherit names feature in all `CSO`, `ESO` and `CSO by KV` nodes

AlanRynne:
## Description

Adds a new experimental feature to our custom inputs:

- New right-click menu option on each input parameter that allows for the name to be inherited based on it's connections.
  
- New shortcut `shift+ double-click` on any of the inputs to trigger the same behaviour.
  
- New shortcut `tab + drag connection into input` will rename the parameter as new connections are added.

This change affects the following components:

- Create Speckle Object
- Extend speckle Object
- Create Speckle Object by Key/Value
- Send

## Type of change

## How has this been tested?

Tested all use-cases I could think of, see gifs

## Docs

- Will be updated ASAP


AlanRynne:
✅ Reviewed with Dim, including the `KeyWatcher` solution to keep tabs of non-hotkey presses.

PR #1097, opened 2022-03-11
feat(revit): 1 click send

clairekuang:
## Description
Adds a 1 click button to Speckle toolbar in Revit.
Behavior: when clicked, finds or creates a stream named after the open file & adds the stream to the file if it isn't already saved. creates a commit from selected objects, or sends everything if no objects are selected. tracks the one click send in mixpanel, and opens the created commit in the web browser when finished.
- Fixes [#https://github.com/specklesystems/admin/issues/258](https://github.com/specklesystems/admin/issues/258)

## Type of change

## How has this been tested?

- Manual Tests 

## Docs

- Will be updated ASAP


teocomi (commented):
Looks good to me! Just two comments:
- why do you need to fire up the main window and hide it? Avalonia is initialized as Revit starts so it might not be necessary
- could be nice to show a minimal progress dialog when sending, you can reuse this one I made for the scheduled sends: https://github.com/specklesystems/speckle-sharp/blob/main/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit2/ConnectorBindingsRevit2.Events.cs#L77-L106

clairekuang:
> * why do you need to fire up the main window and hide it? Avalonia is initialized as Revit starts so it might not be necessary
'UpdateSavedStreams` is null unless an instance of the `HomeViewModel` is created, so this allows me to have the created file stream appear in DUI2 for the first time if a user opens a file and uses the 1 click send before clicking on DUI2. 

I'll add the progress dialog!

teocomi:
Cool, that works for now. Another way to do it is to directly update the stream saved in the Revit file, and only updating the one in DUI if the window is open: https://github.com/specklesystems/speckle-sharp/blob/main/DesktopUI2/DesktopUI2/ViewModels/SchedulerViewModel.cs#L107-L127 
But we can do that later!

PR #1106, opened 2022-03-15
adding material handling for mep elements. ducts, ductfittings, ducta…

### LINKED ISSUE #1099: Material in Revit for MEP objects
Hi specklers!
the new support for materials in connectors is fantastic! Thanks to @JR-Morgan and all in the team!
We need something similar for MEP elements. MEP elements itselve has no material. They end up in speckle in fatastic grey, regardless of the medium inside. Typically a freshwater supply pipe is green, a supply air duct is blue (all german standars, sorry) and so.
Currently most design compynies I knew implement this via display filters applied to specific views. I'm not sure if this is accessible via the api in revit.
The other option is the applied system. All MEP elements (except cable trays) have a system. This has a material. So we can suggest the MEP users to add valid materials to their models and they will end up with perfectly collored models in speckle.

Option 2, the system material approach, needs the modification for pipes, pipefittings, ducts, ductfittings, accessory and so to fetch the material from the system.

Now the question, do you want to implement it or shall we do it on our side and create a pull request?

Many greetings from sunny Germany,
Tim

thoffeller:
…ccessory, same for pipes and mep equipment is implemented and tested locally.

## Description

- Fixes #1099

## Type of change

- New feature 
support for revit materials in mep elements added. The material is taken from the assigned system and its type. So the user can control the coloring for ex. pipes accoring to local standards. In Germany cold water supply is green, while heating return is blue and so one.

## How has this been tested?

Tested locally in Revit 2022 and some local MEP projects. It depends on the project settings and materials used.

## Docs

- No updates needed

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -850,7 +850,8 @@ namespace Objects.Converter.Revit
       }
     }
 
-    public RenderMaterial GetElementRenderMaterial(DB.Element element)
+        #region materials
+        public RenderMaterial GetElementRenderMaterial(DB.Element element)
     {
       var matId = element.GetMaterialIds(false).FirstOrDefault();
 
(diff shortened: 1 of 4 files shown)


JR-Morgan:
Hi Tim!
Thanks for the PR, I've tested my end and looks great! 🙌 

I'm happy to merge this in (if you don't mind, I might clean up the `IsSupportedMEPCategory` method a little)
But quick question, What was the reason you are caching the `Categories` in a field?

https://github.com/specklesystems/speckle-sharp/blob/dd64ebb424a95c75d1a8872ec62270891b06ecd2/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs#L949-L953

I think this might cause an issue when sending objects from different docs.

Thanks for your contributions

JR-Morgan (approved):

JR-Morgan:
Merged via https://github.com/specklesystems/speckle-sharp/pull/1142

Issue #1100, opened 2022-03-11
Revit MEP fabrication parts support for connector


thoffeller:
Hi specklers,

currently the mep fabrication parts are nor supported in the connector. Is there a plan to add them for export to speckle?
If not, shall we implement the parts and create a pull request?

Sunny greetings from Germany,
Tim

teocomi:
Hey Tim, we don't have short-term plans to pick those up, so it'd be great if you could do a PR!

thoffeller:
Hey @teocomi , added the support and send a pr. Hope it is, what you expected?
Cheers,
Tim

thoffeller:
Ahh, it seems that are elements missing if the fabrication parts handled that way... In a bigger project I saw missing mep parts in a system. They are exported when I use the current master or my modifications in materialsformep branch. So, I think this isn't the perfect solution yet...

thoffeller:
Hi @teocomi,  the fabrication issue pops up again. Before we implement it we should agree an the best way to do it.
- Shall we add the fabrication parts 'below' the design elements handling?
- shall we create separate converters for those parts?
-
Not sure what you would prefer and what is the best way in a long way.
If it makes sense, we can have a short call via teams? 

Thanks, Tim

teocomi:
Hey @thoffeller I haven't worked with fabrication parts before, happy to jump on a quick call and see what's the best way to go about them.
Cc @connorivy

thoffeller:
Hi @teocomi, cool, how about next Monday, 11 am german time? Shall I send an teams invite? Thanks, Tim

barocuda:
Hey @teocomi,

As you discussed with @thoffeller,here is the way we would implement the export of fabrication parts in two steps without changing anything in existing code:

`  public RevitElement FabricationPartToSpeckle(FabricationPart revitElement, out List<string> notes)
        {
            notes = new List<string>();

            RevitElement speckleElement = new RevitElement();

            speckleElement.type = revitElement.Name;

            speckleElement.category = revitElement.Category.Name;

            speckleElement.displayValue = GetFabricationMeshes(revitElement);

            //Only send elements that have a mesh, if not we should probably support them properly via direct conversions
            if (speckleElement.displayValue == null || speckleElement.displayValue.Count == 0)
            {
                notes.Add("Not sending elements without display meshes");
                return null;
            }

            GetAllRevitParamsAndIds(speckleElement, revitElement);

            return speckleElement;
        }`
        
[...]

PR #1108, opened 2022-03-16
Revit/fabricationparts

thoffeller:
## Description

Hey there, this is an implementation for export support of Revit fabrication elements. It is implemented on top of the generic exporter, so no additional, separate exporter is added.
based on issue #1100

## Type of change

- New feature 
Adds support for fabrication elements. Other stuff should not be affected.

## How has this been tested?

Tested it locally on my machine with Revit 2022 and fabrication elements inside revit projects

## Docs

- No updates needed


thoffeller:
As stated in #1100 , the solution skips some elements during export. Needs to be revised.

PR #1156, opened 2022-03-31
GH/RVT: Add support for wall openings

### LINKED ISSUE #970: GH BIM: add support for wall openings
Add support for wall openings in the GH BIM nodes like for floors.
See https://speckle.community/t/gh-surface-revit-wallbyface-no-stream-generated/2214/5

AlanRynne:
## Description

- Fixes #970

## Type of change

 ### Breakdown

Added new Schema nodes for `Opening` and `RevitWallOpening`.
Modified the wall opening conversions to throw better errors (it assumed curve was polyline, but that's not necessarily the case)
Improved edge-case where host element was set in the opening, but opening was not part of the wall's elements list.

If a RevitWallOpening is contained in the `elements` list of a wall **and** has it's host set to the wall, the converter will prioritize the parent (i.e.: the object owning the element list). Only when there is no set object we will  try to find the specified host in the document.

If the host is not found, the operation is aborted.

## How has this been tested?

Manually sent/received walls with openings and openings alone.

## Docs

- Will be updated ASAP

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/ConnectorGrasshopper.csproj
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/ConnectorGrasshopper.csproj
@@ -29,8 +29,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Grasshopper" Version="6.32.20340.21001"  IncludeAssets="compile;build" />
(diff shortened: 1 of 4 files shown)


teocomi (approved):
LGTM! Haven't run it locally though :)

PR #1165, opened 2022-04-04
Small fixes to server transport

### LINKED ISSUE #1031: Has Objects check hangs if streamId is null

### LINKED ISSUE #1049: Core: Revise the async mechanisms
We have a number of parallel mechanisms in place for faster serialization and server communication.

We should make sure the appropriate async mechanisms is used in each case.

In particular:
 - calling `.Result` on a task will block the calling thread (used in a couple of places, make sure it's intended)
 - ensure (de)serialization context switching and waiting is done properly
- never block the calling thread of operations.receive / etc  (might be UI thread) (issue described here: https://speckle.community/t/speckle-core-cant-receive-stream-data-on-specific-machine/2409/7 )

This ticket is a result of multiple related issues / questions.

We should also briefly document somewhere the parallelism involved and how it's done.

### LINKED ISSUE #1113: [Core] Retry failed connections to the server
If the speckle server returns with 503 for a connection the client should gracefully retry with a sane retry policy

cristi8:
## Description

- Fixes #1031
- Fixes #1113
- Fixes #1049

## Type of change

## How has this been tested?

- Manual Tests

## Docs

- No updates needed

### DIFF
--- a/Core/Core/Transports/ServerUtils/ParallelServerAPI.cs
+++ b/Core/Core/Transports/ServerUtils/ParallelServerAPI.cs
@@ -57,6 +57,7 @@ namespace Speckle.Core.Transports.ServerUtils
       for (int i = 0; i < NumThreads; i++)
       {
         Thread t = new Thread(new ThreadStart(ThreadMain));
+        t.Name = $"ParallelServerAPI";
         t.IsBackground = true;
         Threads.Add(t);
         t.Start();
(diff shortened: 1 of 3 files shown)


didimitrie (approved):
Only comment would be to, at one point, comment these parts out - we're deep in dragon territory here :D (approving without really understanding the changes, tbh!)

Issue #1145, opened 2022-03-28
Rhino: Conversion of PBR properties leads to very different looking materials.


JR-Morgan:
Current implementation of Rhino material conversion is very lossy and often leads 
to materials looking very different in the viewer, particularly with Reflectivity --> metallic conversion.

We may need to rethink how we define Render Materials to better support Metallic and Specular workflows https://marmoset.co/posts/pbr-texture-conversion/

- [x] ToSpeckle - https://github.com/specklesystems/speckle-sharp/pull/1180
- [ ] ToNative -

paloknapo:
Closing as stale.

PR #1180, opened 2022-04-05
Rhino 7 will now output the correct PBR material properties

JR-Morgan:
## Description

This issue:
 - Rhino materials often don't follow a Physically Based shading workflow. And yet, materials have PBR like properties.
 - Previously, our ToSpeckle conversion was trying to use these properties, and also trying to fit them into our "metalic/roughness" workflow, which lead to some highly undesirable results (like "plaster" and "plastic" materials having a metalness of 100%)
- See https://speckle.community/t/speckle-2-4-0-connector-changes-enhanced-material-support/2442/9

The Fix:
- In Rhino 7, there are functions exposed for converting Rhino preset materials into Rhino PBR materials. This PR updates ToSpeckle conversion to use these. This also removes the need for the hacky opacity work around for glass/gem material (atleast, just for Rhino 7)
- For Rhino 6 API, these functions are not available, so the best we can do is just to ignore metalness and only convert base color, transparency, and emissivity. This will still be somewhat lossy.
- This solves the ToSpeckle part of https://github.com/specklesystems/speckle-sharp/issues/1145

What isn't fixed:
- For ToNative conversion, and `metalness`/`roughness` properties are ignored.
- For Rhino 7, we can in theory convert materials to `PhysicallBasedMaterial`s. I did have a quick go at implementing this, but unsuccessfully. Since this is lower priority, this will can be fixed later. (tracked in https://github.com/specklesystems/speckle-sharp/issues/1145)

### Screenshots!
Ground Truth (cycles)
image
image

---

Before this PR:
image

---

After this PR:
image

## Type of change

- Bug fix - Fixed bad conversion logic of `metalness` property.
- New feature - (Rhino 7) added proper PBR ToSpeckle conversion logic.

## How has this been tested?

- Rhino 6 is untested (because, in theory the code is unchanged, and it builds fine)
- Rhino 7 has been tested with the Rhino test stream, PBR materials stream, and then sending a few random preset materials.

## Docs

- No updates needed
unless we want to mention Rhino 6's limitation?

### DIFF
--- a/Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Other.cs
+++ b/Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Other.cs
@@ -21,6 +21,7 @@ using Text = Objects.Other.Text;
 using RH = Rhino.DocObjects;
 using RenderMaterial = Objects.Other.RenderMaterial;
 using Rhino;
+using Rhino.Render;
 
 namespace Objects.Converter.RhinoGh
 {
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Other.cs
```diff
@@ -66,36 +67,71 @@ public Rhino.Render.RenderMaterial RenderMaterialToNative(RenderMaterial speckle
...
+      Material matToUse = material;
+      if(!material.IsPhysicallyBased)
+      {
+        matToUse = new Material();
+        matToUse.CopyFrom(material);
+        matToUse.ToPhysicallyBased();
+      }
+      using (var rm = Rhino.Render.RenderMaterial.FromMaterial(matToUse, null))
+      {
+        RH.PhysicallyBasedMaterial pbrMaterial = rm.ConvertToPhysicallyBased(RenderTexture.TextureGeneration.Allow);
```
clairekuang, line 126: Did you check if this conversion to a physically based material returns the correct opacity? Just in case this means we can get rid of the transparency hack in line 114

JR-Morgan: From Rhino 7, all properties including opacity are correct. I've just added screenshots to my PR. 

That hack is only necessary for Rhino 6, which doesn't support any of the Physically Based functions we are using here.

PR #1190, opened 2022-04-11
Revit: Fixed references to document field in ToSpeckle conversion functions

JR-Morgan:
- Fixes issue with sending linked document from Revit
- There were still a few places where we were using the main (current) document when converting elements ToSpeckle.
- For elements in linked documents, this was leading to a variety of null reference exceptions etc.

- Also fixed an issue with sending family Instances that use a `LocationCurve` instead of a `LocationPoint`


teocomi (approved):
Txs for finding these @JR-Morgan !

PR #1217, opened 2022-04-25
Fixes 2.5.1 GH crash

AlanRynne:
## Description

Fixes GH crash reported on the forum: https://speckle.community/t/grasshopper-with-speckle-connection-crashes/2795

This was caused by the GH build not properly renaming the `dll` to `gha`,  causing it to either be missing, or not override the existing one.
- If it didn't override the existing one, it would cause a crash.
- If it was named `dll` instead of `gha`, Speckle would not be loaded.

## Type of change

## How has this been tested?

Tested both mac and rhino builds to ensure files where compiled correctly and connectors ran as expected.

## Docs

- No updates needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,6 @@ jobs: # Each project will have individual jobs for each specific task it has to 
           name: Build DesktopUI2
           command: dotnet build DesktopUI2/DesktopUI2.sln -c Release -v q
 
-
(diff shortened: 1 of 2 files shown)


AlanRynne:
Mac CI fails due to partial outage on Github Package Registry (used to install mono). Merging in as i've checked locally.

PR #1262, opened 2022-05-18
Updated ToSpeckle conversion to use 3/4 Mesh face indicator 

JR-Morgan:
All of our connectors have supported ToNative 3/4 indicators for many releases.
But for the sake of a transition period we kept 0/1 around for a while.
Now, for the sake of simple documentation, we can finally switch to using 3/4 indicators the for all ToSpeckle conversion.

This PR should be merged as part of our next release.
Needs testing in all connectors as per usual release testing.


JR-Morgan:
We may want to delay merging this pr for another release. @AlexandruPopovici found a [bug](https://github.com/specklesystems/speckle-server/commit/fbb370191797d36aeb85659842f2d56dca39d9ec) in the viewer that means previous server releases won't be forwards compatible.

AlanRynne:
Dynamo was just updated to use 3,4 mesh indices on #1430

PR #1263, opened 2022-05-18
fix(core): Made Report thread safe :)

AlanRynne:
## Description

- Fixes issue reported in the Forum where Expand Speckle Object (now Deconstruct Speckle Object) and some others would fail on the `BeforeSolveInstance` call due to accessing the Conversion Report lists in a non thread-safe way.

We are now locking the access to the list so that this no longer happens.

## Type of change

This change should not break any of our existing connectors, as they all work in sequence. The only connector where this could possibly happen was Grasshopper, with the introduction of TaskCapableComponents.

## How has this been tested?

The issue is difficult to reproduce, but I managed to reproduce it in my machine quite qickly by just generating a high number of inputs.

Once the fix was in place, the error no longer happened.

## Docs

- No updates needed

### DIFF
--- a/Core/Core/Models/Extras.cs
+++ b/Core/Core/Models/Extras.cs
@@ -100,43 +100,50 @@ namespace Speckle.Core.Models
     /// </summary>
     public List<string> ConversionLog { get; } = new List<string>();
 
+    private readonly object ConversionLogLock = new object(); 
     public string ConversionLogString
     {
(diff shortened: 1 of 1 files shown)


AlanRynne:
Tested in Rhino, Revit and GH. Report continues to work as expected and can no longer reproduce the problem in GH

PR #1270, opened 2022-05-24
GH: Convert Data T

### LINKED ISSUE #1252: Change tree definitions from nested lists to dict["{0;0;0}"] = [...], but keep backwards compatibility
This is not the first time we attempt to do this 😅 But it's been requested again by Izzy and Claire, as it is quite confusing to get all these "seemingly random" empty lists being generated.

It does complicate quite a bit the exploration of the data in the frontend without much need for it.

Only person I remember being against this is @teocomi, so we may need to convince him. His hesitation came mostly from Dynamo, as Dicts are treated differently, it could become a bit messy.

We definitely **do not want** to expose the data tree notation in Dynamo. Instead, ideally it should be converted back into a list of lists (default behaviour).

For any other connector, I think its a safe change that would actually reduce the amount of iterations we do to traverse a "DataTree".

Backwards compatibility for nested lists to data tree conversion **needs to remain**, not only for backwards compatibility, but also for any other object that could contain nested lists (such as, for example, a matrix, an data-only (no headers) CSV file...

AlanRynne:
## Description

- Fixes #1252 

## Type of change

- Breaking change:

Modifies DataTree conversion from nested lists to be a Base object with branch paths as keys.

## How has this been tested?

Tested GH<->GH, DYN<->GH.

## Docs

- Will be updated ASAP

### DIFF
--- a/ConnectorDynamo/ConnectorDynamoFunctions/BatchConverter.cs
+++ b/ConnectorDynamo/ConnectorDynamoFunctions/BatchConverter.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Speckle.Core.Logging;
 
 namespace Speckle.ConnectorDynamo.Functions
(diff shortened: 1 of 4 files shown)


AlanRynne:
Notified the community, will give some time to gather replies before merging this in.

PR #1274, opened 2022-05-24
RVT: Improved polyline conversions

AlanRynne:
## Description

- Fixes issue in Revit where Polyline's with very short segments would completely fail to convert.
- Fixes an issue were we were assuming a `Curve` in a cast, but should have been an `IDisplayable<Polyline>` instead.
- Adds length calculation to `Line` class constructor. This is used during polyline creation to ensure no short segments are converted to Revit.

## Type of change

## How has this been tested?

Original offending commit: https://speckle.xyz/streams/35cc66339a/commits/cb564be587

## Docs

- No updates needed

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertGeometry.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertGeometry.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
+using Autodesk.Revit.ApplicationServices;
 using Arc = Objects.Geometry.Arc;
 using Curve = Objects.Geometry.Curve;
 using DB = Autodesk.Revit.DB;
@@ -458,7 +459,7 @@ namespace Objects.Converter.Revit
       var nurbs = DB.NurbSpline.Create(spline);
       return NurbsToSpeckle(nurbs, units ?? ModelUnits);
     }
-
+    
     public CurveArray PolylineToNative(Polyline polyline)
     {
       var curveArray = new CurveArray();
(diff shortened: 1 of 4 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertGeometry.cs
```diff
@@ -468,18 +469,37 @@ public CurveArray PolylineToNative(Polyline polyline)
...
+          {
+            Report.LogConversionError(e);
+          }
         }
 
         if (polyline.closed)
         {
-          var speckleLine = new Line(new double[] { pts[pts.Count - 1].x, pts[pts.Count - 1].y, pts[pts.Count - 1].z, pts[0].x, pts[0].y, pts[0].z }, polyline.units);
-          curveArray.Append(LineToNative(speckleLine));
+          var speckleLine = new Line(pts[pts.Count - 1], pts[0] , polyline.units);
```
clairekuang, line 494: Should handle the case here where the polyline first and last points are the same and `.closed() = true`
The new Line may throw an exception in this case?

izzylys: fair, tho in theory this shouldn't happen ((the polyline class asks for first and last points to be unique tho this is not enforced so someone could easily do this if they don't read the docstring))

AlanRynne:
Since the same logic was being used in 3 parts of the same method, I refactored it out into the ConversionUtils file.

I also added some XML comments to the new functions, and to the `PolylineToNative` to make it clear that the geometry can be modified when small segments are present.

Screenshot 2022-05-25 at 22 04 09

izzylys (approved):
nice! I like the refactoring the logic out to the utils and I appreciate your helpful xml comments 😊

PR #1314, opened 2022-06-15
Add materials and quantities of materials to RecitConnector

### LINKED ISSUE #1302: Enhancement of ConverterRevit: Extracting material quantities and properties (incl. multi-material elements)
**Problem:** The current Revit converter does not allow for retrieving quantities from an element consisting of multiple materials.

**Example:** A multi-layered wall does only have 1 Volume. However, for analysing quantities for cost-estimation or LCA we need exact quantities of each material.

**Enhancement:** I propose adding an object to the Base-object generated in *ConvertToSpeckle* in the *ConverterRevit*. It needs to be discussed how this quantity should be structured. It could either have a set of the materials with all their parameters and the quantities (volumes, area). Or - to begin with - just the material-id and quantities.

Please let me know what you think. I would be happy to contribute.

Thanks,
Kilian

ks-cph:
## Description

- Fixes #1302 
- Adds all Materials of a Revit project to the stream
- Computes quantities for each material of each element
- Please find a rough UML of the proposed changes

## Type of change

- Generic classes *Material* and *MaterialQuantity*
- Converter for *Material* and *MaterialQuantity* for the RevitConnector:
- Allows for converting Materials (RevitConnector)
- Computes quantities of each material for every element (RevitConnector)
- *Material* interferes with *Autodesk.Revit.DB.Material* -> I added references to some files

## How has this been tested?

- Manual test in Revit 2020 and Revit 2022
- https://speckle.xyz/streams/a70f6daf00 (RVT2020)
- https://speckle.xyz/streams/f78ac8e10c (RVT2022)

UML_speckle-sharp

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterial.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterial.cs
@@ -1,0 +1,32 @@
+﻿using Autodesk.Revit.DB.Structure;
+using Speckle.Core.Kits;
+using Speckle.Core.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using DB = Autodesk.Revit.DB;
+
+namespace Objects.Converter.Revit
+{
+    public partial class ConverterRevit
+    {
+        public Objects.Other.Material MaterialToSpeckle(DB.Material revitmaterial)
+        {
+
+            var speckleMaterial = new Objects.Other.Revit.RevitMaterial(revitmaterial.Name, revitmaterial.MaterialCategory, revitmaterial.MaterialClass, revitmaterial.Shininess,
+                revitmaterial.Smoothness, revitmaterial.Transparency);
+
+
+            GetAllRevitParamsAndIds(speckleMaterial, revitmaterial);
+
+            Report.Log($"Converted Material{revitmaterial.Id}");
+            return speckleMaterial;
+        }
+
+        public List<ApplicationPlaceholderObject> MaterialToNative(Objects.Other.Material speckleBeam, StructuralType structuralType = StructuralType.Beam)
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+}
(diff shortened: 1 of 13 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -0,0 +1,91 @@
...
+
+
+            var speckleMaterial = ConvertAndCacheMaterial(material.Id, material.Document);
+            return new Objects.Other.MaterialQuantity(speckleMaterial, volume, area);
+        }
+        public List<ApplicationPlaceholderObject> MaterialQuantityToNative()
+        {
+            //To-Do: Is this needed?
+            throw new System.NotImplementedException("Missing MaterialQuantity to Native");
+        }
```
teocomi, line 59: This is not needed, as long as there is no "CanConvertToNative" entry ;)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterial.cs
```diff
@@ -0,0 +1,44 @@
...
+            GetAllRevitParamsAndIds(speckleMaterial, revitmaterial);
+
+            Report.Log($"Converted Material{revitmaterial.Id}");
+            return speckleMaterial;
+        }
+
+        public List<ApplicationPlaceholderObject> MaterialToNative(Objects.Other.Material speckleBeam, StructuralType structuralType = StructuralType.Beam)
+        {
+            throw new NotImplementedException();
+        }
```
teocomi, line 29: Can be skipped as well :)

Objects/Objects/Other/MaterialQuantity.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+
+namespace Objects.Other
+{
+    public class MaterialQuantity : Base
+    {
+        public Objects.Other.Material material { get; set; }
+        public Objects.BuiltElements.Revit.Parameter volume { get; set; }
+
+        public Objects.BuiltElements.Revit.Parameter area { get; set; }
```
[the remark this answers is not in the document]
ks-cph: Thanks for the review @teocomi 

I chose `Objects.BuiltElements.Revit.Parameter` because (1) area and volume have different units (2) there is no generic `Parameter` (3) I do not fully understand the concept of `Core.Core.Kits.Units`.

`Core.Core.Kits.Units` only covers 1d units. Or does - for instance - *feet* also imply that the units are *squarefeet* and *cubicfeet*

PR #1329, opened 2022-06-17
fix(dyn): Align isDataTree condition with GH fix on #1320

AlanRynne:
Aligns DYN with fix made in #1320


AlanRynne:
Merging in, disregarding failed CI attempt, as this is a one liner fix.

Issue #1249, opened 2022-05-10
Autocad: annotation conversion laundry list


clairekuang:
With the goal of making the autocad connector more useful, we should support the following additional conversions:

- [ ] leader
- [ ] dimensions
- [ ] leader / multileader
- [ ] table

The following conversions should be checked and improved:

- [ ] hatches
- [ ] text / notes

paloknapo:
Closing as stale.

PR #1348, opened 2022-06-27
feat(autocad): adds dimension conversions

clairekuang:
## Description
Adds a `Dimension` class to objects
Adds `Dimension` conversions to and from Autocad
- part of #1249 

## Type of change

## How has this been tested?

- Manual Tests: sent and received a few different dimensions type in autocad

## Docs

- Will be updated ASAP


izzylys (approved):
yay for fully documented objects classes ✨

PR #1369, opened 2022-06-30
Traversal Optimisation and unit tests

JR-Morgan:


AlanRynne (commented):
This is awesome :)

The extra fix for the non-authenticated account was a good catch!

PR #1380, opened 2022-07-05
Replaced StreamReader.Peek()

FerdiMeijer:
Replaced Peek() function because of issue that occurs only in .net6.0 when downloading large object (e.g. 8000+ child objects) using the Speckle "ServerTransport"

## Description
Because of a .net6.0 issue with StreamReader.Peek() see: https://github.com/dotnet/runtime/issues/68983
The ServerTransport is using the .Peek() function to find the end of the stream and stop reading more object lines.

## Type of change

## How has this been tested?

-The changes to the Speckle Transport have been tested in a separate console app tool referencing the Speckle.Core project. It also has been tested as a Copy/CustomTransport supplied to the Operations.Receive in various production/test environments as a quick fix, it has not been tested as part of a development build of Speckle.Core 2.6.x

## Docs

- No updates needed

### DIFF
--- a/Core/Core/Transports/Server.cs
+++ b/Core/Core/Transports/Server.cs
@@ -479,7 +479,8 @@ namespace Speckle.Core.Transports
       {
         using (var reader = new StreamReader(stream, Encoding.UTF8))
         {
-          while (reader.Peek() > 0)
+          string line;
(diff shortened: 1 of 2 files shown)


JR-Morgan (approved):
Tested and approved!

PR #1381, opened 2022-07-05
fix(rhino): race condition at startup when app is null

teocomi:
## Description

- Fixes issue at startup when the Application is still null, could only replicate while debugging so far


AlanRynne (approved):
LGTM

PR #1382, opened 2022-07-05
[FIX] Brep deserialisation issue

AlanRynne:
## Description

- Fixes Brep deserialisation issue (can't find the issue now, TBD)

Core reason behind it has that all the Brep Edges, Faces and Loops were being serialised as chunks of objects, leading to **chunks with the same id** being deserialised as the "same chunk instance", which is an unwanted side-effect.

This issue has survived quite a while under the radar, as it is a difficult thing to catch, and in theory **is not a bug on the chunking or deserialisation** logic, as Chunks were designed to hold lists of primitives that will be merged together on deserialisation.

In this case, we were using it for to store chunks of **reference types**, which made everything work in unpredictable ways, and is not the intended use-case for Chunks.

Hence, we now serialise/deserialise all brep fields as lists of primitives. **This change is backwards compatible**

## Type of change

This change does not break serialisation/deserialisation for older commits, but it also **does not fix the issue on older Breps**. The new format and the old are interchangeable, but follow different logic upon deserialisation.

Any BREP that was serialised with its Faces, Edges or Loops as Chunks will continue to have the same issues.

## How has this been tested?

Sent/Received breps from Rhino to Revit, GH to Rhino, Rhino to Gh and Gh to Revit, all seem to successfully convert (except the cases we know are not supported by Revit, i.e. closed nurbs curves)

## Docs

- No updates needed, but will talk to @NamitaSingh00  about broadcasting this, maybe alongside the DXF one too?


teocomi (dismissed):
Cannot comment on the implementation per se, but I approve the change!

AlanRynne:
Had removed the `builder.AllowRemovalOfProblematicFaces` while testing and it got staged into one of my commits by accident. Re-adding it, as it is needed for successful Brep conversion in edge-cases.

clairekuang (approved):

PR #1385, opened 2022-07-06
RH-GH: Unified Rhino/Gh solution.

### LINKED ISSUE #1388: Unify Grasshopper and Rhino Installers
Unify Grasshopper and Rhino installers into a single "product".

This change is to prevent any user from installing different versions of Rhino/Gh, which will lead to errors in either side.

The cause of this was that Rhino will always load the Rhino Connector dll files first, and then GH will use whatever Rhino loaded already. Meaning that, if Rhino had a version with breaking changes in core, GH will fail to load.

This is the main cause for the famous "No default kit found" error that has been haunting us forever.

## Implementation

Rhino will automatically add any Rhino Plugin path to the "Grasshopper Developer Settings" paths, allowing Grasshopper to discover any `.gha` files located within the plugin folder itself.

For this to work, we'd just need to add `ConnectorGrasshopper` as a direct dependency of `ConnectorRhino6/7`, so that the `.gha` get's copied over, as well as any other dependencies such as `GHAsyncComponent`.

Any other shared dependencies (Core...) will be taken care of directly by the Rhino connector project build, ending in both plugins (Rhino and Grasshopper) having all their necessary dependency `dll` files in one place.

## How does this fix it?

Since the `dll` files are now **the same files in the same location**, we no longer need to care about Rhino loading an incorrect one.

It will also no longer be possible to have different versions of Rhino and Grasshopper, as they'll be packaged in the same Installer.

AlanRynne:
## Description

- Fix #1388 
- Allows for debugging of Rhino and Gh at the same time.

This PR is still missing:

- [x] Remove Grasshopper jobs from CI
- [x] Coming up with a decent strategy for releasing this new change
- [x] Updating ISS files to properly reflect this new dependencies
    - Add GH converters to Objects dependencies in Rhino installer
    - Ensure GH files are being copied too
- [x] Testing the entire install process alongside manager changes that may be required. 
- [ ] [OPTIONAL] Move all Grasshopper projects into the Rhino solution folder too?

This change is to prevent any user from installing different versions of Rhino/Gh, which will lead to errors in either side.

The cause of this was that Rhino will always load the Rhino Connector `dll` files first, and then GH will use whatever Rhino loaded already. Meaning that, if Rhino had a version with breaking changes in core, GH will fail to load.

This is the main cause for the famous "No default kit found" error that has been haunting us forever.
### Installer changes:

As discussed with @teocomi, the final solution we've come up with for the Installers is the following:

- We'll continue to have two installers (rhino and grasshopper), but both installers will actually be **the same installer with different names**. They will point to the same `AppId` and `Uninstaller` folder, making them interchangeable.
- We'll change Grasshopper's `AppId` in Ghost so that it is now the same as Rhino's.
- We'll keep manager's separate Rhino and Grasshopper pages for discoverability, tutorials.

Once that's done, in theory if you install Grasshopper, manager will also show Rhino as installed, and vice-versa. Preventing the user from trying to install both.

## Type of change

The way this has been finally done, is by just adding the `ConnectorGrasshopper` as a dependency of `ConnectorRhino6` and `7`.

For this to work, I had to modify the `GrasshopperConnector` to align with the SDK style of the Rhino connectors.

Since we're no longer using this specific connector as a "standalone install", this should not be a problem.

## How has this been tested?

- [x] Remains to be tested, it has been mentioned above ☝🏼

## Docs

- [x] ~Updates will be needed on the website too, since now 2 connectors are becoming 1.~ No updates needed on the website, other than maybe a brief mention that they're actually installed "at the same time"


clairekuang (dismissed):

AlanRynne:
Bypassing re-review by @clairekuang, since none of my last changes have any effect on the code.

PR #1397, opened 2022-07-12
Revert "feat(dui): updates to Avalonia 0.10.16"

AlanRynne:
This reverts commit 9d302eec86543dffd2f18569b9d7fef8cd0cabca.

This is being reverted because it has been already squashed into the `release/2.7` branch, and should not live in `main` unless we're planning to release this as a hotfix (which I think may be dangerous, this should be tested first).

Changes are here b2e0ec8aa56ed37b4df186ee4263dc92008c5b0b


izzylys (approved):
pls go ahead!

teocomi (approved):

PR #1398, opened 2022-07-12
fix: Avalonia version bump was missing 2023 Acad and Civil projects

AlanRynne:
## Description

- Bumps version of Acad and Civil 2023 projects to Avalonia 0.10.16
- 
## Type of change

Potentially breaking in some ways, requires further testing in the release.


teocomi:
Ok, sounds like these projects were missing from `All.sln`

teocomi (approved):

PR #1401, opened 2022-07-14
assign value to variable 'existingLevelWithSameElevation'

connorivy:
## Description

- Fixes https://speckle.community/t/speckle-revit-levels-bug-old-levels-not-replaced/2378/7

Currently, a level is being created for each imported object. In the gif below you can see that I import a commit and then highlight what seems to be only a few levels
imports-level-for-each-obj

The text may be difficult to read in the gif, but Revit is showing that I actually have 104 levels highlighted.

This behavior is now almost identical except Speckle will skip levels that already exist
imports-needed-levels

There are now only 4 highlighted levels

## Type of change

## How has this been tested?

- Received several streams to check to make sure duplicate levels were not being created

## Docs

- No updates needed


teocomi (approved):

teocomi:
@bimgeek you can let the user know this has been fixed! We can make a prerelease if he needs

AlanRynne:
I'm guessing this is planned to be released in 2.7. If so, it should be merged into `release/2.7` instead of `main` ;)

PR #1402, opened 2022-07-14
feat(objects/revit): organization classes

izzylys:
## Description

Adds model organization classes to be used as commit root objects. These are of course optional and non-breaking but will give our connectors a consistent way of storing ancillary info alongside other objects. They have been implemented in Revit as a first pass. The new models include:

- `Model`: replaces the plain base root commit object and holds all objects within the commit
- `ModelInfo`: contains information about the model. can be subclassed for more industry-specific model types (eg `BIMModelInfo`)
- `ModelSettings`: additional settings that may be relevant to the commit

more context in [notion](https://www.notion.so/speckle/Objects-Cleanup-392cd1688c684a35a629ca8449bb6889)

sample commit:
https://latest.speckle.dev/streams/0c6ad366c4/commits/b7a8187bcb

## Type of change

## How has this been tested?

- Manual Tests 

## Docs

- Will be updated ASAP


AlanRynne (approved):
Awesome stuff!! 🚀

PR #1403, opened 2022-07-14
feat: adds stream and commit count to User query

teocomi:
## Description

- adds stream and commit count to User query


teocomi:
Going to 2.6 so I can release new nugets

PR #1422, opened 2022-07-20
CORE: Allows for `Item` to be set as a dynamic or typed prop in `Base`-derived objects

didimitrie:
+ other minor cleanups: caches reflection calls & DRYs the GetProperties() calls. 

Regarding the `Item` prop discussions, please see this discourse thread: https://speckle.community/t/why-do-i-keep-forgetting-base-objects-cant-use-item-as-a-dynamic-member/3246/4

## Type of change

- 
## How has this been tested?

- Unit Tests + Added new tests to cover both dynamic and typed `Item` props

## Docs

- No updates needed

### DIFF
--- a/Core/Core/Models/DynamicBase.cs
+++ b/Core/Core/Models/DynamicBase.cs
@@ -95,6 +95,7 @@ namespace Speckle.Core.Models
     /// </summary>
     /// <param name="key"></param>
     /// <returns></returns>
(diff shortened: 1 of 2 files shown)


AlanRynne (approved):
Gave it a quick go and it seems like it hasn't broken anything!

You'll have to walk me through the reasoning behind the cache you've implemented, but I'll accept this changes :D

PR #1431, opened 2022-07-21
Revit: Categories & Core: account and permissions

teocomi:
## Description

- adds method in AccountManager to update the default account
- renames StreamGrantPermission to StreamUpdatePermission to align with latest server changes

## Type of change

- Breaking change: the change to StreamUpdatePermission is breaking!

## How has this been tested?

- Manual Tests
## Docs

- No updates needed

### DIFF
--- a/Core/Core/Api/GraphQL/Client.GraphqlClientOperations.cs
+++ b/Core/Core/Api/GraphQL/Client.GraphqlClientOperations.cs
@@ -532,9 +532,9 @@ namespace Speckle.Core.Api
     /// </summary>
     /// <param name="permissionInput"></param>
     /// <returns></returns>
-    public Task<bool> StreamGrantPermission(StreamGrantPermissionInput permissionInput)
+    public Task<bool> StreamUpdatePermission(StreamUpdatePermissionInput permissionInput)
     {
-      return StreamGrantPermission(CancellationToken.None, permissionInput);
+      return StreamUpdatePermission(CancellationToken.None, permissionInput);
     }
 
     /// <summary>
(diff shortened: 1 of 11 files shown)


Core/IntegrationTests/Api.cs
```diff
@@ -108,10 +108,10 @@ public async Task StreamUpdate()
     }
 
     [Test, Order(30)]
-    public async Task StreamGrantPermission()
```
izzylys, line 111: tests here will still always fail because `streamUpdatePermission` is not a swap and replace with `streamGrantPermission` as you can't "grant" a role to a new user anymore - they need to be invited.

Core/Core/Api/GraphQL/Client.GraphqlClientOperations.cs
```diff
@@ -543,16 +543,16 @@ public Task<bool> StreamGrantPermission(StreamGrantPermissionInput permissionInp
     /// <param name="cancellationToken"></param>
     /// <param name="permissionInput"></param>
     /// <returns></returns>
-    public async Task<bool> StreamGrantPermission(CancellationToken cancellationToken, StreamGrantPermissionInput permissionInput)
```
izzylys, line 546: it's prob a good idea to keep both `StreamGrantPermission` and `StreamUpdatePermission` with a server version check in each as this is p breaking and people are unlikely to upgrade their servers right away

izzylys (changes requested):

AlanRynne:
Oopsies, deleted a branch and this got closed with it! Shouldn't have happened 😅

teocomi:
Closing this as we'll open a new one instead I believe

PR #1438, opened 2022-07-22
fix(ci): Added extension parameter to distinguish between zip and exe

AlanRynne:
## Description

- Fixes issue with new deployment mechanism looking for `.exe` files instead of `.zip`

Added a `extension` parameter to the job that will be set on the config generator.

## Type of change

## How has this been tested?

Ran the validation and generator loally.

## Docs

- No updates needed


AlanRynne:
I accept the consequences of not waiting for a reviewer. But since its a CI only change, it's safe enough.

PR #1440, opened 2022-07-22
fix(objects): Null reference error in GetTransformedGeometry

AlanRynne:
## Description

- Fixes minor issue discovered when testing, where an incompatible geometry or incorrectly transformed would lead to a null reference error while expanding the object in Grasshopper.

Surely this happened in Dynamo too.

## Type of change

## How has this been tested?

## Docs

- No updates needed


izzylys (approved):
👏

PR #1444, opened 2022-07-25
fix(autocad): better dynamic block names

clairekuang:
## Description
Fixes bug by appending all simply typed dynamic block attributes to block name
- Closes #1443 

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

- Manual Tests

## Docs

- No updates needed

PR #1445, opened 2022-07-25
fix(autocad): dynamic block name improved

clairekuang:
## Description

Now concatenates all dynamic block simple props into definition name

- Closes #1444 

## Type of change

- 
## How has this been tested?

- Tests not required

## Docs

- No updates needed


AlanRynne (approved):
Looks good!!

PR #1451, opened 2022-07-26
using units from speckle instead of etabs default

connorivy:
## Description

Receiving in ETABS always defers to the default units in the program. Which leads to most of my models coming from Revit (in feet) to be 1/12 of their actual size when received in ETABS (which defaults to inches for imperial units)

## Type of change

## How has this been tested?

Receiving and sending different models to make sure everything works as expected

## Docs

- No updates needed


Reynold-Chan (approved):
looks great !

PR #1456, opened 2022-07-28
FIX(ci): Nuget version picking nuget prefix instead of version.

AlanRynne:
## Description

- Fixes version picking bash command grabbing the prefix `nugets` instead of the actual version number. This caused the build to not create a nuget package, and the upload step would fail.

## Type of change

## How has this been tested?

## Docs

- No updates needed


AlanRynne:
Skipping review as this does not affect code, just the release pipeline for nugets.

PR #1460, opened 2022-07-31
CORE: Blob support

didimitrie:
NOTE: should be merged in sync with https://github.com/specklesystems/speckle-server/pull/845 ⚠️

## Description

Adds blob support for the ServerTransport in Core. This enables sending/receiving blobs (files!) in an integrated way. 

How does it work? Check out the [notion docs](https://www.notion.so/speckle/NET-Blob-Storage-Support-Scoping-10789fe5c14848dabf452892be2dab36). 

## Type of change

## How has this been tested?

There's a new set of Integration tests that flex the Server Transport's new Blob sending/receiving capabilities. 

## Docs

I would publish an updated set of documentation later, after we implement texture support and nail down any possible bugs and implementation woes. I wouldn't want people to start depending too early on unbaked SDK & API functionality.


teocomi:
Any updates on this @didimitrie ?

AlanRynne:
Hey @didimitrie, I just noticed that the counterpart PR for this issue on `speckle-server` has already been merged 🥳

I'm adding this to the `2.10` milestone so we don't forget to deal with it for the next release! 🙌🏼

didimitrie:
Perfect @AlanRynne, it's gonna be a fun challenge :)

didimitrie:
Post discussion with @gjedlicska, what has been decided: 

- SQLite Transport needs to become `~IBlobCapable`
- SQLite Transport will copy blobs locally on send in its BlobStorageFolder
- BlobStorageFolder will be provided by Core.Api.Helpers, and will point to a single location
- Server Transport will also use BlobStorageFolder as its dumping ground for blobs on receive
- We will no longer store blobs /server/streamId/, they will be just by their own hash in one folder 

What the problem was: when a user would send data including blobs to speckle, and then would want to receive that data from speckle on the same computer, core would assume those blobs already exist there (the local transport would have thought it has *everything*, and would basically not allow the remote transport to download anything). In the case the blobs would actually no longer be there, this would cause essentially an incomplete commit. 

The solution described above is the most robust way we can solve this case without refactoring massively the way transports and deserialisation works - which would be several months of work. Given are current usage patterns, this is not the worst situation.

PR #1464, opened 2022-08-01
rhino: changes user attributes from dict to base

### LINKED ISSUE #1333: rhino: change user strings and user dictionaries to `Base`
Currently, user strings are being attached as `Dictionary<string, string>` and user dictionaries are being attached as `Dictionary<string,object>`. 

To make this more interoperable with GH, (and best practices), should they be attached as `Base` with dynamic props instead?
Any use cases where `Base` would be problematic?

clairekuang:
## Description
Previously, user strings and user dictionaries were being sent as dictionaries in Rhino. They are now attached as `Base`.
Also checks for `name` prop on receive.

Attaching user info has been moved from the connector to the converter

- Fixes #1333

## Type of change

## How has this been tested?

- Manual Tests: user strings, names, and dicts in rhino

## Docs

- No updates needed


AlanRynne (approved):
Checked it in Grasshopper and it does the trick! ✅

Issue #1469, opened 2022-08-02
Dockable panel initialization not working when launching Revit form a file


teocomi:
If launching Revit by double-clicking on a Revit file (instead of first launching Revit and then opening a file) the initialization of the Dockable Panel is not working and throws `The requested dockable pane has not been created yet.`

This seems to be an issue with the Revit API as the panel under both circumstances is actually registered and initialized correctly.

As a matter of fact, as a workaround, **opening another Revit file** is correctly starting the dockable panel.

**Steps to repro**

1. Add command line args to launch revit file:

![image](https://user-images.githubusercontent.com/2679513/182321486-680b87e1-cf14-4b61-b586-a3e0d1f50ce9.png)

2. Click debug
3. After the project has been opened, click on the Speckle command

**Solution tried**

I tried to simplify, and force a re-registration of the panel when clicking on the ribbon button, to no avail.
I also tried changing when the panel is first initialized from the `Idle` to the a `ControlledApplication.ApplicationInitialized` event.

Some more details:

```
        var p1 = DockablePane.PaneIsRegistered(PanelId);  //true
        var p2 = DockablePane.PaneExists(PanelId); //false
        var p3 = App.AppInstance.GetDockablePane(PanelId); //throws "The requested dockable pane has not been created"
```

Also, no events like `DocumentOpened`, or `ViewActivated` are triggered under these circumstances.

![image](https://user-images.githubusercontent.com/2679513/182321026-00f14c5d-069a-421f-bc53-48a96fd10d36.png)

teocomi:
Repro with minimal reproducible project: https://github.com/teocomi/DockableDialogBug

teocomi:
This has been recognized as an issue by ADSK and has been logged internally by them as:

REVIT-195285 [Dockable Panel not registered correctly when Revit is opened from a file -- 19345692]

teocomi:
Will reopen when we have a resolution from ADSK.

GianClaudioScarafini:
Hi @teocomi ,

sorry to bother you.

I still have this issue when I am using a specific workflow, not speckle system related. I am extracting Revit Database info using a "ModelChecker\bestpractices-2023.xml" pointing to the central file where I want to extract information. 

the workflow works in this way. when I start the Revit session it automatically generate and excel file opning in the backgorung the central file.

so when I initialize the software I have this issue that keeps going to pup up even if I click on close.
[...]

PR #1470, opened 2022-08-02
fix(revit): workaround for #1469

teocomi:
## Description

- Workaround for #1469

This is a workaround just to let the user know Revit cannot correctly register Dockable panels when launched from a file, instead of showing an error message.

Other possible solutions:
- try to create/open a doc in the background (i tried and failed)
- launch the old windowed UI (possible conflicts)

We should merge this in 2.7 and release a hotfix.

## Type of change

## How has this been tested?

## Docs

- No updates needed

### DIFF
--- a/ConnectorRevit/ConnectorRevit/Entry/SpeckleRevitCommand2.cs
+++ b/ConnectorRevit/ConnectorRevit/Entry/SpeckleRevitCommand2.cs
@@ -105,7 +105,7 @@ namespace Speckle.ConnectorRevit.Entry
         TaskDialog mainDialog = new TaskDialog("Dockable Panel Issue");
         mainDialog.MainInstruction = "Dockable Panel Issue";
         mainDialog.MainContent =
-            "Revit cannot properly register Dockable Panels when launched double-clicking a Revit file. "
+            "Revit cannot properly register Dockable Panels when launched by double-clicking a Revit file. "
             + "Please close and re-open Revit without launching a file OR open/create a new project to trigger the Speckle panel registration.";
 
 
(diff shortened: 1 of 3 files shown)


ConnectorRevit/ConnectorRevit/Entry/SpeckleRevitCommand2.cs
```diff
@@ -67,6 +71,55 @@ public Result Execute(ExternalCommandData commandData, ref string message, Eleme
...
+      created = DockablePane.PaneExists(PanelId);
+
+      //if revit was launched double-clicking on a Revit file, we're screwed
+      //could maybe show the old window?
+      if (!created && App.AppInstance.Application.Documents.Size > 0)
+      {
+        TaskDialog mainDialog = new TaskDialog("Dockable Panel Issue");
+        mainDialog.MainInstruction = "Dockable Panel Issue";
+        mainDialog.MainContent =
+            "Revit cannot properly register Dockable Panels when launched double-clicking a Revit file. "
```
connorivy, line 108: This is a little nit-picky, but it seems like there should either be 'by' or 'via' between 'launched' and 'double-clicking'

connorivy (approved):

ConnectorRevit/ConnectorRevit/Entry/SpeckleRevitCommand2.cs
AlanRynne: I agree 😅

AlanRynne (approved):
Seems to work as advertised ;)

Issue #1248, opened 2022-05-10
Type mapping across connectors


teocomi:
As different BIM (and CAD) software use and handle types in completely different ways we need to explore a solution that would allow users to set custom "mappings" between the incoming data in a stream and the available families/types in the receiving software.
This solution should not be Revit centric as many other host apps will have the same issues, but Revit could definitely be a starting point.
See: https://speckle.community/t/speckle-library-manager/2853/4
Related also to #1240

teocomi:
Closing as implemented in https://github.com/specklesystems/speckle-sharp/pull/1489 for now

PR #1489, opened 2022-08-08
Feature: BIM Mapping

connorivy:
## Description

- Development toward a solution described in -> https://github.com/specklesystems/speckle-sharp/issues/1248
- This code specifically implements a solution to map types from ETABS/Revit to Revit. There is also code to support tekla, but it looks like we need to change the way we receive tekla objects in order for that code to work. 

## Type of change

## How has this been tested?

I've received several commits using this mapping. However, I have a feeling that there are many corner cases that will not play nice with this feature

## Docs

- Will be updated ASAP
Will update docs before this feature is released


DesktopUI2/DesktopUI2/Views/Windows/Dialogs/MappingViewDialog.xaml
```diff
@@ -0,0 +1,209 @@
...
+      </ScrollViewer>
+      <StackPanel
+        Orientation="Horizontal"
+        HorizontalAlignment="Right">
+        <Button
+          MaxWidth="200"
+          CornerRadius="20"
+          Margin="5"
+          HorizontalAlignment="Left"
+          Command="{Binding ImportFamily}">
```
teocomi, line 194: Very minor remark: I would prefer commands to be named like `ImportFamilyCommand` so it's easier to understand they have a binding in the c# code and avoid the risk of deleting/renaming them involuntarily ;)

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit2/ConnectorBindingsRevit2.Mappings.cs
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    /// <summary>
+    /// This method creates a mapping dictionary by looping through all of the incoming types and mapping them to the most similar host type of the same category
+    /// </summary>
+    /// <param name="incomingTypesDict"></param>
+    /// <param name="hostTypes"></param>
+    /// <param name="progress"></param>
+    /// <returns></returns>
+    public Dictionary<string, List<MappingValue>> returnFirstPassMap(Dictionary<string, List<string>> incomingTypesDict, Dictionary<string, List<string>> hostTypes, ProgressViewModel progress)
```
teocomi, line 89: Very minor remark #2: function names should be camel case, and they should be private or internal instead of public, unless necessary ofc

teocomi (approved):
Looks good! I didn't try to understand the mapping logic, so I fully trust you there ;)
Just had 2 minor stylistic comments.

The UI also looks great, I'd just add some margin down here:

image

PR #1494, opened 2022-08-09
Update `main` with changes from `release/2.7`

AlanRynne:
Just your normal updating PR... nothing to see here :)


AlanRynne:
Squash merging this commit and deleting the 2.7 delete branch. All hotfixes should now start directly in main, and all stable releases should be made in main.

PR #1497, opened 2022-08-10
Mixpanel connectors

teocomi:
Minor improvements for our analytics class


AlanRynne (approved):
I approve! Haven't fully tested it but we should all do so once this is merged in our own connectors, so I'd argue it should go in 👍🏻

PR #1500, opened 2022-08-11
GH: SOM reported bug fixes

### LINKED ISSUE #1499: GH: Bugs reported by SOM
Kindly reported by conference call by the lovely SOM team ♥️

## Describe the bug

- [x] Account details returns a null reference exception
   Screenshot 2022-08-11 at 07 57 40
- [ ] #1501
   Screenshot 2022-08-11 at 07 58 14
- [x] Inherit names shortcuts (shift+tab, tab+connect, shift+double-click) don't seem to work correctly, or sometimes not at all.
Screenshot 2022-08-11 at 07 59 35
- [x] Extract parameter function uses wrong offset, landing the new params on top of the existing component.
   Screenshot 2022-08-11 at 07 59 20

AlanRynne:
Closes #1499

- Account details issue was related to the `StreamsGet` function in Core not expecting the user to be `null`, which can be if there are no provided credentials, the credentials are invalid, or the user does not exist anymore. I added a better exception for that case to provide more clarity to the user.
Screenshot 2022-08-11 at 11 22 17

- `T0: One or more errors ocurred` while sending. This seems to be related to the first item, but I haven't managed to repro so this one remains **unfixed**. Could have been caused by a faulty connection. We checked the server logs but nothing showed up. Tried invalidating my token to verify if this would replicate the error, but I got the expected warnings 👇🏼
Screenshot 2022-08-11 at 11 23 18

- Renaming input. behaviour on the Send component should now work as expected.

- **Slow renaming** was caused (at least in the provided example file) by the **grasshopper auto save feature**, deactivating it made everything work smoothly again. This is usually caused by having too many internalised objects that need to be serialised before saving the GH document.

- Extract parameter nodes not being placed properly: It should now work exactly as the original "ExtractParameter", same offset and position, but since we're talking about outputs, it would go to the right of the node.

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/Extras/GenericAccessParam.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/Extras/GenericAccessParam.cs
@@ -39,7 +39,7 @@ namespace ConnectorGrasshopper.Extras
         switch (Owner.Kind)
         {
           case GH_ParamKind.input:
-            (Owner as GenericAccessParam)?.InheritNickname();
+            (Owner as SpeckleStatefulParam)?.InheritNickname();
             return GH_ObjectResponse.Handled;
           case GH_ParamKind.output:
             Clipboard.SetText(DocObject.NickName);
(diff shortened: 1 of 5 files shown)


izzylys (approved):
by the power vested in me as an old and grouchy speckler, i hereby approve this pr #1500. a beautiful and satisfyingly round yet arbitrary number is blessed on this most glorious thursday

*shoots confetti cannon* 🥳🎉🎊✨

PR #1505, opened 2022-08-11
fix(revit): updated GetElementType method to return bool with logging

### LINKED ISSUE #1498: DUI report issues with missing/replaced types
When receiving a commit containing missing types in Revit, a user should be informed that these are missing and have been automatically replaced with a different type from the same category.
This was happening in the old report and seems to have slipped out in the latest version?

To replicate, try receiving this stream with 3 different types of beams in a new Revit project that uses the `Architectural` template: https://latest.speckle.dev/streams/dc18cd5fa5/commits/ecf7120e71

No warning is shown, but all types have been replaced:
image

When instead receiving missing families that do not have substitutes, the report seems to indicate something, but there is no message or tooltip to understand what happened, maybe this is still WIP?

It can be replicated by receiving this commit in a new `structural` project: https://latest.speckle.dev/streams/dc18cd5fa5/commits/2c4be60e8c

image

Also the legibility of the items could be improved :)

clairekuang:
- the `GetElementType()` methods now returns a bool to allow for easier logic checks in primary conversions
- logging has been updated so that `ApplicationObjects` are always being returned on instances of failing `GetElementType` methods instead of a speckle exception being thrown
- fixes #1498

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -482,7 +482,7 @@ namespace Objects.Converter.Revit
 
     #region  element types
 
-    private T GetElementType<T>(string family, string type, ApplicationObject appObj)
+    private bool GetElementType<T>(string family, string type, ApplicationObject appObj, out T value)
     {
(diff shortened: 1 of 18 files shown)


teocomi (approved):
Works!

PR #1511, opened 2022-08-12
CORE: Fixes `SpecialFolder.ApplicationData` pointing to `Default` user

AlanRynne:
Reported in the forum:

We were assuming that `Environment.SpecialFolder.ApplicationData` would always return the correct user path for the **current roaming user**; but this is not the case. For some network users, the path returned is the one for the `Default` user, which is used as a template for new users in windows.

This leads to our KitManager, AccountManager and others to not be able to find the Speckle folder, as they're looking in the wrong place! 😱

**Solution**

After a bit of back and forth we decided to assemble our own **AppData** path by getting the username of the current user and attaching `/AppData/Roaming` at the end.

In order to make this change universal, there are some new helper methods `Helpers.SpeckleFolderPath` being the most important one; this will return the correct speckle folder path, skipping this problem, but also taking into account the install location for users who installed systemwide.

All places where this was in use have been properly updated to use this new helper.

For reference: https://speckle.community/t/cant-find-speckle-kits-when-using-gh-sdk/3297/21


Core/Core/Transports/SQLite.cs
```diff
@@ -48,7 +49,7 @@ public SQLiteTransport(string basePath = null, string applicationName = "Speckle
       Log.AddBreadcrumb("New SqlLite Transport");
 
       if (basePath == null)
-        basePath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+        basePath = Helpers.CorrectedApplicationDataPath;
```
teocomi, line 52: See comment for accounts, this should not be shared across users.

Core/Transports/DiskTransport/DiskTransport.cs
```diff
@@ -30,7 +31,7 @@ public class DiskTransport : ICloneable, ITransport
     public DiskTransport(string basePath = null)
     {
       if (basePath == null)
-        basePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Speckle", "DiskTransportFiles");
```
teocomi, line 33: See comment for accounts, this should not be shared across users.

teocomi (changes requested):
Added inline comments, happy to chat about them.

AlanRynne:
I've refactored things a bit. We now have inside `Core.Api.Helpers`:

- `UserApplicationDataPath` returns the %appdata% path for the current user
- `InstallApplicationDataPath` returns the application data folder where speckle was installed (either _appdata_ or _programdata_)
- `InstallSpeckleFolderPath` returns the path to the Speckle installation folder
- `UserSpeckleFolderPath` returns the path to the **users Speckle folder**, were their respective accounts and cache should live.

teocomi (approved):

PR #1513, opened 2022-08-12
fix unit type change from <=2021 to >=2022

connorivy:
Due to changes in the Revit API, if a commit is sent or received in version <=2021, then the parameters are set with a DisplayUnitType value. If a commit is sent or received in version  >=2022, then the parameters are set with a ForgeTypeId value.

This makes it so that if you send a commit in Revit 2021 and then receive it in Revit 2022, then some parameters will not set properly. Here are two example streams

https://speckle.xyz/streams/445f21d47d/commits/01a748dc60
https://speckle.xyz/streams/5a087f4c9a/commits/78267ed4ce

Here are some screenshot of receiving these two commits (sent from Revit 2021) in 2022 before this PR:
jesuit not working
fincher not working

Here are some screenshot of receiving these two commits after this PR (these are correct):
jesuit working
fincher working


teocomi (approved):
Great! I think I was aware of this but never got around doing anything about it ;)

PR #1517, opened 2022-08-15
Fix to direct mesh receiving in Tekla

Reynold-Chan:
- direct meshes weren't working for other streams other than Tekla Structures receive in Tekla Structures .
- Now working for other streams sent from other programs. 
- Tested manually


AlanRynne (approved):
I'm trusting you on this one, as I have no way of testing it in my machine.

Changes seem harmless to other connectors ✅

Only thing I'd say is that `ConvertBeam.ColumnToNative` has a bunch of commented out code that could be removed entirely?

PR #1521, opened 2022-08-16
Tekla Structures Installer TSEP File

Reynold-Chan:
TBD


.circleci/scripts/config-template.yml
```diff
@@ -143,6 +143,51 @@ jobs: # Each project will have individual jobs for each specific task it has to
                 paths:
                   - speckle-sharp-ci-tools/Installers
 
+  build-connector-new: # Reusable job for basic connectors
```
AlanRynne, line 146: Now that we have this working, let's try to come up with a more elegant solution than duplicating the entire connector build step.

.circleci/scripts/config-template.yml
```diff
@@ -143,6 +143,51 @@ jobs: # Each project will have individual jobs for each specific task it has to
...
+            $semver = if($tag.Contains('/')) {$tag.Split("/")[0] } else { $tag }
+            $ver = if($semver.Contains('-')) {$semver.Split("-")[0] } else { $semver }
+            $channel = if($semver.Contains('-')) {$semver.Split("-")[1] } else { "latest" }
+            $version = "$($ver).$($env:WORKFLOW_NUM)"
+            msbuild << parameters.slnname >>/<< parameters.slnname >>.sln /p:Configuration=Release /p:WarningLevel=0 /p:IsDesktopBuild=false /p:AssemblyVersionNumber=$version /p:AssemblyInformationalVersion=$semver /p:Version=$semver  
+            if(-Not [string]::IsNullOrEmpty($env:CIRCLE_TAG)) { New-Item -Force "speckle-sharp-ci-tools/Installers/<< parameters.slug >>/$channel.yml" -ItemType File -Value "version: $semver" }
+          environment:
+            WORKFLOW_NUM: << pipeline.number >>
+
+      - run: # Creates the inno installer
```
AlanRynne, line 181: This is basically the step we need to deal with, as everything else is unique :)

Maybe we just copy it over to the build-connector job, and just do some conditional steps to just run it on it's slug.

AlanRynne (changes requested):
TSEP file should never be committed! Oopsies! 🫢

Lets delete that and add it to gitignore

Reynold-Chan:
> TSEP file should never be committed! Oopsies! 🫢
> 
> Lets delete that and add it to gitignore

haha yep

AlanRynne (approved):
YAY! 🥳 It all looks good now

PR #1523, opened 2022-08-17
Added static accessor to `DynamicBase.GetInstanceMembers`

JR-Morgan:
## Description & motivation
While creating some editor UI in Unity to inspect properties.

My current needs are to cache PropertyInfo of types inheriting `Base`, since using reflection every UI update is unacceptable.
We already cache PropertyInfo of speckle types inside `DynamicBase`.
But this data is only accessible through `GetInstanceMembers`

The problem is, I don't have an instance to call `GetInstanceMembers` on, just the `Type`
And I don't want to have to create a dummy instance of a `Type` (which I would then have to cache in the UI), just to call `GetInstanceMembers` (which just calls `GetType()` to access its internal static `propertyInfo` dictionary)

So I propose this small change. (first two lines are the important change)
https://github.com/specklesystems/speckle-sharp/blob/711238ecf7311a12e782302244f6bb06e4d20ad9/Core/Core/Models/DynamicBase.cs#L188-L200

Secondly, I also noticed that setting properties of an instance frequently, was leading to a lot of memory allocations (slow for GC, especially when done frequently). And I thought, while I was there, I'd try and tackle any low-hanging fruit, optimisation wise.

## Changes:
1. Added static accessor for `GetInstanceMembers` and `GetInstanceMemberNames`.

2. Small performance optimisation (mostly surrounding memory allocation) of DynamicBase methods.
All changes are non breaking.

Here are the performance results if anyone is interested!
New methods show a slight optimisation to basically all methods, in both execution time, and memory allocation. Most are very slight, and likely won't yield any tangible benefits.
But this was low hanging fruit, and (imo) only improves code readability.

image

The biggest benefit can be seen in the `this[string]` operator (which calls `IsPropNameValid`) we can see memory allocation is roughly 10% of current method.
Here I'm testing setting a prop 1000 times. 4MB of allocated memory is not insignificant.

```
|                    Method |       Mean |    Error |   StdDev |    Gen 0 | Allocated |
|-------------------------- |-----------:|---------:|---------:|---------:|----------:|
|     SetAProperty1000Times | 3,422.0 us | 18.65 us | 16.53 us | 570.3125 |  4,398 KB |
| SetAProperty1000Times_new |   347.9 us |  0.78 us |  0.65 us |  62.0117 |    477 KB |

```

## To-do before merge:
 - [ ] Maybe test a connector?  Core tests pass, and I'm reasonably confident of their quality. 

## Validation of changes:

I noticed a couple cases missed in the  `BaseTests`. Mostly surrounding `Obsolete` and `SchemaIgnore` tests.

## References
https://stackoverflow.com/questions/14719104/is-there-a-benefit-of-using-isdefined-over-getcustomattributes

### DIFF
--- a/Core/Core/Models/DynamicBase.cs
+++ b/Core/Core/Models/DynamicBase.cs
@@ -54,15 +54,14 @@ namespace Speckle.Core.Models
       return valid;
     }
 
+    // Rule for multiple leading @.
+    private static Regex manyLeadingAtChars = new Regex(@"^@{2,}");
+    // Rule for invalid chars.
+    private static Regex invalidChars = new Regex(@"[\.\/]");
     public bool IsPropNameValid(string name, out string reason)
     {
-      // Regex rules
-      // Rule for multiple leading @.
-      var manyLeadingAtChars = new Regex(@"^@{2,}");
-      // Rule for invalid chars.
-      var invalidChars = new Regex(@"[\.\/]");
       // Existing members
-      var members = GetInstanceMembersNames();
+      //var members = GetInstanceMembersNames();
 
       // TODO: Check for detached/non-detached duplicate names? i.e: '@something' vs 'something'
       // TODO: Instance members will not be overwritten, this may cause issues.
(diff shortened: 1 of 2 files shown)


AlanRynne (approved):
Tested in Grasshopper. Found nothing broken 🚀

didimitrie (approved):
I'll need a a mini master class on the performance side of things in here :)

AlanRynne:
_In Jedd's numbers we trust_

PR #1524, opened 2022-08-17
Update existing elements enabled for CSI connectors

Reynold-Chan:
Tested manually

Need to update docs

A bit slow since it refreshes the window with every update. But this is the best solution so far I can think of that works.


AlanRynne (approved):
Just had a quick look and everything seems fine!

No potential conflicts with other connectors makes me happy🥳

PR #1526, opened 2022-08-17
Added ApolloGraphQL headers to GraphQL `Client`

### LINKED ISSUE #1429: Add client identification headers to GraphQL API requests
It would be great if all GraphQL requests to the Speckle Server GraphQL API would have headers attached that correctly identify the client making the request, and its version.

These are the headers that should be used:
```
apollographql-client-name: {{UNIQUE IDENTIFIER OF YOUR CLIENT}}
apollographql-client-version: {{VERSION}}
```

JR-Morgan:
## Description & motivation

Closes #1429 

## Changes:

Added `apollographql-client-name` and `apollographql-client-version` headers to `Client`

`Speckle.Core.Logging.Setup` will use the current process name as a fallback `HostApplication` string, rather than simply falling back to "other"


teocomi (changes requested):
Looks good! Although we might wanna merge in 2.8, and I'm afraid there will be a couple of minor conflicts (we usually work from the current release branch, which is 2.8 in this case)

teocomi (approved):

PR #1529, opened 2022-08-18
CORE: Adds extension method to extract all messages from an Exception

AlanRynne:
We usually get errors that are wrapped in other errors, where the meaningful message is somewhere deep inside the nested `Exception` structure.

I've added a new Exception extension in `Speckle.Core.Models.Extensions` (may not be the best place... TBD) called `ToFormattedString()`, which will fetch all messages from an Exception and print them.

So, instead of getting a useless `One or more errors occurred` message, you will now also get all the inner errors that actually occurred, right after the initial message (one line per exception).

This does not support indenting them, as I thought it would make it weird for our reports, but it could be a nice optional thing to have.

### Why

Because `exception.ToString()` will print all the exception messages along with their call stack. Even though having the call stack is nice in some occasions, the end user doesn't really care, and we should have that information in Sentry anyway.

`ToFormattedString` is intended to be used only when an exception message is going to be displayed to the user.

The method may require some improvements, unit tests will be added as soon as we agree this is a good idea

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/Ops/Operations.VariableInputSendComponent.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/Ops/Operations.VariableInputSendComponent.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Drawing;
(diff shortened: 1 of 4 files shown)


clairekuang:
Updated DUI2 report and tested in autocad, looks ok to me!

clairekuang (approved):

PR #1537, opened 2022-08-23
fix(ci): core test flag

izzylys:
ci flag was wrong which meant core tests were never run

making a PR mainly to test/trigger the ci since core tests were passing on my server invites pr before i merged in 2.8 and am now having issues


AlanRynne (approved):
This didn't do the trick to fix the tests hanging, but it's a necessary fix either way

AlanRynne:
Fix for failing test is coming in #1550, so I'm merging this in

PR #1538, opened 2022-08-23
fix(core): AppData path fix for windows only

AlanRynne:
JUST TESTING; DO NOT MERGE THIS.

Was created to force CI to run


AlanRynne:
Fixed in #1550

PR #1540, opened 2022-08-23
feat(dui2): adds report filters by object status

clairekuang:
## Description & motivation
Adds buttons to the report to filter report objects by status (created, updated, skipped, failed, etc).
Works together with the search bar to filter report objects.

## Changes:
DUI2 Report


clairekuang:
> Looks good, although I haven't manually tested it.
> 
> Could you maybe add a fake report generation to the DummyBindings so it's easier to test locally?

Added!

PR #1548, opened 2022-08-26
fix(rh): Bumped mac ref to dylib in Rhino csproj

AlanRynne:
Fixes failing mac build introduced in #1542

### DIFF
--- a/ConnectorRhino/ConnectorRhino7/ConnectorRhino7.csproj
+++ b/ConnectorRhino/ConnectorRhino7/ConnectorRhino7.csproj
@@ -44,7 +44,7 @@
     <ProjectReference Include="..\..\DesktopUI2\DesktopUI2\DesktopUI2.csproj" />
   </ItemGroup>
   <ItemGroup Condition="'$([MSBuild]::IsOSPlatform(OSX))' == 'true'">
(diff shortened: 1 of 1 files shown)


ConnectorRhino/ConnectorRhino7/ConnectorRhino7.csproj
```diff
@@ -44,7 +44,7 @@
     <ProjectReference Include="..\..\DesktopUI2\DesktopUI2\DesktopUI2.csproj" />
   </ItemGroup>
   <ItemGroup Condition="'$([MSBuild]::IsOSPlatform(OSX))' == 'true'">
-    <None Include="$(HOME)/.nuget/packages/avalonia.native/0.10.17/runtimes/osx/native/libAvaloniaNative.dylib" CopyToOutputDirectory="PreserveNewest" />
+    <None Include="$(HOME)/.nuget/packages/avalonia.native/0.10.18/runtimes/osx/native/libAvaloniaNative.dylib" CopyToOutputDirectory="PreserveNewest" />
```
teocomi, line 47: would a wildcard work here? So we do not forget updating this in the future...

teocomi (approved):

ConnectorRhino/ConnectorRhino7/ConnectorRhino7.csproj
AlanRynne: Not sure it would... I think we should actually be using this:

https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#generatepathproperty

PR #1557, opened 2022-08-29
feat(gh): Added ability for the receiver to fetch public streams with an old account

AlanRynne:
The way the `Client` is implemented does not allow for the fetching of data without an Account for that server being present in the system. This makes sense to some degree, but not necessarily so when you're dealing with _public streams_ in _public servers_

I added a modification in the `Receiver` that will pass in a "fake account" with the basic info set so that the client will not throw an exception.

**Bonus**: removed local redundant client creation in favor of `ApiClient`

**Possible better workaround**

This solution could be pushed up to `Core`, so that the `StreamWrapper` would provide the correct "fake account" for any url, including ones that you don't have setup yet.

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopper/Ops/Operations.VariableInputReceiveComponent.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopper/Ops/Operations.VariableInputReceiveComponent.cs
@@ -24,6 +24,7 @@ using Speckle.Core.Api.SubscriptionModels;
 using Speckle.Core.Credentials;
 using Speckle.Core.Logging;
 using Speckle.Core.Models;
+using Speckle.Core.Models.Extensions;
 using Speckle.Core.Transports;
 using Utilities = ConnectorGrasshopper.Extras.Utilities;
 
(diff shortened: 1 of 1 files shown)


ConnectorGrasshopper/ConnectorGrasshopper/Ops/Operations.VariableInputReceiveComponent.cs
```diff
@@ -424,8 +425,20 @@ private async Task ResetApiClient(StreamWrapper wrapper)
...
-        ApiClient = new Client(acc);
+        Account account = null;
+        try
+        {
+          account = wrapper?.GetAccount().Result;
+        }
+        catch (Exception e)
+        {
+          AddRuntimeMessage(GH_RuntimeMessageLevel.Warning, e.ToFormattedString());
+          account = new Account
```
AlanRynne, line 436: Here's the "fake account". It basically just contains the  **server url** and the **stream id**.

teocomi (approved):
Seems innocent enough as the client and account are not shared outside the node...!

PR #1562, opened 2022-08-30
fix(gh): Moved RevitMaterial and RevitWallOpening to Revit tab

AlanRynne:
RevitMaterial and RevitWallOpening were incorrectly placed inside the `BIM` tab.

They should now be placed in the Revit tab as expected.

image


AlanRynne:
Ignoring CodeCov fail. I'm fixing this on a separate PR

PR #1564, opened 2022-08-30
fix(gh): Mac installer missing GH converter due sln config misconfiguration

AlanRynne:
Mac Installer for 2.8 is missing the Grasshopper converter, which causes it to fail when loading the default kit.

This was caused by the `Release Mac` config running some of the projects on `Debug` config, hence none of the converters would be found by our installer zip logic.

Additionally, added csproj flag to prevent VSCode from naming the converters as `gha`


AlanRynne:
Ignoring CodeCov fail. I'm fixing this on a separate PR

PR #1570, opened 2022-08-31
Objects: null point equality operator fix

### LINKED ISSUE #1516: Point == null operator overload not working as expected
Hi I ran into a tiny issue and thought I should let you know.

In my app I was making sure the basepoint of a speckle revitfamilyinstance was not Null, because for some hosted objects it is.
This stopped working when I upgraded to speckle Objects version 2.7.0.

   var pointA = (Point)null;
   Debug.Assert(pointA == null, "Point is not null."); // fails

It probably has something to do with the fact that the '==' operator is overridden.
If have changed my code to "if(pointA is null)" and that is working.
Haven't tested the other operator overloads in combination with null.

I think this might be confusing for future developers who don't know the == is overridden.

JR-Morgan:
## Description & motivation
Given a quick spin at this issue (hope you don't mind @AlanRynne)

closes #1516

## Changes:

`Point.operator==` now correctly returns true if both points are null.

## To-do before merge:
- [x] Test Revit connector (family instance to native)
I fear we may have been incorrectly relying on `basePoint == null` never being true.
https://github.com/specklesystems/speckle-sharp/blob/fa655fd5e883d5db26e7f3242352b8daa6c63965/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial%20Classes/ConvertFamilyInstance.cs#L277-L280

- [x] Check if we are overloading == incorrectly in other objects (I haven't checked)

### DIFF
--- a/Objects/Objects/Geometry/Point.cs
+++ b/Objects/Objects/Geometry/Point.cs
@@ -100,13 +100,17 @@ namespace Objects.Geometry
       point.y / val,
       point.z / val, point.units);
 
-    public static bool operator ==(Point point1, Point point2) =>
-      !(point1 is null) &&
-      !(point2 is null) &&
-      point1.units == point2.units &&
-      point1.x == point2.x &&
-      point1.y == point2.y &&
-      point1.z == point2.z;
+    public static bool operator ==(Point point1, Point point2)
+    {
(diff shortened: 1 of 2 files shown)


AlanRynne (approved):
🥳 Sorry for the delay on reviewing this. Looks great!

PR #1571, opened 2022-09-01
fix(dui2): disables send/receive preview button if binding returns null

clairekuang:
## Description & motivation

Disables the DUI2 `PreviewSend` and `PreviewReceive` buttons if the connector binding method returns null. This includes updating all connectors so that `PreviewSend` returned `Task<string>` instead of `void`.

Implementation notes:
- The original issue came from crahses on repeated click with the original preview button enabled, but bindings returning null or void. Better to disable for better metrics as well.
- This is achieved with a `Can` method check on the DUI2 stream view control that is bound to `Preview`, and is accomplished by triggering the `PreviewSend` or `PreviewReceive` bindings with `null` arguments.
- For any connector with implemented previews, this null check needs to be in place to prevent crashes!

## Changes:

DUI2 StreamEditViewModel `CanPreview` method
All DUI2 connector `PreviewSend` methods
Rhino and Revit `PreviewSend` and Rhino `PreviewReceive` methods to handle null values from `CanPreview` check

### DIFF
--- a/ConnectorArchicad/ConnectorArchicad/ConnectorBinding.cs
+++ b/ConnectorArchicad/ConnectorArchicad/ConnectorBinding.cs
@@ -98,7 +98,7 @@ namespace Archicad.Launcher
       return new List<ReceiveMode> { ReceiveMode.Create };
     }
 
-    public override Task<StreamState> PreviewReceive(StreamState state, ProgressViewModel progress)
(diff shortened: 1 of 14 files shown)


AlanRynne (approved):
This looks good! Approving this in case you need to merge it in while I'm offline!

clairekuang:
Closing this without merging, found a more stable solution for DUI2

PR #1577, opened 2022-09-01
fix(Rhino): send everything bug

clairekuang:
## Description & motivation

Bug reported by @AlanRynne : rhino connector hangs when using the send everything filter.

While testing, also found that named views selection in the report causes a crash, which is also resolved in this fix

Caused by null exception when sending named views

## Changes:
Rhino connector

## Validation of changes:

Sent some views


AlanRynne (approved):
Thanks @clairekuang! That was quick!

PR #1581, opened 2022-09-02
fix(gh): Get account fail when no OriginalInput was set in the StreamWrapper

AlanRynne:
Fixes an issue where `commitUrl` would not reset the ApiClient, leading to null exceptions or errors because the client was not refreshed.

Also uncovered a smaller issue with the `commitUrl` from the output of the `Send node` would be missing the `OriginalInput` string, leading to a null error if connected to a receive node.


izzylys (approved):
🐛🔨

Issue #1277, opened 2022-05-26
MacOS: Top-menu dropdown's stop working when Speckle loads


AlanRynne:
# Description

This issue is related to the hijacking of the App menu name issue, it may even be the same root cause.

Our Speckle DUI window is somehow hijacking the menu-bar of Rhino somewhere between calling `Start()` and when `CreateOrFocusSpeckle()` is called.

Someone at Avalonia also suggested in the past that we use `SetupWithoutStarting()` instead of `Start()`, but that doesn't seem to have any effect.

This leads to the top-menu's not displaying the dropdown when clicked.

This doesn't seem to affect Rhino at all, all commands and buttons in the main rhino window work.

# How to reproduce

In order to reproduce the issue:

1. Install latest Speckle Rhino for Mac
2. Start Rhino
3. You'll see the Rhinoceros menu is now actually called "Speckle". Click it to open it.
4. Now start Speckle by writing "Speckle" in the Rhino command line.

Once the window opens:
- The app menu is back to being named Rhinoceros
- None of the other top-menu's work.

![2022-05-26 09 15 22](https://user-images.githubusercontent.com/2316535/170439711-6a23c8ac-4323-439d-a3b4-36e0253ffba2.gif)

## How to get the menus to work

To get the menus to work, you just need to open any of the top menus before starting Speckle:

![2022-05-26 09 17 54](https://user-images.githubusercontent.com/2316535/170439793-cf43d8d4-4de7-400c-8e45-7b5f72e43f4e.gif)

jsdbroughton:
This isn't a zero affect other than UI. The speckle for rhino plug-in leaves a process running that causes a crash on quit. Grasshopper is not affected.

AlanRynne:
Latest version seems to have fixed the "locked up menus" problem.

The crash on exit persists, and Avalonia people are looking into it.

PR #1584, opened 2022-09-05
Patch(Rhino): Temporarily disable Rhino on Mac and delay `Init` call

AlanRynne:
Disable Rhino Connector on Mac only to prevent issue #1277.

> THIS IS NOT A FIX, JUST A PATCH.

TODO:
- [x] Mac connector does respond to the Disabled flag
- [x] Mac connector does not load Avalonia **ever** (for now!! 😅)
- [x] Win connector is not broken by these changes

### Reason

#1277 

Avalonia is not playing nice with Rhino at the moment, causing it to hijack the top-menu and making any items un-clickable.

This only happens as soon as we call the `Init` method inside our `Plugin` class, so I moved the initialisation to whenever we call `CreateOrShowSpeckle()`.

This allows me to directly show a warning on Mac before any Avalonia logic get's executed.

### "Solution"

For now, the warning directs the user's to use Speckle Grasshopper, as it is installed alongside Rhino and does not suffer from these problems.

image

If necessary, the plugin can still be enabled manually from the settings page (this will cause the menu issue to occurr though... but the connector works fine otherwise.

image

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/ConnectorRhinoCommand.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/ConnectorRhinoCommand.cs
@@ -65,7 +65,11 @@ namespace SpeckleRhino
 #endif
 
 #if MAC
-      CreateOrFocusSpeckle();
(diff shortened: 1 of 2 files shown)


teocomi (approved):

AlanRynne:
Tested in windows both opening from scratch in a new doc, opening an doc with existing streams, and switching docs. Everything worked as expected.

PR #1586, opened 2022-09-05
fix(dui2): uses old report log as fallback in new report

clairekuang:
## Description & motivation

The new DUI2 report relies on connectors to correctly implement `ApplicationObjects` and add log items to each object for reporting. However, some connectors have not switched to the new rpeorting system yet, and their conversion log should now be displayed as a fallback in the report tab (instead of no report)

## Changes:
 DUI2


AlanRynne (approved):
Nice! I'll wait for the CI to pass and merge it in

PR #1593, opened 2022-09-07
fix(core): Prevent invalid account from slipping out

AlanRynne:
Prevents invalid accounts from being included in the `GetAccounts()` result.


AlanRynne:
I'm an idiot, created the PR the wrong way...

PR #1596, opened 2022-09-08
fix (CSI) - receiving new Analytical models from revit

Reynold-Chan:
quick fix for receiving new Revit commits for analytical models

tested : manually


AlanRynne (approved):
:) Looks harmless enough! 🚀

PR #1603, opened 2022-09-09
fix(gh): Unload top-menu if it exists before loading

### LINKED ISSUE #1601: New menu tabs on GrasshopperReloadAssemblies
Hi, I'm doing a lot of plugin development where i need to run the GrasshopperReloadAssemblies command.

Turns out that every time speckle is loaded, it creates a new context menu. Should probably check if one exists of the same name before creating a new one. See below

image

AlanRynne:
Closes #1601

Instead of just skipping the menu creation, I'm deleting it and creating it again to ensure all events are properly set.


teocomi (approved):

izzylys (approved):
pog 🚀

Issue #1530, opened 2022-08-18
Investigate & add update mode cross-connectors


teocomi:
We should investigate how and where we can bring the **update functionality on receive** that we currently only have in Revit (I believe).

See: https://speckle.community/t/speckle-should-be-able-to-update-existing-objects-in-the-dwg-file/3424/2

clairekuang:
An update receive mode has been added for rhino, autocad, and civil3d!

teocomi:
Update:
- [x] Revit
- [x] Rhino
- [x] Autocad
- [x] Civil 3D
- [ ] Structural connectors ? @Reynold-Chan

clairekuang:
This seems to be closed - though we should document how this was implemented across each connector and how well round-tripping works for cross connector workflows

PR #1609, opened 2022-09-13
feat(rhino): adds update receive setting

clairekuang:
## Description & motivation

Adds an update receive mode for Rhino connector.
This will remove any existing objects on receive that have the same guid or application id as an incoming base object
part of #1530

## Changes:
 Rhino connector and converter

## Validation of changes:

Tested on rhino to rhino and revit to rhino basic workflows


AlanRynne (approved):
Looks good to me!

PR #1614, opened 2022-09-17
fix(rhino): stream edit view selection issues in rhino

teocomi:
**Description:**
Fixes 2 (unlogged) issues in DUI that were affecting the Rhino connector:
1. When the Speckle panel was closed and reopened and the user would click on a Stream, an exception was thrown about a visual parent being reused in Avalonia. This would also crash Rhino.
2. When copy-pasting across different Rhino documents (AND under some other unclear circumstances), the stream edit navigation would break. Nothing was happening when clicking on streams.

**Solution:**
1. A new view model is now generated each time the panel is initiated, instead of reusing one stored in the `SpeckleRhinoConnectorPlugin` class
   - note: now, if closing the panel and re-opening it, saved streams in DUI that have not been baked in the file will be lost, but this is probably a very rare use case
3. There was a duplicated `HostScreen` property in the `StreamViewModel` class! Deleting it seems to have solved the problem.

**Release:**
I've set to merge this in 2.9, but if required, we could cherry-pick the changes and backport them to 2.8 as well, I'll let @clairekuang decide 😊

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
@@ -21,7 +21,6 @@ namespace SpeckleRhino
     private static string SpeckleKey = "speckle2";
 
     public ConnectorBindingsRhino Bindings { get; private set; }
-    public MainViewModel ViewModel { get; private set; }
 
     internal bool _initialized;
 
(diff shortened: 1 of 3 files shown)


clairekuang (approved):
Tested locally and looks good!

I'd prefer this to be merged into main as a hotfix and then pulled into 2.9, since I've seen this affect other DUI2 connectors as well

teocomi:
The base was 2.9, so it might be easier to merge there and cherry pick to main, WDYT @AlanRynne ?

AlanRynne:
Let me do a local attempt and I'll see how bad it gets 😉

AlanRynne:
Rebased this branch onto `main`. Found no conflicts so I'm merging this in as soon as the CI passes (which it should...)

AlanRynne:
Failed rebase, some 2.9 features were pulled in during rebase, I'll have to cherry pick the commit and force push again.

PR #1616, opened 2022-09-19
SQLite library update

teocomi:
🚨 This PR touches a critical piece of Speckle 🚨

The `System.Data.SQLite` reference is being replaced with `Microsoft.Data.Sqlite` as it is also compatible with the new Silicon (arm64) Macs.

This simplifies the Core project as the new lib works out of the box in both OS.

 ☑️ Manual & unit tests are passing on Win and Mac (both x64 and arm64).

Up to @didimitrie and maybe @JR-Morgan if they want to do any benchmarking!


JR-Morgan:
Results of some quick performance testing of `Operations.Send`
Sending a total of `100_000` objects with a payload of `1_000` random chars each

We see the following results.
|  Method  |    Mean |   Error |  StdDev |
|--------- |--------:|--------:|--------:|
|  send2.9 | 44.63 s | 7.968 s | 2.069 s |
|  send2.8 | 42.96 s | 8.520 s | 2.210 s |

```
* Legends *
  Mean   : Arithmetic mean of all measurements
  Error  : Half of 99.9% confidence interval
  StdDev : Standard deviation of all measurements
  1 s    : 1 Second (1 sec)
  ```

This was with the following benchmark.net settings.

```cs
[SimpleJob(RunStrategy.Monitoring, launchCount: 1, warmupCount: 1, targetCount: 5)]
```

---

These measurements show a slight performance degradation of about 4%. Considering the high Error of both results, I'd put a lot of that difference down to run-to-run variance.

AlanRynne (approved):
👍🏼 not sure if i'm happy or nervous about _yet another SQL library_ hehehe but I approve this changes!

We disabled the TSEP build in Tekla since it was doing blocking this from being merged, and can be fixed separately.

PR #1618, opened 2022-09-20
Fix: Renames Structure.Material to StructuralMaterial

AlanRynne:
All our specific app/domain classes have are named with a prefix as to distinguish them from each other.

The `Material` class in the Structure objects was not doing this, causing issues with the SchemaGenerator in Grasshopper, and making it difficult to distinguish between the 2 on the Schema Builder pop-up.

This change addresses this issue, and regenerates the nodes from the TT template to reflect this change.


Reynold-Chan (approved):
looks good to me :)

PR #1620, opened 2022-09-20
GH: Fixes `ShallowCopy` issue with `displayValue` only containing 1 item

AlanRynne:
Fixes #

When`Base.ShallowCopy` was called, it will use `GetMembers`, which was not coded to exclude Obsolete properties, which would cause:

- `displayValue` to be set, but also
- `displayMesh` to be set, which will override the `displayValue` list

To fix this, we had to change the signature of `GetMembers` in a non-breaking way. It now accepts a set of `Flags` that allow you to control which members are included in the output. By default, it works as it used to.

Updated tests to check for these new cases inside `GetMembers`, and changed all hard coded property name strings to use `nameof()` instead so errors will pop up if properties are renamed. Previous tests were checking for _non-inclusion_ of a property that had been renamed, so they would always pass.

### DIFF
--- a/Core/Core/Models/DynamicBase.cs
+++ b/Core/Core/Models/DynamicBase.cs
@@ -186,6 +186,7 @@ namespace Speckle.Core.Models
     /// </summary>
     /// <returns></returns>
     public IEnumerable<PropertyInfo> GetInstanceMembers() => GetInstanceMembers(GetType()); 
+    
     public static IEnumerable<PropertyInfo> GetInstanceMembers(Type t)
     {
       PopulatePropInfoCache(t);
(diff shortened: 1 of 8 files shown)


Core/Core/Models/DynamicBase.cs
```diff
@@ -203,37 +204,40 @@ public static IEnumerable<PropertyInfo> GetInstanceMembers(Type t)
...
+    
     /// <summary>
-    ///  Gets the typed and dynamic properties that don't have a [SchemaIgnore] attribute.
+    ///  Gets the typed and dynamic properties. 
     /// </summary>
-    /// <returns></returns>
-    public Dictionary<string, object> GetMembers()
+    /// <param name="includeMembers">Specifies which members should be included in the resulting dictionary. Can be concatenated with "|"</param>
+    /// <returns>A dictionary containing the key's and values of the object.</returns>
+    public Dictionary<string, object> GetMembers(DynamicBaseMemberType includeMembers = DynamicBaseMemberType.Instance | DynamicBaseMemberType.Dynamic)
```
AlanRynne, line 215: Most critical change is here ☝🏼 and it's non-breaking. All unit test's passed in their original implementation; and also after fixing some non-related issues in them.

teocomi (approved):

PR #1621, opened 2022-09-20
Objects: Adds tests to CI job

AlanRynne:
Adds a step in `build-objects` to run the Objects unit tests (currently quite lacking).

Adds unit test to ensure **all objects in `ObjectsKit`** have an empty constructor implemented.

In order to not duplicate builds, `build-objects` will now only build:

- `Objects.csproj`
- `Tests.csproj`

All other projects correspond to `Converters` that are already being built on each of their respective `Connector` builds.

This also allowed to change the VM from windows to linux, which has less minute cost.


AlanRynne:
Disabled CodeCov upload. Should coordinate later on if we want to add these to the coverage, as we'd need to also add Core tests so that the coverage is "complete".

See current coverage report dropping from 40% to 1% because `Core` tests were not run (nor uploaded)

teocomi (approved):

PR #1628, opened 2022-09-26
feat(autocadcivil): interactive reports and update receive mode

clairekuang:
## Description & motivation

Adds two main features to the autocad and civil3d connectors: 
- Interactive reports for send and receive, aligning the autocad civil conenctors with functionality in rhino and revit
- Update receive mode that uses application ids of incoming objects to replace existing document objects if they share the same app id

Related to: https://speckle.community/t/speckle-should-be-able-to-update-existing-objects-in-the-dwg-file/3424/2
Part of: #1530

## Changes:

- Autocad Civil3d connector: adds implementation for `SelectClientObjs` binding as a prerequisite for interactive reports 

## Screenshots:

ezgif com-gif-maker (32)

## Validation of changes:

manually tested sending/receiving in Autocad <> Autocad and Revit > Autocad

### DIFF
--- a/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
+++ b/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
@@ -187,19 +187,20 @@ namespace Speckle.ConnectorAutocadCivil.UI
         {
           try
           {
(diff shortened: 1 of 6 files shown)


AlanRynne (approved):
🤘🏼 Yeah! This looks nice.

I'm starting to get worried about the massive size of our DUI bindings files, but I'll open a separate issue for that!

PR #1635, opened 2022-09-27
chore: Added paper-cut issue template

AlanRynne:
Added a papercut issue template.


AlanRynne:
Not sure how to verify this other than merging it into main, and since it won't break anything... here it comes!

PR #1649, opened 2022-09-28
feat(Dyn): Send/Receive tolerate conversion errors and reports them as node warnings

### LINKED ISSUE #1387: DYN: Receiver fails entire operation as soon as one object fails
Reported in the forum: https://speckle.community/t/dynamo-polycurves-may-be-branching/3164

Related to https://github.com/specklesystems/speckle-sharp/issues/1386

The Dynamo receive node stops conversion whenever a single item fails, and returns `null` as an output.

Grasshopper handles this on a "per-object" basis, which means that only the affected object will be returned as null, and a warning will be shown.

Dynamo should align with this behaviour.

AlanRynne:
Closes #1387

- [x] Adds a new `OnError` event to `BatchConverter` to report conversion errors while continuing with the conversion of other objects. The conversion will return `null`
- [x] Hook's up the new `BatchConverter` event to the `Receive` function.
- [x] Adds events to every place where `BatchConverter` is used. This includes:
    - Conversion nodes
    - Local transport nodes
    - Send nodee
    - Receive node
- [x] Improve error messages to include ID of the object that failed.
- [x] TEST, TEST, TEST!

Screenshot 2022-09-28 at 08 23 22

Bonus:

- Modified `MeshToNative` to use `TriangulateMesh` instead of `TriangulateFace` to remove duplication of code, since N-gons are not supported in Dynamo.

### DIFF
--- a/ConnectorDynamo/ConnectorDynamo/ReceiveNode/Receive.cs
+++ b/ConnectorDynamo/ConnectorDynamo/ReceiveNode/Receive.cs
@@ -15,6 +15,7 @@ using Speckle.Core.Api;
 using Speckle.Core.Api.SubscriptionModels;
 using Speckle.Core.Credentials;
 using Speckle.Core.Logging;
+using Speckle.Core.Models.Extensions;
 
 namespace Speckle.ConnectorDynamo.ReceiveNode
 {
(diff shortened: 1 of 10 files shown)


teocomi (approved):
Looked quickly through it, and it's good!
Let's make sure to test on older versions of Dynamo too.

AlanRynne:
Doing some extra testing today and I'll get this merged before hacktoberfest starts! 🚀

PR #1654, opened 2022-09-30
feat(objects): Completely removes `IDisplayMesh` and all it's implementations

AlanRynne:
IDisplayMesh has been obsolete for several releases now, so it's time to pull the plug.

Removes all references to `IDisplayMesh` from the `Objects` Kit, as well as any of the `displayMesh` properties.

Removes `IDisplayMesh` interface


Objects/Objects/BuiltElements/Topography.cs
```diff
@@ -21,23 +21,16 @@ public class Topography : Base, IDisplayMesh, IDisplayValue<List<Mesh>>
 
     public Topography()
     {
-      this.displayMesh = new Mesh();
+      displayValue = new List<Mesh>();
```
AlanRynne, line 24: @clairekuang, added you as a Reviewer on this, since my guess is that this is used in Rhino too.

This was the only usage of `displayMesh` I found, and it should work the same way with my implementation, except for the fact that it now creates an empty list instead of an empty mesh.

If this is ok with you, we can go on with it; but I'm happy to modify it if needed

Objects/Objects/BuiltElements/Topography.cs
```diff
@@ -21,23 +21,16 @@ public class Topography : Base, IDisplayMesh, IDisplayValue<List<Mesh>>
...
     }
 
     [SchemaInfo("Topography", "Creates a Speckle topography", "BIM", "Architecture")]
     public Topography([SchemaMainParam] Mesh displayMesh)
     {
-      this.displayMesh = displayMesh;
+      displayValue = new List<Mesh>{displayMesh};
```
AlanRynne, line 30: And here too, but this works just like is used to, now using displayValue directly instead of letting the `displayMesh` set the list.

JR-Morgan (approved):

clairekuang (commented):
Noticed some lingering deprecated `DisplayMesh` props in tekla object classes, otherwise seems fine

PR #1663, opened 2022-10-03
Civil3d/converter fix

teocomi:
Aligns the converter name for consistency and to make it work with the latest changes in `Applications.cs`

### Note
Goes hand in hand with this change in the installer: https://github.com/specklesystems/speckle-sharp-ci-tools/commit/8ac9091461238f5be18795d8da1e7c0daec55c06
We should not make new pre-releases for c3d on the 2.8 branch since the above has been merged already.


clairekuang (approved):
Looks good 👍

PR #1665, opened 2022-10-04
fix(ci): All mac connectors and Archicad for windows deployment

AlanRynne:
Fixes CI deploy jobs for mac, which were not uploading due to misconfiguration on the extension and expected file name.

**Mac deploys**:

- The output of our mac installer build would be called `rhino-mac.zip`, following the `slug`. The slug has now been changed to be universal (`i.e. `rhino` or `grasshopper`), and the deploy job uses the correct parameters for `os` and `arch`.
- Added `arch` parameter to the `deploy-connector-new` job
- Extension parameter was not being used in all necessary places.
- `deploy-connectors` job no longer uploads to the new manager. Instead, it's just used as a buffer step to wait for all `build` jobs to be completed before proceeding to deployment.
- Some changes were made on the `Speckle.Manager.Feed` CLI so that the job would fail if the upload is not successful.

**Python CircleCI config generator script**:

Since the slug no longer contains `-mac` suffix, we now use the `name` of the job if it exists to figure out if it's a mac build or not.

**Archicad windows installer**:

While fixing this we found out that the Archicad connector was not properly configured:

- Added `MSBuild` nuget to `ConnectorArchicad.csproj`
- Added `Properties/AssemblyInfo.cs`
- Added <GenerateAssemblyInfo> as `false`.

### DIFF
--- a/.circleci/scripts/config-generator.py
+++ b/.circleci/scripts/config-generator.py
@@ -177,11 +177,12 @@ def createConfigFile(deploy: bool, outputPath: str):
 
 def getNewDeployJob(jobName: str):
     slug = jobName.split("-build")[0]
-    isMac = slug.find("-mac") != -1
+    isMac = jobName.find("-mac") != -1
     deployJob: Dict[str, Any] = {
         "slug": slug.split("-mac")[0] if isMac else slug,
-        "name": slug + "-deploy",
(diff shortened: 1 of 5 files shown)


AlanRynne:
I'm merging this in as it has no changes in connector code, and the CI runs as expected

PR #1672, opened 2022-10-06
Etabs/sqlite fix

teocomi:
Fixes the `e_sqlite.dll` file not getting copied to the bin folder by default by adding:

`<PlatformTarget>x64</PlatformTarget>`


connorivy (approved):
🥳

PR #1686, opened 2022-10-10
feat(civil3d): receive TIN meshes as surfaces

### LINKED ISSUE #1634: feat(autocadcivil): send and receive surfaces as surfaces

Autocad / Civil3d converter

## Is your feature request related to a problem? Please describe.

Currently, the autocad / civil3d converter sends all surfaces and solids as meshes for convenience and ease of conversion. This is a request to send and receive surfaces as a `Surface` class object.

*Note*: iirc, many converters do not handle surface to native conversions well, and no other connector is sending surfaces alone (eg, rhino sends surfaces as breps). Receiving conversions will need to be checked in rhino/gh, revit, etc.

From a community PM request for Civil3D:

> Something for the wish list on the Civil 3D connector. When we send a Civil 3D surface to Speckle and try to bring it back into Civil 3D, we only receive a polyface mesh. It would be great if the mesh was being converted back into a surface during the import from speckle. 

## Describe the solution you'd like

Quoted from PM: 

> Ideally it would need to do the following tasks during the process:
> 1. use the same (or similar) surface name as when it was uploaded to Speckle
> 2. use the polyface as drawing object in the surface creation and make sure that we maintain edges from the objects (see screen capture below).
image
> 3. use the polyface extent to define an outer boundary (red boundary below) for the surface so the software doesn’t create additional triangles as shown in the blue area below
image
> The alternative would be to create the surface on the fly without the polyface but not sure what would work best in practice for data integrity.

clairekuang:
## Description & motivation

Sends Tin Surfaces as meshes with a tin surface class prop, and adds conversion method to receive meshes with tin surface class name as a tin surface. 

Includes minor fix to the `GetExistingElementsById()` method to ignore deleted objects. 
Fixes #1634 

## Changes:

AutocadCivil converter

## Validation of changes:

 Manually tested on community-contributed test file with a TIN surface. Needs additional testing to make sure it doesn't bork on various tin surface cases.

## Checklist:


AlanRynne (approved):
Ok, I haven't built or tested this, but since I'm going to do so anyway with the latest release I'll approve it.

Had a look through the code and it seems legit 😎

Also, I added a new issue in the docs to update the support table https://github.com/specklesystems/speckle-docs/issues/119

PR #1696, opened 2022-10-11
fix(rh/gh): Fixes build step to include SQLite dylib dependencies

teocomi:
Fixes sqlite issues in gh and hopefully rhino too


AlanRynne:
Does not fix the issue in Rhino, we get the following error when trying to initialise it.

`SystemNative_GetUnixName assembly:<unknown assembly> type:<unknown type> member:(null)` but since the Rhino connector is disabled in mac, we can go ahead and merge this, since it seems to not be the same issue.

The missing `libe_sqlite3.dylib` problem is now being copied correctly.

AlanRynne (approved):
Reviewed and tested ✅

PR #1703, opened 2022-10-11
Fix: Adds <PlatformTarget> property to all connector csproj files

AlanRynne:
Adds <PlatformTarget> to all connector projets that were missing it.

Moved it to a global position in all connectors that only had it on `debug`


AlanRynne:
This should do more good than harm... In it goes! 🚀

PR #1720, opened 2022-10-14
fix(acadcivil): fixes reporting in update mode and other minor bugs

### LINKED ISSUE #1710: ACAD Release 2.9 issues
- [x] `WasErased` state in report appears as `eWasErased`
image

- [x] Standard geometry file reports error in blocks but blocks but they seem to be baked into the document successfully.
image

- [x] Civil 3D title reads `Speckle for Civil3D 3D`
- [ ] Civil3D and ACAD use the wrong version (should use the semantic one that includes `beta3`)
Screenshot 2022-10-13 at 18 57 26

# Minor stuff
- [ ] Default filter should be `everything` (but we can leave this for 2.10, it was done to prevent accidental sending on massive Civil files that would crash the connector and civil...)
- [x] Blocks `update` behaviour is not implemented (and will be tricky to do so).

clairekuang:
## Description & motivation

Fixes the following:
- Incorrect error message on updating previously deleted elements
- refactors block to native method to return application object (also fixes reporting)
- civil3d dui2 name typo introduced by kit naming refactor

Closes #1710


AlanRynne (approved):
Nice!! Tested locally and everything seems to work.

Good job on the block updating behaviour ♥️

Issue #1718, opened 2022-10-14
DUI Search returns inconsistent results


AlanRynne:
When searching for a stream, the results will be empty at random times. Other times it will be correct. I didn't get to reproduce it but I also managed to get the same results **twice** at one point.

This seems to me like a problem of calling the search endpoint too eagerly. We should ensure that the search call is properly debounced, and previous search calls are discarded when a new search get's executed.

![2022-10-14 10 45 47](https://user-images.githubusercontent.com/2316535/195805984-0665949a-ff43-460a-961d-8c0264562f62.gif)

This affects all the connectors I've tested (ACAD, Civil, Rhino)

PR #1730, opened 2022-10-17
Dui/fix search

teocomi:
Attempts a fix for #1718 by adding a debouncer and cancellation token to the stream search function.
It's not perfect but I already spent too much time on this 😅

@AlanRynne give it a go pls, you can just run a standalone DUI to save time!


AlanRynne (approved):
It works! 🥳 I love the search bar again ♥️

PR #1736, opened 2022-10-18
fix(autocadcivil): get from id collection should check for `CanRead` prop

### LINKED ISSUE #1708: Bug: Error thrown when trying to import an alignment from a stream

Speckle Civil 3D Connector 2.9.0-beta3

## Describe the bug

Error thrown when trying to import an alignment from a stream

## To Reproduce

Start new civil drawing (drawing)
Create an alignment
Upload alignment to a new empty stream
Open up another new civil drawing (drawing2)
Receive the contents of the new stream into drawing2

## Expected behaviour

Alignment should have been inserted into drawing2

## Screenshots

image

## System Info

- OS: Windows
- Civil 3D 2022 
- Speckle Civil Connector 2.9.0-beta3
image

## Failure Logs

Failed conversion. eWasErased
Couldn't convert object or any fallback values
Could not bake object
```
![image](https://user-images.githubusercontent.com/67244422/195409469-0bade7fa-75b7-493d-ac85-a8e3f460231e.png)

## Additional context

<!---
Add any other context about the problem here.
-->

## Proposed Solution (if any)

No ideas!

#### Optional: Affected Projects

N/A

clairekuang:
## Description & motivation

The `GetFromObjectIdCollection()` util method in the acadcivil converter wasn't previously checking whether or not an existing prop has a getter. This was breaking alignment to native conversions when trying to retrieve the alignment style.

Method has been updated to return a defualt value if null, and also to check for `CanRead` before accessing an object prop.

Fixes #1708 

## Changes:

AutocadCivil converter: Utils and AlignmentToNative conversion

## Validation of changes:

Alignments received in new docs are now created without error.

### DIFF
--- a/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/Converter.AutocadCivil.Utils.cs
+++ b/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/Converter.AutocadCivil.Utils.cs
@@ -137,18 +137,19 @@ namespace Objects.Converter.AutocadCivil
       return ids;
     }
 
-    public ObjectId GetFromObjectIdCollection(string name, ObjectIdCollection collection)
+    public ObjectId GetFromObjectIdCollection(string name, ObjectIdCollection collection, bool useFirstIfNull = false)
     {
       var id = ObjectId.Null;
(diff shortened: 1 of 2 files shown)


AlanRynne (approved):
Nice! I'm merging it in so we can release `beta5` with this

PR #1762, opened 2022-10-21
Fix(Revit): add backup calc if structuralMatType of family is undefined

connorivy:
## Description & motivation

currently to set the structural material of an analytical member from Revit, we rely on the structuralMaterialType property of the family instance. I've found that this is sometimes "undefined". So I've added a backup way to find the structural material using the structural asset class of the family instance's material.

## Changes:

add second method of determining structuralMaterialType if the first method resulted in "undefined"


AlanRynne (approved):
I'm going to go ahead and approve this bc they seem like solid changes ;)

Nice fix @connorivy!

PR #1780, opened 2022-10-26
fix(ci): Adds `packandpublish-bash` for any bash jobs that require nuget pushing

AlanRynne:
Fixes an issue introduced in #1621 inadvertently.

The PR modified the `Objects` job to run on bash instead of powershell. This seemed like the right choice at the time, since it simplified setup, reduced credit cost and actually ran faster.

I did not realise at the time that our `packandpublish` command was written in `powershell` so it would throw a syntax error on any jobs running on `bash`.

## Solution

This PR fixes the issue by creating a specific job to pack and publish when using `bash` that follows the same logic.

Had to downgrade `ConverterDxf` language version usage to the default, as it was using version `9` and it was causing issues when building. Required some updates on the DXF tests to get them to work again too.

### DIFF
--- a/.circleci/scripts/config-template.yml
+++ b/.circleci/scripts/config-template.yml
@@ -98,7 +98,7 @@ jobs: # Each project will have individual jobs for each specific task it has to 
             HOMEBREW_NO_AUTO_UPDATE=1 brew install mono mono-libgdiplus
       - run:
           name: Install dotnet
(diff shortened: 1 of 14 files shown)


AlanRynne:
The only way to test this as of right now is to go ahead and release the nugets. So I'll create a tag on this branch before merging.

If passed, merging should be done **without squashing**

AlanRynne:
Preliminary CI runs were successful in building all nupkg files. Going ahead with this

PR #1781, opened 2022-10-26
chore: Update the release branch with changes in main

AlanRynne:
Just your normal PR to keep things up to date.

Includes several fixes to be released in `2.9.1`.


AlanRynne:
Not sure why, the CI ran the job twice, and when I cancelled the second one, it got stuck on "waiting" for the jobs that were already running.

Since the original job passed, I'm merging this in.

PR #1792, opened 2022-10-28
fix(acad): oneclick conflict

teocomi:
Fixes I conflict I didn't spot with the new oneclick send views


AlanRynne:
CI is having a day, seemed to not like my forcepush but was passing before so I'm merging in.

PR #1795, opened 2022-10-28
fix(rhino): Enable's rhino for mac connector with top-menu fix

### LINKED ISSUE #1277: MacOS: Top-menu dropdown's stop working when Speckle loads

AlanRynne:
Fixes #1277


AlanRynne:
IRL Reviewed with @teocomi

Issue #1793, opened 2022-10-28
Archicad: Mac DUI shows default `1.0.0.0` version


AlanRynne:
Caught by @gjedlicska.

I suspect this is caused by us not using `MSBuild.AssemblyInfo` nuget on mac for the Archicad build, as it was not capable of being built on Mac.

We should look into what is the alternative to this and add it. Our build already passes the `\p:Version` parameter, but it seems it's not enough.

![image](https://user-images.githubusercontent.com/2316535/198564048-2b17e715-1760-40e0-a785-1c479360ecc8.png)

PR #1796, opened 2022-10-28
fix(archicad): version patching issues #1793

teocomi:
Versions the dlls correctly on build.
Note: no MSBuild nuget/action is needed for .net core/standard projects. There was an old assemblyinfo.cs and a setting in the csproj file conflicting.


AlanRynne (approved):
Beautiful thx!! This was clearly me not fully understanding how MSBuild.AssemblyVersion affected our builds.

Now I think I got a clearer picture! 🙌🏼

PR #1800, opened 2022-10-28
fix(rhino): cleaning up version numbers

teocomi:
Starndardizes connector and converter projects to use the `csproj` version and not the `assemblyinfo` version, so that patching also works on mac.


AlanRynne (approved):
Thanks for this @teocomi! 🙌🏼

PR #1806, opened 2022-11-01
Feature(Revit): Support sending fabrication parts from Revit

connorivy:
feat(revit): Support sending fabrication parts from Revit

## Description & motivation

add support for sending fabrication parts to speckle as generic revit elements

## Changes:

- add new method for getting meshes from fabrication parts
- use this^ method in the generic element conversion if no meshes are found using the traditional method
- add basePoint or baseLine to the generic element if it is availible


teocomi:
@connorivy we should also bring 2.10 in this branch before merging, some tests are failing

teocomi (approved):
Looks good, but could not check ConvertMeshUtils.cs given the line changes

PR #1815, opened 2022-11-05
fix(dui): close notification button crashes the app

teocomi:
An optional parameter on the close notification command binding was making Avalonia crash.


AlanRynne (approved):
Looks good to me! I've reset the `ci-tools` repo to a working state so this should be good to be released

PR #1825, opened 2022-11-09
feat(core): fallback to default Objects folder if no converter is found

teocomi:
**🙌 this PR touches delicate parts of Speckle!**

The suggested changes were discussed with Dim and allow us to safely reference `Objects` in DUI for us to unlock some features in the Mapping tool ([see](https://www.notion.so/speckle/CAD-to-BIM-4bde12f16ab847c788b17bce9b55fa25#79ec640a6d27482dbcfa21cffb4d8bff)).

What this does:
- if Objects.dll is already loaded in the AppDomain, currently, the Kit only looks for converters in the same directory
- if Objects.dll is also referenced by DUI (or any other plugin used by the host app) there will not be any converters in the same folder (because it will be in the connector folder instead)
- these changes allow falling back to the default Kits/Objects folder in the above scenario

NOTE: while DUI will reference Objects for use in the Mapping tool, we'll take measures not to ship it with DUI to avoid even further versions misalignments etc.

### DIFF
--- a/Core/Core/Kits/KitManager.cs
+++ b/Core/Core/Kits/KitManager.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -22,10 +22,10 @@ namespace Speckle.Core.Kits
     {
       get
       {
-        if (_kitsFolder != null)
-          return _kitsFolder;
+        if (_kitsFolder == null)
+          _kitsFolder = Path.Combine(Helpers.InstallSpeckleFolderPath, "Kits");
 
-        return Path.Combine(Helpers.InstallSpeckleFolderPath, "Kits");
+        return _kitsFolder;
       }
       set
       {
--- a/Objects/Objects/ObjectsKit.cs
+++ b/Objects/Objects/ObjectsKit.cs
@@ -1,4 +1,5 @@
 ﻿
+using Speckle.Core.Api;
 using Speckle.Core.Kits;
 using Speckle.Core.Logging;
 using Speckle.Core.Models;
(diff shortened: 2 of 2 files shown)


Core/Core/Kits/KitManager.cs
```diff
@@ -22,10 +22,10 @@ public static string KitsFolder
     {
       get
       {
-        if (_kitsFolder != null)
-          return _kitsFolder;
+        if (_kitsFolder == null)
+          _kitsFolder = Path.Combine(Helpers.InstallSpeckleFolderPath, "Kits");
 
-        return Path.Combine(Helpers.InstallSpeckleFolderPath, "Kits");
+        return _kitsFolder;
```
teocomi, line 28: This is not really required, but it seemed the lazy loading wasn't done correctly.

AlanRynne:
I'm arriving late at this thread, and I'm not aware of the full scope of the limitations you've encountered in DUI + Mappings, so I may need you to walk me through this @teocomi. But here's my 2 cents:

I think I get where this is coming from, and these changes by themselves seem harmless, but the idea of "blurring the lines" between the kit and the connectors does scare me a bit 😅 

If we go down this road, it will then be impossible to support any kit other than ours in the future, as they'd be baked into the connector layer. In my opinion, anything "object" related should be created in a way that it could be "pulled out" and replaced with something different (a completely different kit) without the connector breaking in the process.

Wouldn't it be better to come up with some other strategy that doesn't break that separation of concerns between Connector and Converter?

AlanRynne (approved):
After our chat with Matteo, I'm waving my magic wand and approving this 🪄 😉

I'm also going to open a backlog issue to revise this decision once the Mappings feature is fully understood and we have a better picture of what are the needs of it, and how could that be abstracted into something better than directly referencing `Objects` in `DUI`.

PR #1830, opened 2022-11-10
Fix(gh):  Handle `GH_Goo<Base>` objects in `ESO`

### LINKED ISSUE #1829: GH: ESO node fails to handle `GH_Goo<Base>` objects as input
From the forum: https://speckle.community/t/speckle-base-deserialization-issue/4074/3

The `ESO` node has a missing verification step that existed in `DSO`. Without that, it will fail to recognise any `GH_Goo<Base>` (or `GH_ObjectWrapper` containing a `Base` object) as a valid input, and would refuse to run.

Our own `GH_SpeckleBase` is already being handled properly. This only affects any other 3rd party components that do not use this class as it's not accessible to them (it lives inside the connector project)

AlanRynne:
Fixes #1829 

Aligns the `ESO` input verification with the `DSO` node, which already included a provision to test for `Base` as a value of any `IGH_Goo`.

This issue was not a problem within our own components, but a result of 3rd party nodes generating `GH_ObjectWrapper` classes containing base objects.


AlanRynne:
It's a 3 line change in a single component. I'm approving this myself.

PR #1841, opened 2022-11-11
feat(gh): Prevents empty meshes from being sent

### LINKED ISSUE #1324: Rhino: Prevent meshes with no vertices/faces from being sent
This is a weird edge case, Namita found it while prepping the City model.

You can send a mesh with 0 vertices and 0 faces, but when receiving it back in Rhino it will fail to convert.

Best way to approach this I think would be to just not send them, and add this to the report.

AlanRynne:
Fixes #1324 

When encountering a mesh with no faces or vertices, we will now return `null` instead of sending a mesh that won't work anywhere.

This has been tested in Rhino and Grasshopper and does not interfere with the send operation.

Screenshot 2022-11-11 at 21 25 10

Screenshot 2022-11-11 at 21 23 11


didimitrie (approved):

fraguada:
@AlanRynne can you update the status of this PR if it has been tested?

AlanRynne:
I left it there so that we wouldn't forget to add it to the testing doc, and obviously ensure this case works from now on.

PR #1865, opened 2022-11-17
Feat(Revit): Project previews

connorivy:
## Description & motivation

This feature allows the user to preview the commit that they are about to receive as non-interactable meshes in the Revit modeling space before going through the expensive conversion of each element. I already am working on a way to greatly improve the performance of this feature. Currently it makes the modeling space a bit laggy when a large commit is previewed so it should be considered an alpha feature.

Also there is a known issue where the previews do not look right in the "shaded" view. I suspect this is due to the way that I am calculating the mesh normals, but I haven't tried to fix that. I could just turn the previews off for that setting or leave them where they don't really work for fine meshes.

## Changes:

- implemented Revit's IDirectContext3DServer to create multiple "servers" and use those to generate preview geometry

## Validation of changes:

- only some manual testing. Definitely needs a little more attention during the testing process.

## Checklist:


Objects/Converters/ConverterRevit/ConverterRevitShared/DirectContext3DServer.cs
```diff
@@ -0,0 +1,461 @@
...
+using Objects.Converter.Revit;
+using Speckle.Core.Models;
+using OG = Objects.Geometry;
+
+namespace Objects.Converter.Revit
+{
+  public partial class ConverterRevit
+  {
+    public static ConverterRevit Instance;
+    public class DirectContext3DServer : IDirectContext3DServer
```
teocomi, line 19: Hope you found most of this somewhere as it doesn't look fun 😅

teocomi (approved):

PR #1873, opened 2022-11-18
fix(revit): bug fixes for receiving mep networks

clairekuang:
## Description & motivation

Fixes some issues receiving mep networks in Revit, including:

- removing `JsonIgnore` flags on revit network link tags as these are actually set on send and not on receive
- `Conduit` bug handling
- omitting some extra `Pipe` props on send to prevent incorrect elevations on receive

## Changes:
- Objects
- Revit converter


connorivy (approved):

clairekuang:
replacing this with a clean pr

PR #1896, opened 2022-11-22
Fix: revit analytical stick conversion

dtnaughton:
## Description & motivation

- Handling of null material asset on conversion (occurs when material without physical properties is assigned to analytical member)
- Fix conversion of analytical restraints

## Changes:

See above

## Validation of changes:

No tests added or amended. Tested manually.

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertAnalyticalStick.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertAnalyticalStick.cs
@@ -18,8 +18,8 @@ namespace Objects.Converter.Revit
     public ApplicationObject AnalyticalStickToNative(Element1D speckleStick)
     {
       ApplicationObject appObj = null;
-      XYZ offset1 = VectorToNative(speckleStick.end1Offset ?? new Geometry.Vector(0,0,0));
-      XYZ offset2 = VectorToNative(speckleStick.end2Offset ?? new Geometry.Vector(0,0,0));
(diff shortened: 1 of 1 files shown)


teocomi:
Hey @dtnaughton thanks for this!

Afraid we fixed it in another PR before we could review yours: https://github.com/specklesystems/speckle-sharp/pull/1931

Can you please have a look and see if that works for you? It should also address a similar issue for the AnalyticalSurface.

dtnaughton:
> Hey @dtnaughton thanks for this!
> 
> Afraid we fixed it in another PR before we could review yours: #1931
> 
> Can you please have a look and see if that works for you? It should also address a similar issue for the AnalyticalSurface.

Hey @teocomi - I took a brief look. #1931 does fix it in terms of preventing an exception, but if I'm understanding it would not assign a material as null to the analytical stick/surface. The fix in this PR (though only for stick) would at least provide a material with the information that can be fetched from Revit e.g. materialType which I think it important, disregarding properties relying on the material asset.

Secondly, I think we should still absorb the analytical releases change in this PR as I don't believe that is working as intended currently. I should have split this into two PRs - I apologize for that.

dtnaughton:
Hey @teocomi - after a bit of extra thought, I think it's preferable to have the check on the invalid structural element id at the level of code which #1931 has, rather than buried in the GetStructuralMaterial method as in this PR. Can we reconcile the logic and have an else statement to fetch the basic material properties that do not rely on material asset (and only rely on material type) in this case?

teocomi:
Thanks @dtnaughton , @connorivy is on this :) will keep you posted

PR #1897, opened 2022-11-22
Fix(CSI): Invalid cast exception

connorivy:
## Description & motivation

very simple exception fix

## Changes:

- remove unsafe cast


teocomi (approved):
Haven't tested this, but looks simple enough :)

PR #1898, opened 2022-11-23
Fix(CSI): refresh objects list every time we send to Speckle

connorivy:
## Description & motivation

When receiving a commit into Speckle and then trying to send it to Speckle again, my speckle commits were not showing any displayable objects in them. Upon further investigation I found that the master list of object IDs that the send operation then narrows down to your specified filter (Everything, Selection, Group, etc.) is only refreshed in the GetSelectionFilters() method which is only called when a user opens a stream view. Therefore between the actions of receiving and sending, the master list of elements is never refreshed (unless the user goes back to the main list of streams and re-opens the stream view) which makes the send operation show 0 visible elements.

It sounds reasonable to assume that we could maybe only call this method once when the user opens the stream view, as most likely not many people are going to be receiving and then immediately sending. However it could certainly happen by a new user just testing out Speckle. Also, if a user closes the plugin and then reopens it during the same session, the plugin will open up to the page that it was last on, which could be the stream page. That means that a user could receive a commit, close the plugin and make some additions to the model, and then try to send the model again, but Speckle would not recognize the new objects that were added. 

This change makes it so the master list is refreshed during the send operation before the list is needed. 

## Changes:

- remove null constraint for refreshing the ObjectIDsTypesAndNames dictionary

## Validation of changes:

- manual sending and receiving

## Checklist:

### DIFF
--- a/ConnectorCSI/ConnectorCSIShared/UI/ConnectorBindingsCSI.Selection.cs
+++ b/ConnectorCSI/ConnectorCSIShared/UI/ConnectorBindingsCSI.Selection.cs
@@ -88,8 +88,7 @@ namespace Speckle.ConnectorCSI.UI
         case "manual":
           return GetSelectedObjects();
         case "all":
-          if (ConnectorCSIUtils.ObjectIDsTypesAndNames == null)
-            ConnectorCSIUtils.GetObjectIDsTypesAndNames(Model);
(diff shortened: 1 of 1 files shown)


teocomi (approved):
Haven't tested this, but looks simple enough :)

Issue #1915, opened 2022-11-23
Bug(Revit): Family instances flagged as MEP elements not sending to Speckle


connorivy:
## Prerequisites

- [x] I read the [contribution guidelines](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)
- [x] I checked the [documentation](https://speckle.guide/) and found no answer.
- [x] I checked [existing issues](../issues?q=is%3Aissue) and found no similar issue. 
- [x] I checked the [community forum](https://speckle.community/) for related discussions and found no answer.
- [x] I'm reporting the issue to the correct repository (see also [speckle-server](https://github.com/specklesystems/speckle-server), [speckle-sharp](https://github.com/specklesystems/speckle-sharp), [specklepy](https://github.com/specklesystems/specklepy), [speckle-docs](https://github.com/specklesystems/speckle-docs), and [others](https://github.com/orgs/specklesystems/repositories))

## What package are you referring to?

Revit Converter

## Describe the bug

Previous the contents of the Revit model [here](https://we.tl/t-IcsBS0Lixx) (Revit 2022) would send to Speckle as family instances correctly, but now the elements are getting flagged as MEP elements (and they are) but they are not correctly sending to Speckle.

Here is the error that the report is showing for both elements
![image](https://user-images.githubusercontent.com/43247197/203632733-cb4fb5ca-b951-41eb-a75f-8228eacb5e6f.png)

## To Reproduce

Download the revit file -> https://we.tl/t-IcsBS0Lixx
Open in Revit 2022
Send all 4 objects to Speckle
The 2 MEP elements fail to convert

## Expected behavior

All elements convert successfully

### Desktop (please complete the following information):

- OS: Windows11
- Revit Version: 2022
- Version : 2.10.0-rc3

bimgeek:
Closing this ticket as we are moving to Jira to reduce noise. You can access it there.

PR #1917, opened 2022-11-23
Fix(revit): MEP elements handle invalid shape

connorivy:
## Description & motivation

Fixes error that was stopping Revit elements from being sent to Speckle

## Changes:

- add initial value of 0 to connection pair dimensions
- don't try to set dimensions if the connectorProfileType is invalid

## Validation of changes:

- was getting an error that said "The shape of the connector is neither rectangular nor oval." but now we are not getting that error

## References

this partially fixes this issue https://github.com/specklesystems/speckle-sharp/issues/1915

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Revit/ConnectionPair.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Revit/ConnectionPair.cs
@@ -18,7 +18,7 @@ namespace ConverterRevitShared.Revit
         Height = Connector.Shape != ConnectorProfileType.Round ? curve.Height : 0;
         Width = Connector.Shape != ConnectorProfileType.Round ? curve.Width : 0;
       }
-      else
+      else if (Connector.Shape != ConnectorProfileType.Invalid)
       {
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Revit/ConnectionPair.cs
```diff
@@ -36,11 +36,11 @@ private ConnectionPair(Element element, Connector connector, Connector refConnec
...
       : $"{Owner.Category.Name}: {Connector.Owner.Name} --> NULL";
 
-    public double Diameter { get; private set; }
+    public double Diameter { get; private set; } = 0;
 
-    public double Height { get; private set; }
+    public double Height { get; private set; } = 0;
 
-    public double Width { get; private set; }
+    public double Width { get; private set; } = 0;
```
chuongmep, line 43: Default value of property double is 0, we don't need add 0 after that.

teocomi (approved):

clairekuang (approved):

clairekuang:
A note: this is a partial fix as these mep family instances are now being received, but hosted instances are received as rotated and moved, and are no longer hosted to the parent non-mep family instance

clairekuang:
image

PR #1922, opened 2022-11-24
fix(rhino): incorrect arg type for ConvertObject method

clairekuang:
Fixes a bug introduced with the rhino block bug pr where `ConvertObject` in the connector was expecting a convertible `Base`, not a `ApplicationObject`

## Changes:

- Rhino Connector


JR-Morgan (changes requested):
doesn't work... investigating...

JR-Morgan (approved):

PR #1924, opened 2022-11-25
Archciad/#1907 archicad connector receive does not work

jozseflkiss:
We have two options to load kits dlls:
1. Searching and loading them from the execution processes' referenced assemblies (`GetLoadedSpeckleReferencingAssemblies` function) or
2. Loading them from the KIts folder ("C:\Users\...\AppData\Roaming\Speckle\Kits\Objects") via `LoadSpeckleReferencingAssemblies`

The supported way is No 2, because by design the connector plugin mustn't know about the Kit is used, Kit dlls should be dynamically loaded during the conversion process.

For the Archicad Connector we have a shortcut here (because we have no Converter dlls yet), we use No 1.
The problem was that the original AppDomain.CurrentDomain.GetAssemblies only returns the assemblies loaded into memory, and at that point no Objects.dll was loaded. Now a new enumeration was implemented which recursively determine the referenced assemblies, and for not to load huge amount of dll we skip and don't load systems assemblies.

Good document in this topic: https://dotnetcoretutorials.com/2020/07/03/getting-assemblies-is-harder-than-you-think-in-c/

### DIFF
--- a/ConnectorArchicad/ConnectorArchicad/Converters/ElementConverterManager.cs
+++ b/ConnectorArchicad/ConnectorArchicad/Converters/ElementConverterManager.cs
@@ -248,7 +248,7 @@ namespace Archicad
 
         var subElements = currentObj["elements"] as List<Base>;
 
-        if (subElements.Count() > 0)
+        if (subElements != null && subElements.Count() > 0)
         {
           var convertedSubElements = await ConvertSubElementsToNative(subElements, token);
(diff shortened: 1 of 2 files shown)


teocomi:
Great find! It looks good to me, but I'd ask for a  blessing from @didimitrie, who wrote this logic in the first place

didimitrie (approved):

PR #1929, opened 2022-11-25
fix(core): null check for GetReferencedAssemblies

teocomi:
Seems like the new logic added in [#1924](https://github.com/specklesystems/speckle-sharp/pull/1924) was not working well in Revit, as some of the loaded assemblies were null.

This nul check seems to do the job.


JR-Morgan:
Looks like this is broken in Rhino too! 
null check is welcome.

PR #1931, opened 2022-11-26
fix(revit): adds null checks for analytical model in revit 2022 and b…

teocomi:
fix: adds null checks in the analytical model conversion that were preventing some elements from sending in Revit 2022 and below


teocomi:
@connorivy please have a looksie when you’re back to ensure it’s a proper fix.
To repro, just send all analytical members from the sample arch model 2022.

PR #1932, opened 2022-11-27
fix(rhino): Fixed regression converting elements from connector bindings

JR-Morgan:
Fixed the regression introduced by #1916

I've commented the code with a short explanation of the problem

targeting #1922 for @clairekuang to review

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
@@ -238,6 +238,9 @@ namespace SpeckleRhino
 
     #region receiving 
     public override bool CanPreviewReceive => true;
+
+    private static bool IsPreviewIgnore(Base @object) => @object.speckle_type.Contains("Block") || @object.speckle_type.Contains("View");
+    
     public override async Task<StreamState> PreviewReceive(StreamState state, ProgressViewModel progress)
     {
       // first check if commit is the same and preview objects have already been generated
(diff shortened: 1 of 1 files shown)


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
```diff
@@ -386,32 +389,41 @@ public override async Task<StreamState> ReceiveStream(StreamState state, Progres
...
-        // flatten the commit object to retrieve children objs
-        int count = 0;
-        Preview = FlattenCommitObject(commitObject, converter, progress, commitLayerName, ref count);
-
-        // convert
-        foreach (var previewObj in Preview)
+        // create preview objects if they don't already exist
+        if (Preview.Count == 0)
         {
-            var storedObj = StoredObjects[previewObj.OriginalId];
```
JR-Morgan, line 399: This line was sometimes throwing an exception, as `previewObj.OrianlId` is not in `StoredObjects`

Previously (rc3), this lookup wouldn't happen until Convert was called. And wouldn't happen for paths where the object is not convertible.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
```diff
@@ -386,32 +389,41 @@ public override async Task<StreamState> ReceiveStream(StreamState state, Progres
...
+          Preview = FlattenCommitObject(commitObject, converter, progress, commitLayerName, ref count);
+
+          // convert
+          foreach (var previewObj in Preview)
+          {
+            if (previewObj.Convertible)
             {
-              previewObj.Update(status: ApplicationObject.State.Failed, logItem: $"Couldn't retrieve stored object from bindings");
-              continue;
+              var storedObj = StoredObjects[previewObj.OriginalId];
```
JR-Morgan, line 404: Moving this line inside the `if(previewObj.Convertable)` block fixes this!

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
```diff
@@ -238,6 +238,9 @@ public override void ResetDocument()
 
     #region receiving 
     public override bool CanPreviewReceive => true;
+
+    private static bool IsPreviewIgnore(Base @object) => @object.speckle_type.Contains("Block") || @object.speckle_type.Contains("View");
```
JR-Morgan, line 242: additionally, I noticed that this check was being performed in 3 places. so extracted a method.

clairekuang:
ooh ty for the catch!

PR #1953, opened 2022-11-30
fix(core): creates Speckle folders if they do not exist

teocomi:
- fixes: https://github.com/specklesystems/speckle-manager2/issues/16
Ensures the special folders used by Accounts and kits are created if missing, by default, if they do not exist, the path returned is empty.


AlanRynne (approved):
Looks good! Great catch @teocomi 🙌🏼

PR #1954, opened 2022-11-30
fix(rhino): no longer using commit top layer in update receive mode

### LINKED ISSUE #1836: bug(rhino): Rhino creates a new layer for each receive
## Description

For each model version received from Speckleverse, a new layer structure is created in Rhino. Let's say a version of a Revit model is received. Rhino creates the layer structure for this version. Later, the Revit model was updated and now a new version will be received. When the new version is received, a layer structure is created **again** next to the previous one. In other words, **when 10 versions are received, users end up with 10 separate layer structures.** This is an undesired and unexpected problem from a user's pov.

## Expected Behaviour

Create a layer for a model and track versions through this single layer structure.

## Steps to reproduce

1. Open Rhino and enable Speckle connector.
2. Add below model by URL.
3. Receive first version. Rhino will create the layers.
4. Receive the second version. Rhino will create another layer structure.

> If you don't want to follow these steps, you can watch the video recording below.
> 

## Streams/Commits Used

[https://latest.speckle.dev/streams/1fedfb65da/branches/main](https://latest.speckle.dev/streams/1fedfb65da/branches/main)

## Screenshots

https://user-images.githubusercontent.com/51519350/201307939-1d9b1441-b1c4-4f6b-96ad-473f2aac7966.mp4

image

clairekuang:
## Description & motivation

Previously, receiving in Rhino always creates a top layer with the stream, branch, and commit info, except in update mode where the received object is placed in the same layer as the object it is replacing.

Now, the top commit layer will only be created in `Create` receive mode.
In `Update` mode, objects will be placed in the layer structure as determined by the commit traversal. This means any object that is replacing an existing object in the rhino file will no longer be placed in the existing object's layer.

Fixes #1836 

## Changes:

Rhino Connector

## Screenshots:

Before:
image

After (update mode):
image

After (create mode):
image

## Checklist:


clairekuang:
closing, opened a clean pr for 2.10

PR #1955, opened 2022-11-30
fix(acadcivil): commit prefix no longer used for layers on update receive mode

### LINKED ISSUE #1899: feat(acad): received layer names are way too long
## Description

When a commit is received in Autocad, the layer naming convention is as follows:

`Stream Name [ branch_name @ commit_id ] Layer Name`

This is especially inconvenient when a user receives multiple commits in the same drawing. They’ll end up with duplicate layers since layers are created on a commit basis.

Take a look at the recording below👇:

## Expected Behaviour

Layer name should be a lot shorter. Only layer name is enough. Can’t see any use case for stream-branch-commit names/ids. New layer naming should only contain `Layer Name`.

Speckle should NOT create new layer if there is already a layer with the same name.

## Steps to reproduce

1. Open the Autocad test file and send it to Speckle.
2. Open a new drawing and receive this commit you just sent. (or any other commit from any other app)
3. Take a look at the Layers.

## Streams/Commits Used

[https://latest.speckle.dev/streams/3ed8357f29/branches/standard/meters](https://latest.speckle.dev/streams/3ed8357f29/branches/standard/meters)

## Files Used

Autocad Test File

[https://drive.google.com/open?id=1-Ci9fX0xfOgtMy0phTIlqzIZeqMV72pM&authuser=bilal%40speckle.systems&usp=drive_fs](https://drive.google.com/open?id=1-Ci9fX0xfOgtMy0phTIlqzIZeqMV72pM&authuser=bilal%40speckle.systems&usp=drive_fs)

## Screen Recordings

https://user-images.githubusercontent.com/51519350/203483447-12c29073-579c-4643-a9b7-e3f4df7f3c36.mp4

clairekuang:
## Description & motivation
Aligning update receive behavior with Rhino, this pr removes the `stream branch commit` info as a prefix to received layers on update mode. This means that objects roundtripping in autocad on update mode should be placed in the same layers.

Receiving commits on create mode will still create new layers.

Also includes a minor fix for blocks that were being reported as updated instead of created

Fixes #1899

## Changes:

Autocad connector

## Screenshots:

Before on update mode:
image

After on update mode:
image

## Checklist:

### DIFF
--- a/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
+++ b/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
@@ -359,7 +359,7 @@ namespace Speckle.ConnectorAutocadCivil.UI
 
           // flatten the commit object to retrieve children objs
           int count = 0;
-          var commitObjs = FlattenCommitObject(commitObject, converter, progress, commitPrefix, ref count);
+          var commitObjs = FlattenCommitObject(commitObject, converter, progress, ref count);
 
           // open model space block table record for write
           BlockTableRecord btr = (BlockTableRecord)tr.GetObject(Doc.Database.CurrentSpaceId, OpenMode.ForWrite);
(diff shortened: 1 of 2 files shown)


clairekuang:
Including this fix in #1965

PR #1958, opened 2022-11-30
feat(dui): adds support for hiding the receive tab

teocomi:
Some connectors cannot receive, eg Navisworks, this PR adds a virtual binding prop to specify if Receive should be enabled or not.
This PR also fixes the "advanced settings" button not being hidden when no settings were available for the connector.


jsdbroughton:
👍image

Beyond the scope perhaps, but are the advanced settings ever likely to partition Send and Receive options?

jsdbroughton (approved):

teocomi:
Txs!
> Beyond the scope perhaps, but are the advanced settings ever likely to partition Send and Receive options?
Yes, we can definitely do so!

PR #1962, opened 2022-12-01
Matteo/notifications hotfix

teocomi:
- subscriptions not correctly disposed when a stream is not saved and the user simply clicks on the back button
- disables the file import/export alert in Revit


AlanRynne (approved):
👍🏻 I'll open a new issue to delete the leftovers for `2.11`, and we can release this with the next hotfix with minimum changes

PR #1971, opened 2022-12-02
Archicad Connector: codestyle conformability

jozseflkiss:
Code style reformat:
* for C# code
  * our common codestyle is used (.editorconfig copied from the Core module)
* for C++ code
  * new .editorconfig is created which is conform with the Archicad API

There should not be any code change, only whitespace changes.


teocomi (approved):

AlanRynne (changes requested):
Looks good, but this **should be merged into `release/2.11`** afaik, so setting this to "request changes", although no code changes seem necessary, just the branch target.

Since this can be a delicate operation and sometimes screw things up, I'll let @jozseflkiss change the target branch since he'll be most likely to identify what's wrong.

AlanRynne (approved):

PR #1975, opened 2022-12-05
Feat(CSI): Add report

connorivy:
## Description & motivation

the report functionality up to this point has not been working in CSI. Implementing the report and saving all the received Application Objects is the first step to implementing proper update behavior.

## Changes:

- eliminate the support of the objects.structural.analysis.model class on receive. This class is purely for organization and it was preventing the commit from being properly flattened.
  - Had to change flatten function to traverse through all object properties, not just dynamic props
- change the convertToNative function to return application objects
- change every "object"ToNative function to edit the application object and update it's status after conversion
- remove project references that were causing circular references
- misc code cleanup

## Screenshots:
Before:
image

After:
image

## Validation of changes:

received several commits in both ETABS and SAP

## Checklist:


Reynold-Chan:
This is a pretty big change in terms of the logic that is handling conversion logic. The big change being that we're getting rid of the model class which has historically been used to encompass the whole model. Currently other connectors like GSA are sending an analysis model which encompasses everything else. Adding in @dtnaughton for further discussion. If we ourselves want to get rid of the model class because it's causing more trouble, we need a further discussion on why we're not using it to traverse the logic.

Reynold-Chan (dismissed):
Need to implement a method still for ModelToNative at least to handle other people's model. Or we need to discuss with the community why we want to get rid of this class. It affects the GSA connector from Arup quite a bit if we do this.

connorivy:
@Reynold-Chan we're not getting rid of the model class (that was poor wording on my part), we are just removing it from CanConvertToNative so that we can flatten it. The model that I received above is a "Model" and it still works. By removing the model class from the CanConvertToSpeckle method, the traverse function will go through it an find all the elements that can be converted (element1Ds, nodes, materials, properties, etc.). 

The reason that I made this change is because previously the flatten function on the converter side would only return 1 object, a Model. This was a bit of an issue because that was the only object that the connector knew about, so it was the only item showing up in the report. By traversing the Model on the connector side instead of the converter side, we are able to see all the child elements that we care about in the report.

So my understanding is that I've just moved the traversal of the Model class from the converter to the connector and that this change isn't breaking at all, but defintiely let me know if you see some issue that I don't.

Reynold-Chan:
ah okay then ! go ahead

Issue #1944, opened 2022-11-29
RevitConverter: Material quantities add length of elements


ks-cph:
Related to issue #1302 and PR #1314 
#1314 implements the `material_quantities` to the revit connector. `material_quantities` contains the volume and area of each `Material` in an element. 

Sometimes it is alos essential to have the length of an element divided by `Material` Thus, I would like to add another parameter to material quantities.  Besides `area` and `volume` it should also have a `length`.

Please let me know if you see any problem with that?

Cheers,
Kilian

teocomi:
I don't see issues with it, please go ahead :)

fraguada:
@connorivy @AlanRynne @ks-cph  what is left to do to close #1976

connorivy:
@fraguada The ball is in my court. I haven't been working on Revit at all this sprint which is why this hasn't been closed yet. I just made my last structural PR that I'm going to make this sprint so I'll work on this now.

PR #1976, opened 2022-12-05
Update: Length for materialQuantities in RevitConverter

ks-cph:
## Description & motivation

- Add length of elements to MaterialQuantities
- Allows for query of material quantities
- fixes issue #1944 

## Changes:

- ```Objects.Converter.Revit.ConverterRevit```
- ```Objects.Other.MaterialQuantity```

## Screenshots:

Two differnet elements before(l.) and after(r.)

## Validation of changes:
I have testes it with Revit projects and different element categories.

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
@@ -25,16 +25,18 @@ namespace Objects.Converter.Revit
       //Get quantities
       double volume = element.GetMaterialVolume(material.Id);
       double area = element.GetMaterialArea(material.Id, false); //To-Do: Do we need Paint-Materials
-
+                                                                 //length
+      double length = 0;
+      if (LocationToSpeckle(element) is ICurve) length = (LocationToSpeckle(element) as ICurve).length;
       //Convert Feet to meters
       string units = Speckle.Core.Kits.Units.Meters;
       double factor = Speckle.Core.Kits.Units.GetConversionFactor(Speckle.Core.Kits.Units.Feet, units);
       volume *= factor * factor * factor;
       area *= factor * factor;
-
+      length *= factor;
       //Create and return materialquantity
       var speckleMaterial = ConvertAndCacheMaterial(material.Id, material.Document);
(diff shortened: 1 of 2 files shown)


teocomi (changes requested):
Thanks @ks-cph , this looks good.
Although since we're working with release branches, this would need to be started from the `release\2.11` branch and merged there, instead of using `main`.
Can you please do so?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -25,16 +25,18 @@ public Objects.Other.MaterialQuantity MaterialQuantityToSpeckle(DB.Element eleme
       //Get quantities
       double volume = element.GetMaterialVolume(material.Id);
       double area = element.GetMaterialArea(material.Id, false); //To-Do: Do we need Paint-Materials
-
+                                                                 //length
+      double length = 0;
+      if (LocationToSpeckle(element) is ICurve) length = (LocationToSpeckle(element) as ICurve).length/1000;  //LocationCurve default is mm (Speckle)
```
AlanRynne, line 30: Not sure about this line. I'll assign @connorivy to give this a quick check before going in.

AlanRynne (commented):
Added my thoughts on a suspicious line of code. 🙂

connorivy:
Hey @ks-cph,

Thank you very much for the PR! Also, I apologize that it took so long to get around to. Everything looks really good except for that line that Alan previously pointed out and the following few lines. It wasn't a huge issue by any means, but it seems that the units were hard coded to be meters. 

Perhaps you were just trying to save me from using the imperial system of units, and for that I thank you, but sometimes I am intent on making my life more difficult by using those units. So I made another PR (https://github.com/specklesystems/speckle-sharp/pull/2025) that I started from your code where I just send using the default units of the speckle commit whatever those may be. 

Also, I noticed that the length prop was not used by the majority of the elements that I happened to be sending at the time when I was testing this, so I ended up just adding the length prop dynamically only when applicable instead of adding it to the materialQuantities object.

Could you take a look at this pr and see if it still works for what you are trying to do? If not then please let me know what doesn't work for you and we can talk about a solution.

PR #1979, opened 2022-12-06
Tekla structures/reynold/tekla 2022

Reynold-Chan:


connorivy:
hmm I'm not sure what happened with this PR but its showing way too many "changes". We should figure that out before moving forward

Core/Core/Kits/Applications.cs
```diff
@@ -133,4 +133,94 @@ public static HostApplication GetHostAppFromString(string appname)
     }
 
   }
+
+  [Obsolete("Use HostApplications instead.", true)]
```
teocomi, line 137: Why have these obsolete methods been re-introduced?

ConnectorTeklaStructures/ConnectorTeklaStructures2022/packages.config
```diff
@@ -0,0 +1,51 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
```
teocomi, line 2: I don't think `packages.config` are used since we have SDK style projects and they can be removed

ConnectorTeklaStructures/ConnectorTeklaStructures2022/app.config
```diff
@@ -0,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
```
teocomi, line 2: What is this file? Is it needed?

ConnectorTeklaStructures/ConnectorTeklaStructures2022/ConnectorTeklaStructures2022.csproj
```diff
@@ -0,0 +1,129 @@
...
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="Sentry">
+      <Version>3.20.1</Version>
+    </PackageReference>
+    <PackageReference Include="Speckle.Newtonsoft.Json">
+      <Version>12.0.3.1</Version>
+    </PackageReference>
+    <PackageReference Include="Stylet">
```
teocomi, line 105: We should not include this NuGet, it's no longer being used.

ConnectorTeklaStructures/ConnectorTeklaStructures2022/ConnectorTeklaStructures2022.csproj
```diff
@@ -0,0 +1,129 @@
...
+      <Version>1.3.6</Version>
+    </PackageReference>
+    <PackageReference Include="TeklaOpenAPI">
```
teocomi, line 108: As mentioned previously, this NuGet is legacy: https://www.nuget.org/packages/TeklaOpenAPI/
Please see if it's possible to upgrade to the new ones: https://www.nuget.org/packages/Tekla.Structures/

ConnectorTeklaStructures/ConnectorTeklaStructures2022/ConnectorTeklaStructures2022.csproj
```diff
@@ -0,0 +1,129 @@
...
+    </ProjectReference>
+    <ProjectReference Include="..\..\DesktopUI2\DesktopUI2\DesktopUI2.csproj">
+      <Project>{3d522de4-1f73-486f-b07d-6ad794f784fd}</Project>
+      <Name>DesktopUI2</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Objects\Converters\StructuralUtilities\PolygonMesher\PolygonMesher.csproj">
+      <Project>{60b78c35-560a-4111-82a2-d03a68a047c0}</Project>
+      <Name>PolygonMesher</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Objects\Objects\Objects.csproj">
```
teocomi, line 79: The connector project should not reference Kits

teocomi (changes requested):

ConnectorTeklaStructures/ConnectorTeklaStructures2022/packages.config
AlanRynne: I think this one is in use bc the 2022 project may have been generated from a template.

It is easy to upgrade to `PackageReferences` so I'd do that instead.

teocomi: Yeah, it's already using PackageReference ;)

PR #2000, opened 2022-12-12
Fix dependency conflicts after update

teocomi:
- removes direct depencencies of speckle newtonsoft and sentry from structural connectors
- adds system.forms and windowbase to gh6
- updates hardcoded version of sqlitepclraw.lib.e_sqlite3 in rhino7


AlanRynne (approved):
Tested Grasshopper6 and 7 and everything seems to work :)

Can't test CSI but the CI builds

PR #2018, opened 2022-12-20
fix(acadcivil): better native application ids for update mode

### LINKED ISSUE #2017: bug(autocad): application ids are not unique between documents

Autocadcivil3d connector

## Describe the bug

Currently, updating behavior on receive doesn't work as expected because Autocad/Civil3d generates persistent native ids from the same seed value in every document. This means that duplicate persistent ids are almost guaranteed between two different files, resulting in recieved geometry in update mode being incorrectly updated or modified.

## To Reproduce

1. create some geometry in file 1
2. create some geometry in file 2 and send
3. receive file 2 geometry in file one on update mode

## Additional context

This issue was mentioned in https://github.com/specklesystems/speckle-sharp/pull/1912

## Proposed Solution (if any)

Create a custom unique id with the document id hash as a prefix, and store this on the object on send.

clairekuang:
## Description & motivation

Fixes an issue in the Autocad/Civil3D connector that results in buggy receiving on update mode when receiving into a new document. This was caused by non-unique native ids, and is fixed by attaching a custom application id `{documentId}-{handle}` to the xdata of every object on send.

Note: this is kinda hacky and works better in documents that have been saved before sending or receiving. This should be noted in our connector docs.

Fixes #2017 

## Changes:

Acadcivil connector

## Validation of changes:

Sent and recieved changed objects in a new document

## Checklist:


ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
```diff
@@ -7,6 +7,7 @@
 using DesktopUI2.Models.Filters;
 using DesktopUI2.Models.Settings;
 using DesktopUI2.ViewModels;
+using Newtonsoft.Json;
```
teocomi, line 10: We should not use `Newtonsoft.Json` as it might conflict, should instead use our fork `Speckle.Newtonsoft.Json`

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
```diff
@@ -926,111 +936,130 @@ public override async Task<string> SendStream(StreamState state, ProgressViewMod
...
+
+          foreach (var autocadObjectHandle in state.SelectedObjectIds)
           {
-            progress.Report.LogOperationError(new Exception($"Failed to find doc object ${autocadObjectHandle}."));
-            continue;
-          }
+            // handle user cancellation
+            if (progress.CancellationTokenSource.Token.IsCancellationRequested)
+            {
+              tr.Commit();
```
teocomi, line 971: Nor sure if possible, but maybe we should roll back the transaction if the op is canceled?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
```diff
@@ -926,111 +936,130 @@ public override async Task<string> SendStream(StreamState state, ProgressViewMod
...
+                progress.Report.Log(reportObj);
+                continue;
+              }
+
+              /* TODO: adding the extension dictionary / xdata per object 
+              foreach (var key in obj.ExtensionDictionary)
+                converted[key] = obj.ExtensionDictionary.GetUserString(key);
+              */
 
 #if CIVIL2021 || CIVIL2022 || CIVIL2023
```
teocomi, line 1018: could simplify with `#if !CIVIL2020`

teocomi (commented):
Just added some minor comments, did not test it, though!

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
clairekuang: Ah good catch, swapped this out 👍

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
clairekuang: fixed!

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
clairekuang: Unforch this also doesn't apply to autocad as well as civil2020 😅

teocomi: Ah forgot about that ;)

teocomi (approved):

PR #2020, opened 2022-12-20
feat: http proxy support

teocomi:
Adds support for HTTP Proxy to all the HTTP clients in Core.
The new `Helpers.GetHttpProxyClient` simply checks if there is any `WebRequest.GetSystemWebProxy()` and uses that together with the default credentials.

I'm unaware if this can have any negative repercussions, but it's working for me and a forum user who is behind a proxy.

This PR also improves how we check for internet connectivity, since `Ping`s do not work behind proxies, and centralizes where the `ReleasesUrl` is stored.

Info: https://blog.codeinside.eu/2022/03/28/how-to-use-ie-proxy-settings-with-httpclient/ 
Context: https://speckle.community/t/could-not-ping-google/4432/6


gjedlicska:
I'd group the http related helpers into a new `Core/Helpers/Http.cs`
But that is just nitpicking. Othersie looks good

teocomi:
Thanks @gjedlicska done so!

gjedlicska (approved):

Issue #2030, opened 2022-12-21
bug(autocad): saved streams with long serialized strings are not successfully retrieved in DUI


clairekuang:
## Describe the bug

Autocad/Civil3D connector truncates long strings in the document database, so saving too many streams or streams with many collaborators will no longer be restored properly by DUI on deserialize.

**Note** This exception is usually a result of serializing avatar jpegs, which run over 3k chars.  Due to how Autocad writes to the document database, this can create performance issues with streams with many collaborators, especially those with custom avatar images.

Proposed solution is to handle overflow by creating multiple document database entries for saved streams

## To Reproduce

1. Save `GrasshopperTestStream` on latest.dev to a file
2. Switch to another document
3. Switch back to the file with the saved stream - the saved stream no longer appears in DUI's saved streams list

This results in the following error when attempting to deserialize a truncated string: `{"Unterminated string. Expected delimiter: \". Path '[0].CachedStream.collaborators[2].avatar', line 1, position 16285."}`

## Expected behavior

All saved streams will be restored correctly in DUI on document open.

PR #2033, opened 2022-12-21
fix(acad): removes collaborators form the stream state cache to limit…

teocomi:
Quick attempt at fixing #2030 in a way that does not require editing the Core classes.


clairekuang (approved):
Tested, doesn't completely resolve the linked issue but functionally fixes retrieving saved streams 👍

PR #2042, opened 2022-12-24
fix(autocad): resolved the serialization of streams with long serialized strings

### LINKED ISSUE #2030: bug(autocad): saved streams with long serialized strings are not successfully retrieved in DUI

phliberato:
## Description & motivation
- Fixes #2030 - Long serialized strings are working in streamstate's write and read functions 

## Changes:
- Mechanism for saving json string has been modified. Now the json string is being base64 encoded and then split into 1024 characters then inserted into many TypedValue inside ResultBuffer.
- New TransactionContext class to handle Transaction and DocumentLock with a disposable interface.

## Validation of changes:
I tested the ReadState function with a 300k string.

## References

@clairekuang  I think you might consider rolling back fix #2033


phliberato:
@clairekuang Another thing I forgot to mention, it's not a good strategy to put json serialization or deserialization in the try catch scope. I did that once and discovered the deserialization issue months later.

clairekuang (approved):

PR #2048, opened 2023-01-03
fix(csi): frames appear in wrong location & deleted frames not tracked

connorivy:
## Changes:

- refresh the "Beam Object Connectivity" table as a workaround for a bug on CSI's end where frame elements that are moved via api calls will be moved in the database yet still appear in the previous location in the modelling space
- delete frame elements that were deleted from the stream to mimic the behavior of other connectors (namely Revit)
- cleanup the streamState saving process. Don't call it via the connector because it will be called later from DUI
- because the changes to the user's model cannot be undone (thanks CSI) save a backup in the Speckle folder before making any changes

## Validation of changes:

Sending and receiving the same elements in different locations several times. Now experiencing a fairly smooth user experience)

## Checklist:


AlanRynne (approved):
Trusting you on this one @connorivy as I no longer have ETABS installed

PR #2049, opened 2023-01-04
fix(revit): incorrect ceiling locations and don't allow level as host

connorivy:
## Changes:

- return true from ShouldConvertHostedElement if the host element is a level because the level will definitely not be responsible for converting the element
- don't exclude height offset from level prop and don't store seperately for ceilings. This change is in line with existing behavior of floors and roofs
- depricate offset property of ceiling because it was only ever being set, not used.

## Validation of changes:

manually sending and receiving led to correct location for ceiling objects

## Checklist:

## References

https://speckle.community/t/isses-with-attached-walls-in-revit/4171


AlanRynne (approved):
✅ Beautiful 🚀 Schema deprecation looks correct too 👍🏻

PR #2052, opened 2023-01-05
CI: Adds use of contexts and pulling through ssh

gjedlicska:
## Description & motivation

Due to the recent CircleCI debacle, we've been forced to rotate all credentials. Since we were at it, we converted all repeated env vars into contexts and applied those on the ci jobs.

This is a better strategy than the one we were following previously.

We're also now fetching the `speckle-sharp-ci-tools` repo via `ssh` instead of via a github_token.


AlanRynne (approved):
CI is running smoothly up to the build steps. Approving so we can test a full `rc` release on the release branch.

Issue #1905, opened 2022-11-23
bug(acad): Groups/Components from Sketchup comes transparent/invisible


bimgeek:
## Description

Groups/Components sent from Sketchup can be received in Autocad, but with a catch: They are transparent/invisible. Take a look at the video recording below 👇.

## Expected Behaviour

Viewing Faces/Lines of the received mesh block.

## Steps to reproduce

1. Open Autocad and enable Speckle connector.
2. Receive below stream.

## Streams/Commits Used

Below branch contains bunch of geometry from Sketchup. One of the cubes is a component and the other one is a group. 

[https://latest.speckle.dev/streams/3ed8357f29/branches/sketchup2autocad](https://latest.speckle.dev/streams/3ed8357f29/branches/sketchup2autocad)

## Screenshots

Speckle Viewer👇

![image](https://user-images.githubusercontent.com/51519350/203545047-f053582e-d3cb-4007-97dc-07b6d835d38f.png)

Autocad 👇

![image](https://user-images.githubusercontent.com/51519350/203545092-838307f2-fb54-4577-bd0e-85a161ecc899.png)

2 cubes are missing as you can see. But they can be selected. So their transparency is at 100%.

![acad_rsnMqSqs8S](https://user-images.githubusercontent.com/51519350/203545110-2f65f8dd-5fce-4017-856f-0c8d96efd0b3.gif)

## Screen Recording

https://user-images.githubusercontent.com/51519350/203545738-8c6a6728-de2c-4340-beee-2c941fd881c0.mp4

clairekuang:
This may have been resolved by #2357 , verifying now

clairekuang:
![image](https://user-images.githubusercontent.com/16748799/228791676-425b45e2-bb25-4a0b-a367-9ee52351efbd.png)
Yup fixed, thanks @JR-Morgan !

PR #2054, opened 2023-01-06
fix(rhino): updating blocks and views

clairekuang:
## Description & motivation

Blocks and Views should now be correctly received in update mode, with correct reporting as well.
This also includes an improvement to the block to native methods to handle dynamically detached block properties (eg blocks sent from sketchup)

This fix may be relevant for #1905

## Changes:
 Rhino connector and converter


AlanRynne (approved):
Nice work! Merging in so I can make an `rc` release

PR #2055, opened 2023-01-09
fix(ci): Prevents duplicate `get-ci-tools` job on deploy

AlanRynne:
Quick fix on the config generator script that was assuming `get-ci-tools` item was a string, but since last Friday it is now an object (as it now needs to have a `context` attached to it to work)


AlanRynne:
Not waiting for CI to finish as the fix was in the CI setup itself

PR #2062, opened 2023-01-11
Feature: Schema info for Pointcloud

haitheredavid:
## Description & motivation

Clouds are neat and it would be super neat to be able to send them to a stream in gh. 

## Changes:

- Added a `SchemaInfo` attribute


Objects/Objects/Geometry/Pointcloud.cs
```diff
@@ -27,9 +27,18 @@ public class Pointcloud : Base, IHasBoundingBox, ITransformable<Pointcloud>
     public string units { get; set; }
 
     public Pointcloud()
+    
+    /// <param name="points">Flat list of x,y,z coordinates</param>
+    /// <param name="colors">Optional list of colors</param>
+    /// <param name="sizes">Optional list of sizes</param>
+    [SchemaInfo(nameof(Pointcloud), "Create a point cloud object", "Objects", "Geometry")]
+    public Pointcloud(List<double> points, List<int> colors = null, List<double> sizes = null)
```
teocomi, line 35: Seems like `colors` is not being used?

teocomi:
Hey David, isn't it already possible to use the native GH/Rhino point cloud objects instead?

haitheredavid:
@teocomi rhino has this feature supported but there is no native grasshopper component that builds point clouds.

Objects/Objects/Geometry/Pointcloud.cs
haitheredavid: ah good catch, didn't see those lines missing. just pushed the fix

PR #2066, opened 2023-01-13
Object Traversal Refactor MVP (Revit & Rhino)

JR-Morgan:
Traversal Refractor mvp implementation in Rhino and Revit.

+ Added `GraphTraversal` and `TraversalRule` builder to core
+ Added `DefaultTraversal` helper functions to core (this isn't set in stone, and likely will be moved elsewhere)
+ Implemented new GraphTraversal in Rhino
+ Implemented new GraphTraversal in Revit (matching existing behaviour of traversing until convertible, then let converter to the rest)

Things to do before merge:
 - [x] Remove any `FlattenCommitObject_old` methods
 - [x] Test Revit connector with non-revit commits
 - [ ] Discuss fallback value handling in Revit. should we do it the same way as in Rhino?


clairekuang (commented):
When testing receiving the rhino test stream, noticed some objects showing up in the report that shouldn't be - `displayStyle` and `Plane`. These are marked as failed, will look into it!

connorivy (approved):

PR #2068, opened 2023-01-16
ci: verbose output when pushing nugets

teocomi:
To be merged in **main and 2.12**:

- adds verbose output when pushing nugets, so ,see we can understand what's triggering that ghost error in the CI


AlanRynne (approved):
🪄Approved! Seems _harmless enough_ (flw)

PR #2093, opened 2023-01-24
Deserialiser will no longer silently fail when root commit object was…

JR-Morgan:
Previously, if the root commit object was not a Base object (e.g. a dictionary)
The BaseObjectDeserializer.Deserialize would simply return null due to a `return ret as Base` checked cast.

Instead, I have added a throw `ArgumentException` when the deserialized object is not a Base object

Question 1:
Is ArgumentException the right exception? 
 - the argument given was not a base JSON
 - but if a bug exists in `DesierlaizeTransportObject` then this exception type could be misleading.
 (most of the exceptions in the deserializer are simply typed as `Exception`)


gjedlicska (approved):

AlanRynne (approved):
🙌🏼

PR #2097, opened 2023-01-25
(AdvanceSteel): Connector first code

phliberato:
## Description & motivation

First release of Advance Steel Connector

The main goal of this release is to integrate Autocad and Civil connector with Advance Steel API to run on host.
Create a POC to validate the Triangle library and use Delaunay's triangulation algorithm to create the beam faces. Test the Mesh Speckle object.

## Changes:

- Connector and Converter VS projects created for Advance Steel 2023. ADVANCESTEEL2023 directive has been inserted into all files that need to deal Advance Steel API. A packaged for nuget was also created.

- A TransactionContext class was modified to use Advance Steel transaction instead of ACAD Transaction.

- Advance Steel plugin is not loaded at the same moment as the ACAD plugin. IExtensionApplication class is handle the AS initialization.

- FIX: In ConverterAutocadCivil when ApplicationObject is instantiating, obj.ToString() changed to obj.Handle.ToString() to fix progressReport object update.

- AdvanceSteelBeam class was created in BuiltElements.

- Added descriptor parameter in the Update function of the Extras class to update the Advance Steel object type.

- ToString function in Utils class has been modified, o.ToString() -> o.Handle.ToString().The Handle is a persisitent indentifier which is unique per drawing. The ObjectId is a non - persitent identifier(reassigned each time the drawing is opened) which is unique per session.

## Checklist:


ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -58,8 +62,10 @@ public static class Utils
     /// <returns></returns>
     /// <remarks>
     /// This is used because for some unfathomable reason, ObjectId.ToString() returns "(id)" instead of "id".
+    /// The Handle is a persisitent indentifier which is unique per drawing.
+    /// The ObjectId is a non - persitent identifier(reassigned each time the drawing is opened) which is unique per session.
```
teocomi, line 66: @clairekuang FYI

PR #2100, opened 2023-01-26
rework the grahpql client to reduce usage of the SpeckleException

gjedlicska:
This PR reworks the `Client` class, that we use to execute graphql operations.

The goal of the PR is to reduce the excessive usage of `SpeckleExceotions` in the client operations.
The way that this is achieved:

1. new GraphQL error specific exceptions are implemented to convey useful error data about the failed graphql request.
1. a common generic method is implemented in the client, that handles the execution of all graphql request:
a. this method translates the graphql error code statuses into the new specific exception types
b. the execution method adds a retry mechanism, to support retrying failed requests. Note, this is only retrying the graphql specific errors, anything related to http exceptions will be handled separately on the http level 
c. the errors are logged by the client common execution method 

**TODOs**
- [x] Improve logged error data with request and response information
- [x] add tests for request retrying and not retried
- [x] refactor websocket client with a similar mindset


teocomi (commented):
Thanks for tidying up these classes :D
No comments except for the fact that some CI builds are failing, maybe something needs updating in DUI/Connectors? 
You can use `All.sln` to ensure everything is building fine.

gjedlicska:
Yeah, i've removed the definition of the deprecated `UserGet` query. Gonna have to fix those issues :)

AlanRynne:
Pushing this to `2.13` after a quick chat with @gjedlicska 🙌🏼 still some work to be done, and lots of reviewing on our side!

PR #2108, opened 2023-01-31
feat(gh): Improves Rhino.Compute support with AccountToken nodes and ActiveDoc changes

AlanRynne:
Adds 2 new nodes to the `DEV` section of Grasshopper `Speckle 2`.

Screenshot 2023-01-31 at 11 21 24

- **Stream Get with Token**: Uses a token instead of an account as input to authenticate a user that has not been setup in Manager.
- **Get Account Token**: Gets a user's account token stored in manager. This can be used to run a GH script locally that would otherwise require pasting a token in the document (very unsafe).

This would enable a Grasshopper user to design workflows that run in Rhino.Compute and allow other users to make commits on their behalf, instead of having to have a fixed user or setup the correct user on the server.

Screenshot 2023-01-31 at 11 32 35

Screenshot 2023-01-31 at 11 34 28


CapnK:
@AlanRynne - This would support auth when a GH script using Speckle is submitted from a users desktop machine (i.e. Get Account Token is running on a PC with the Speckle Account Manager)

Another scenario is when a GH script (using Speckle) is being invoked from a webserver.

We've also been looking at this and will be in touch with some ideas

AlanRynne:
The other scenario is also covered, `Stream get with token` does not use the local accounts to authenticate, it will use the provided token to fetch the account details directly from the server and create a short-lived account object. but the details of how to obtain said token and pass it as part of the GH definition inputs is responsibility of each developer.

How to securely obtain, store and pass along that token to the GH definition is not something we can have control over from the GH side.

The `get account token` node exists for the sole purpose of testing these Gh definitions locally using hops.

Hope this brought some clarity! And if there's any edge cases you'd like us to cover don't hesitate to ping us on this!

PR #2115, opened 2023-02-03
fix(revit-csi): various fixes for revit <-> csi workflow

connorivy:
## Checklist:


Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -711,8 +711,6 @@ public bool CanConvertToNative(Base @object)
         //Structural
         STR.Geometry.Element1D _ => true,
         STR.Geometry.Element2D _ => true,
-        STR.Geometry.Node _ => true,
-        STR.Analysis.Model _ => true,
```
teocomi, line 715: Curious why these were removed?

teocomi (approved):
Looks good to me, let us know if you'd like us to perform any specific testing :)

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
connorivy: For the node, there is no way to create those standalone in Revit. They are automatically created when you create 1D or 2D elements. Our current [NodeToNative method](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial%20Classes/ConvertAnalyticalNode.cs#L20) just returns an appObject without making any changes to the Revit document so I'm not sure why that conversion exists.

For the model, I also recently removed that from CSI. Our current [modelToNative method](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial%20Classes/ConvertStructuralModel.cs) uses its own hardcoded traversal logic to loop through the properties of the analyticalModel and convert them. To me this seems bad because it takes away from the flexibility of our current traversal logic provides. When we get rid of this object from the CanConvert list, our default traversal logic will flatten it to find the elements that are convertable. So even if the structure of the AnalyticalModel changes (which I would like it to), the conversion to Revit will not break. This method was also filled with try catches that would silence errors that we probably wanted to know about and this was also resulting in the Model object showing up as a single line in the report as opposed to showing the results of each beam, column, floor, etc.

teocomi: Got it; sounds good then :) thanks!

PR #2126, opened 2023-02-09
fix(objects): Added fix for sketchup blocks using dynamic blockDefinition prop

AlanRynne:
SketchUp sends every detached property with the prefixed `@`.

This causes issues with the `transformedGeometry` computed property of blocks, which uses `BlockInstance.blockDefinition` and `BlockDefinition.geometry` properties.

Both of these properties, on sketchup objects, would end up being deserialised as dynamic props, which would throw null reference errors.

This PR addresses this issue, while also adding warnings on the code to inform us in the future about this.


JR-Morgan (commented):
Looks like the SKP (dynamic) -> .NET problem spreads further.
`GetTransformedGeometry` throws for skp blocks.

Perhaps it would be cleaner to use the `[OnDeserialize]` callback to set the instance props

AlanRynne:
@JR-Morgan I'm re-requesting your review, but mostly to open the conversation of how cringy this is codewise... and that maybe we should not ever merge this 😅

teocomi:
Closing as we decided to fix this on the SketchUp side.

Issue #2149, opened 2023-02-14
Connector error reporting: RHINO 


gjedlicska:
Implement connector side error reporting in Rhino

This includes the first trial of DUI2 so this effort is going to be longer than subsequent ones.

PR #2151, opened 2023-02-15
log(dui2): Replacing SpeckleException usages with new logger 

JR-Morgan:
This PR applies the new logging to DUI2

Knowing that the fate of DUI2 is uncertain, we have not gone too crazy with implementing perfect logging.
Simply replacing existing logging/SpeckleException with the new
and adding a few other error logging.

see #2149


Core/Core/Logging/Analytics.cs
```diff
@@ -191,9 +192,11 @@ private static void TrackEvent(string hashedEmail, string hashedServer, Events e
           query.Headers.ContentType = new MediaTypeHeaderValue("application/json");
           client.PostAsync(MixpanelServer + "/track?ip=1", query);
         }
-        catch (Exception e)
+        catch (Exception ex)
         {
-          // POKEMON: Gotta catch 'em all!
+          Log.ForContext("eventName", eventName.ToString())
```
teocomi, line 197: This could potentially lead to a lot of noise if that endpoint is blocked...

teocomi (approved):

JR-Morgan:
Closed via #2156

PR #2156, opened 2023-02-16
log(rhino): connector logging refactor/scoping

JR-Morgan:


DesktopUI2/DesktopUI2/ViewModels/AccountViewModel.cs
```diff
@@ -113,7 +113,8 @@ public void DownloadImage(string url)
       }
       catch (Exception ex)
       {
-
+        Serilog.Log.ForContext("imageUrl", url)
+          .Debug(ex, "Swallowing exception {exceptionMessage}",ex.Message);
```
gjedlicska, line 117: All the Debug events are retained for a short period of time (3 days) for now.

I'd make all of these "swallows" a warning level for now, so that we can better diagnose stuff.

DesktopUI2/DesktopUI2/ViewModels/HomeViewModel.cs
```diff
@@ -290,6 +290,7 @@ internal void UpdateSelectedStream()
       }
       catch (Exception ex)
       {
+        //FIXME: This branch can't ever get hit right?
         Log.Error(ex, ex.Message);
```
gjedlicska, line 294: Add a message template to this pls, otherwise its very hard to trace back from the error message...

DesktopUI2/DesktopUI2/ViewModels/OneClickViewModel.cs
```diff
@@ -107,9 +107,10 @@ public async void Send()
       {
         fileName = Bindings.GetFileName();
       }
-      catch
+      catch(Exception ex)
       {
         //todo: handle properly in each connector bindings
+        Serilog.Log.Debug(ex, "Swallowing exception {exceptionMessage}",ex.Message);
```
gjedlicska, line 113: warn this

DesktopUI2/DesktopUI2/ViewModels/OneClickViewModel.cs
```diff
@@ -90,7 +90,7 @@ public OneClickViewModel(IScreen screen)
       }
       catch (Exception ex)
       {
-        new SpeckleException("Could not initialize one click screen", ex, true, Sentry.SentryLevel.Error);
+        Serilog.Log.Fatal(ex, "Could not initialize one click screen {exceptionMessage}",ex.Message);
```
gjedlicska, line 93: its prob more consistent to use `using Serilog;`

DesktopUI2/DesktopUI2/ViewModels/StreamSelectorViewModel.cs
```diff
@@ -147,53 +146,51 @@ private async Task GetStreams()
...
-            new SpeckleException("Could not fetch streams", e, true, Sentry.SentryLevel.Error);
-          }
+          streams.AddRange(result.Select(x => new StreamAccountWrapper(x, account.Account)));
         }
-        if (StreamGetCancelTokenSource.IsCancellationRequested)
-          return;
-
+        catch (Exception ex)
+        {
+          bool isError = ex is TaskCanceledException || ex.InnerException is TaskCanceledException;
```
gjedlicska, line 179: the logic here should be reversed no? this should be an `isNotConsideredAsError` variable as we do not care about task cancellations.

gjedlicska (changes requested):
I've added some comments inline

DesktopUI2/DesktopUI2/ViewModels/StreamSelectorViewModel.cs
JR-Morgan: Good catch!

DesktopUI2/DesktopUI2/ViewModels/AccountViewModel.cs
```diff
@@ -171,7 +171,7 @@ private void DownloadComplete(object sender, DownloadDataCompletedEventArgs e)
       }
       catch (Exception ex)
       {
-        Log.Debug(ex, "Swallowing exception {exceptionMessage}",ex.Message);
+        Log.Warning(ex, "Swallowing exception {exceptionMessage}",ex.Message);
```
teocomi, line 174: Would it make sense to add more content here, like "Swallowing exception in avatar download"?
Sometimes the full trace is lost, and it's tricky to understand where an exception was thrown...

DesktopUI2/DesktopUI2/ViewModels/StreamViewModel.cs
```diff
@@ -1014,7 +1014,7 @@ public async Task DownloadImage(string url)
       }
       catch (Exception ex)
       {
-        Serilog.Log.Debug(ex, "Swallowing exception {exceptionMessage}", ex.Message);
+        Serilog.Log.Warning(ex, "Swallowing exception {exceptionMessage}", ex.Message);
```
teocomi, line 1017: There is a risk this would introduce a lot of noise being an HTTP method, no?

DesktopUI2/DesktopUI2/ViewModels/AccountViewModel.cs
gjedlicska: yeah, you are right. Hopefully with the Exception rework efforts we've made the stack traces more consistent, but its prob better to add extra context.

I propose making it like this:
```
Log.Warning(ex, "Swallowing exception in {methodName}: {exceptionMessage}", nameof(method), ex.Message);
```

DesktopUI2/DesktopUI2/ViewModels/StreamViewModel.cs
gjedlicska: Possibly yes, but I'd add this for now. Its just warning, so its not polluting the Sentry event log, but we can have a better idea, what is going wrong.

Also info and debug messages are retained for a short period of time on the `seq` server. It's a bit better to keep these for a while, while we act on the results.

DesktopUI2/DesktopUI2/ViewModels/AccountViewModel.cs
JR-Morgan: 👍 I'll make these changes

JR-Morgan: I've added the method name into all "Swallowing" exception templated logs, as per your proposal.
However, perhaps there is some value in enriching all our logs like this.

https://stackoverflow.com/questions/29470863/serilog-output-enrich-all-messages-with-methodname-from-which-log-entry-was-ca

Looks like it may be possible either through a custom ILogger extension method (with `CallerMemberNameAttribute`), or through `AutofacSerilogIntegration` to enrich the LoggerConfiguration

gjedlicska (approved):

teocomi (approved):

Issue #1273, opened 2022-05-24
Any plan for .NET6 SDK Style ? 


chuongmep:
Dear @teocomi, 
I have a question, we have any plan to update use NET6- visual studio 2022 with target framework .net48 for Revit or not ? 
I want colab something to improve structural Revit project and some process use CI/CD. Is your team willing to let me improve this and will it affect your team's workflow? 
This is a sample project developed: https://github.com/chuongmep/RevitAddInManager
![image](https://user-images.githubusercontent.com/31106432/170010320-dfdc7927-4038-4ecc-80af-ccce5d29d429.png)

chuongmep:
> 
It still is net 4.6 or 4.8 but in file config is net6 style, it shorter and easier for expand product in future.

chuongmep:
@teocomi see example in my branch : https://github.com/chuongmep/speckle-sharp/blob/d1e9796e9e86ac4d2fabfa9ec200519f2e495cb9/ConnectorRevit/ConnectorRevit/ConnectorRevit.csproj#L4-L49

teocomi:
Ah, thanks! This is actually how we had the multiple projects set up in the past, with multiple build configs, but we decided to move and use the shared projects instead as it gives us a bit more control and it's easier to work within our CI/CD setup.

chuongmep:
Thanks you @teocomi .

PR #2172, opened 2023-02-20
[WIP] Unify to .net6

martijn00:
The current state of the repo and code is hard to maintain and build. By moving to .net6 it is lots easier and code can be improved.

This is a starting point and if this is the correct approach I can pick up the work to complete this.

## Changes:

Update project files, references and code to work with .net6

## To-do before merge:

Validate that everything works.

## Checklist:

### DIFF
--- a/ConnectorArchicad/ConnectorArchicad/ConnectorArchicad.csproj
+++ b/ConnectorArchicad/ConnectorArchicad/ConnectorArchicad.csproj
@@ -3,8 +3,6 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
-    <Company>Speckle</Company>
     <Product>ConnectorArchicad</Product>
     <Version>2.0.0-dev</Version>
   </PropertyGroup>
--- a/ConnectorArchicad/ConnectorArchicad/Elements/Object.cs
+++ b/ConnectorArchicad/ConnectorArchicad/Elements/Object.cs
@@ -8,7 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using Speckle.Newtonsoft.Json;
+using Newtonsoft.Json;
 using Archicad.Model;
 
 
--- a/ConnectorAutocadCivil/ConnectorAutocad2021/Properties/AssemblyInfo.cs
+++ b/ConnectorAutocadCivil/ConnectorAutocad2021/Properties/AssemblyInfo.cs
@@ -1,4 +1,4 @@
-﻿using System.Reflection;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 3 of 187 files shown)


teocomi:
Hey @martijn00 thanks for this PR.
We cannot update Core to .net6 because, AFAIK, it cannot be referenced by older .NET Frameworks (basically all our connectors): https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0

We could, though, add additional `<TargetFrameworks>` if needed.

Also, please note that in your PR you are replacing our `Speckle.Newtonsoft.Json` with `Newtonsoft.Json`. That can't be done, we are purposely using a fork to avoid dll hell in various host applications.

martijn00:
@teocomi it should be possible to update most of those to .net6 too? Adding Additional `TargetFrameworks` only works when using the new project system, which is not used in the connectors.

The problem with `Speckle.Newtonsoft.Json` is that this way applications can't trim their code on that level, and need to include their own Newtonsoft version. I was planning to replace it with `System.Text.Json` anyways because that is the new standard.

Are there things that I should know that prevent this?

martijn00:
I understand that this is quite a change, and shouldn't be merged overnight or anything. Still quite a lot of work need to go into something like this, and needs testing on your side.

My motivation is that we are about to start quite a big new project that uses Speckle and if we want or need to make changes to the source, the current state of the repo makes it hard. In the long term it is necessary to move to a unified dotnet experience based on .net6 or 7. I think the timing is right because of things like desktop support in .net6. This opens up for lots of improvements to the code and speed optimalizations as well.

Maybe something that helps on your end is my experience with something like this. I'm a long time Microsoft MVP on .NET, and creator of library's like MvvmCross, Lottie, Exoplayer for .NET.

For Newtonsoft, wouldn't it be possible to use binding redirects? In the end I think it should still move to System.Text.Json. Unless there is something very special in the code I can't think of a reason not to do it in .net6 or 7. It should be compatible too. But maybe a good strategy is to cut that up into another PR?

teocomi:
Thanks for the extra context, @martijn00; how about we jump on a quick call to discuss? 

https://calendly.com/d/z2z-hnh-2y2/15-min-chat

martijn00:
Do you have time in 5min?

AhmedZero:
it'll be better .Net 7 too, there is a lot of improvement in performance.
I don't know why Autodesk doesn't upgrade its .NET Framework 4.7.2 to .NET 6 or 7?

chuongmep:
I also a same issue before, https://github.com/specklesystems/speckle-sharp/issues/1273. This is a good to maintain, I know that, but need to consider because some SDK of Autodesk still base on .Net Framework, I guess maybe need 2-3 year for Autodesk to can upgrade it to net 7 or net standard .

teocomi:
Hey @martijn00 , we're closing this for now as .net 6 is not supported in most of the host applications where our Connectors run. 
Happy to discuss more if needed - you know where to find us! :)

Issue #2164, opened 2023-02-17
bug(revit): updating analyticalMembers location may change the location of connected members


connorivy:
## Prerequisites

- [x] I read the [contribution guidelines](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)
- [x] I checked the [documentation](https://speckle.guide/) and found no answer.
- [x] I checked [existing issues](../issues?q=is%3Aissue) and found no similar issue. 
- [x] I checked the [community forum](https://speckle.community/) for related discussions and found no answer.
- [x] I'm reporting the issue to the correct repository (see also [speckle-server](https://github.com/specklesystems/speckle-server), [speckle-sharp](https://github.com/specklesystems/speckle-sharp), [specklepy](https://github.com/specklesystems/specklepy), [speckle-docs](https://github.com/specklesystems/speckle-docs), and [others](https://github.com/orgs/specklesystems/repositories))

## What package are you referring to?

Revit Connector

## Describe the bug

In revit when an analytical member is dragged, it will also drag the end of any analytical members that are connected to it. In the two screenshots below you can see that the first one is the original state of the building. The second one is after a version from ETABS was received (however the geometry was UNCHANGED). The problem is that during the receive operation, the column baselines were set vertically, and then the crossbeam (highlighted in the images) that connects the tops of the columns had its baseline set with it's endpoints flipped. This normally would not be an issue except that it dragged the tops of the columns with it even though they had already been set and were not supposed to move.

![image](https://user-images.githubusercontent.com/43247197/219792521-0d1d7dc6-96bb-4101-9ea0-8db2842abacb.png)

![image](https://user-images.githubusercontent.com/43247197/219792471-2a0a1d10-0bee-421c-8bd2-49079f0dce99.png)

## Expected behavior

The geometry should be unchanged since it was not changed in ETABS and the data that was sent to Speckle shows the same geometry as the original geometry in Revit.

## System Info

If applicable, please fill in the below details - they help a lot!

AlanRynne:
Closing!

PR #2175, opened 2023-02-20
fix(revit): various fixes to analytical elements in Revit

connorivy:
## Changes:

- simple change in conversionUtils because .toString was being called on null members and throwing errors
-  create cache for SectionProfiles so sections don't have to be created every time
- when updating, make sure not to flip the baseLine of elements (this partially fixes https://github.com/specklesystems/speckle-sharp/issues/2164 but not fully)
- don't hard cast curves to a baseLine because that will throw errors sometimes
- use reflection to get all the props of the structuralSection from Revit. There are so many structural section classes in Revit that this seemed like the cleanest solution to me
- change the ShapeType prop of Speckle StructualSections to a virual prop and then give each child prop a default value. It doesn't make sense to always have to tell an angle section that its type is an angle. It should know that

## Validation of changes:

this is a commit made before this pr https://latest.speckle.dev/streams/85bc4f61c6/commits/471e9287f4. You can inspect each element and see the property -> profile. Nothing is wrong with it but there is missing information
this is a commit made from this branch https://latest.speckle.dev/streams/85bc4f61c6/commits/9ffb056eb2. It has much more info. Some of it is wrong, but that is Revits fault.

Also, the specific case with frames becoming twisted no longer happens in this issue https://github.com/specklesystems/speckle-sharp/issues/2164

## Checklist:

## References

https://speckle.community/t/analytical-stick-conversion-from-revit-to-speckle/4951 This is where one of the issues originated

### DIFF
--- a/Objects/Objects/Structural/Property/SectionProfile.cs
+++ b/Objects/Objects/Structural/Property/SectionProfile.cs
@@ -10,7 +10,7 @@ namespace Objects.Structural.Properties.Profiles
   public class SectionProfile : Base //section profile description
   {
     public string name { get; set; }
-    public ShapeType shapeType { get; set; }
+    public virtual ShapeType shapeType { get; set; }
     public double area { get; set; }
     //Moment of inertia about the Major axis
     public double Iyy { get; set; }
@@ -42,6 +42,7 @@ namespace Objects.Structural.Properties.Profiles
     public double width { get; set; }
     public double webThickness { get; set; } // tw 
     public double flangeThickness { get; set; } // tf
+    public override ShapeType shapeType { get; set; } = ShapeType.Rectangular;
 
     public Rectangular() { }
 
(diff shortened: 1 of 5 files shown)


Objects/Objects/Structural/Property/SectionProfile.cs
```diff
@@ -10,7 +10,7 @@ namespace Objects.Structural.Properties.Profiles
   public class SectionProfile : Base //section profile description
   {
     public string name { get; set; }
-    public ShapeType shapeType { get; set; }
+    public virtual ShapeType shapeType { get; set; } = ShapeType.Undefined;
```
connorivy, line 13: Changing this to a virtual prop isn't a breaking change, is it @teocomi?

teocomi (approved):
Looks good, if this is not a hotfix it should go into `dev`

PR #2179, opened 2023-02-22
Optimised polyline to native performance

JR-Morgan:

PR #2181, opened 2023-02-22
feat: Improved performance in Progress Report

JR-Morgan:
When profiling performance in Rhino, we noticed the updating report objects in the progress report was unnecessarily costly.
So we have switched from a list to using a dictionary. 

Now the performance impact of the progress report is trivial


clairekuang (approved):

teocomi (approved):

AlanRynne (changes requested):
This PR seems to contain no changes, which is rather suspicious...

Let's double check before doing anything silly 😅

AlanRynne:
Closing as this changes seem to have been already merged here 👉🏼 https://github.com/specklesystems/speckle-sharp/pull/2179

PR #2184, opened 2023-02-23
feat(core): Added lock for `AccountManager` Auth flow

### LINKED ISSUE #2142: Use locking mechanism to prevent AccountManager form starting multiple flows simultaneaously
Currently, if multiple apps are using the AccountManager class to add an account at the same time an error will be thrown about the port being used.

We should instead be able to differentiate if the port is used by a separate process and/or a Speckle one. A solution is to use a lock file/flag (ideally in the SQL transport) so that it can be checked cross-app and we can give a meaningful error message to the user.

Eg
If the lock exists and still valid -> Complete the account add in the app you started it in / or close all other apps or something / if it crashed pls wait until the timeout.
If the lock is not valid any more -> Please close all stuck instances of revit manager rhino whatever
If no lock and port taken -> Please close the app, that is not a Speckle app, that is running on port 27... whatever

gjedlicska:
Produces this in manager if a file based lock is there

image

fixes #2142


Core/Core/Credentials/AccountManager.cs
```diff
@@ -450,16 +464,111 @@ public static async Task AddAccount(string server = "")
...
+    /// <param name="server">Server to use to add the account, if not provied the default Server will be used</param>
+    /// <returns></returns>
+    public static async Task AddAccount(string server = "")
+    {
+      Log.Debug("Starting to add account for {serverUrl}", server);
+
+      server = _ensureCorrectServerUrl(server);
+
+      // locking for 1 minute
+      var timeout = TimeSpan.FromMinutes(3);
```
teocomi, line 537: 3 or 1 minute? :)

Core/Core/Credentials/AccountManager.cs
```diff
@@ -450,16 +464,111 @@ public static async Task AddAccount(string server = "")
...
+        );
+      }
+
+      // make sure all old locks are removed
+      foreach (var id in lockIds)
+      {
+        AccountAddLockStorage.DeleteObject(id);
       }
+
+      var lockId = DateTime.Now.Add(timespan).ToString("o");
```
teocomi, line 511: We don't need to do this here or store the `lockId`, can just clear all LockIds in the `_unlockAccountAddFlow`, right?

teocomi (commented):
💅

Core/Core/Credentials/AccountManager.cs
gjedlicska: good point, when unlocking, we can just nuke the lock table.

teocomi (approved):

PR #2185, opened 2023-02-23
feat(GH): New K/V components and Speckle Group for list support

### LINKED ISSUE #2060: GH: Extend Speckle Object K/V is not working as expected.
Reported in: https://speckle.community/t/extend-keys-component-mixes-values/3786/8

## The problem

I think something is off with our data matching logic, but it also may be that what we want to do goes against all Grasshopper common-sense and it may be better to re-think this.

At the very least, we should flag any limitations as errors so the user is aware that they're trying to do something out of the scope of that component.

In essence, we need to be able to automatically differentiate if the value of a specific provided key is `an item` or `a list`. The way this is done is by bypassing GH data matching capabilities and running our own matching logic based on the path of the current item to extend `{X;Y}`.

We fetch the `key` list for each object by finding the branch in the `key` input with the same path as the object.
Given a specific path, we'll look for the specific "subTree" in the value input tree (i.e. all the branches that start with `{X;Y;...}`. 

Once that subTree is obtained, we iterate through the branches assuming each branch contains the value/s of each key.

Somewhere along the way this is failing to work as intended.

## Possible solutions

- Stop fighting grasshopper and flag this component (and the object by k/v) as **obsolete**. This is the easy solution, it requires 0 maintenance for us. Downside is that users would have to create their own python scripts to replace these components. Upside is that they'd have total control over what is a list, what is a tree and how things are mapped.

- Determining what scenarios would not work under this logic and report them as errors, preventing execution and any weird behaviours that may happen. Downside of this is that this is very annoying to the user, and would require some pretty detailed explanations of why and what to do to fix it. Upside is that no mixed data should ever be possible.

AlanRynne:
Fixes #2060

Adds 2 new versions for the `Create Speckle Object by Key/Value` and `Extend Speckle Object by Key/Value` components.

The new nodes will replace the old ones (upon user request) automatically, this ensures backwards compatibility with older scripts for a couple of versions until we remove them.

The new nodes behave exclusively as item value input (i.e. one list of keys and one list of values per object). This change means that list support is no longer directly supported, which is why `Speckle Group` was created.

**Speckle Group**

This node is purely a Grasshopper wrapper for a `List<object>` and has absolutely no effect on the data that is sent. They are only intended to be used in K/V components, but others like CSO also know how to unwrap the list to prevent weird situations from happening.


teocomi (approved):
Haven't tried upgrading the old components, but otherwise, all good!

image

Issue #2188, opened 2023-02-23
bug(navis): Nested Saved View filter selections are not returned correctly


jsdbroughton:
## What package are you referring to?

Navisworks Connector

## Describe the bug

When Saved Viewpoints is set as the selected source, and a nested view is saved, recalling that saved selection again fails due to string escaping in the DUI. Restoring the selection should be straightforwardly recreating the view reference and proceeding as with untested views.

## To Reproduce

1. Select a nested view as a selection filter resource.
2. Saved the stream card
3. Activate the stream card
4. Click Send
5. DUI reports zero objects selected.

## Expected behaviour

Objects in the Saved Viewpoint would be selected and sent.

PR #2189, opened 2023-02-24
fix(navisworks): Fixes recalling saved views from a filter selection #2188

jsdbroughton:
Actually, it fixes it by saving a different value to `filter.selection`


AlanRynne (approved):
This looks harmless enough! FLW 😅

PR #2190, opened 2023-02-24
fix:(navisworks): No longer require wgl. 

jsdbroughton:
One of a list of Avalonia implementation fixes. This cuts out the lost WGL context fatal crash.


jsdbroughton:
Polluted commits

PR #2191, opened 2023-02-24
fix:(navisworks): No longer require wgl.

jsdbroughton:
One of a list of Avalonia implementation fixes. This cuts out the lost WGL context fatal crash.


AlanRynne (approved):
I feel enlightened now that I now what WGL is, and that it can be turned off 😅

PR #2209, opened 2023-02-28
feat(core): Adds http retry policy

### LINKED ISSUE #1688: Retry dropped connections
Currently if one http request in the send or receive operation fails due to a dropped connection, the whole operation errors out.
We should gracefully retry dropped connections.

gjedlicska:
Fixes #1688 

Adds a low level (Speckle) http client handler that handles retries and error reporting out of the box for any client generated by the `GetHttpProxyClient`  

Also, this should address one of our most common errors around SSL connection establishment:
image

It happens ~ 300 - 700 times per day
image


AlanRynne:
Builds and runs just fine! Tested some sending/receiving operations in Grasshopper successfully.

I'll approve after @gjedlicska and myself to some failure tests to ensure the retry policy works as expected

AlanRynne (approved):
Tested live in our `latest` sever, and in a custom server by @gjedlicska made to fail some requests, and force a retry on our side.

Everything worked beautifully 🙌🏼

PR #2223, opened 2023-03-02
fix(gh): Schema Object no longer fails on first drop

AlanRynne:
Fixes an issue where Create Schema Object would fail to create it's input, leading to a non-usable component. Existing components were not affected by this.

Also fixes an issue reported on the forum where Schema Object nodes would not run on compute properly.


AlanRynne:
I'm blessing my own PR here 🙇🏼‍♂️ Only affects grasshopper and includes only fixes for the next alpha.

PR #2232, opened 2023-03-03
Fix(Navisworks): Complete send of saved viewpoint commits: Fixes #2231

### LINKED ISSUE #2231: bug(Navisworks): Sending Saved Viewpoints fails
## Prerequisites

- [x] I read the [contribution guidelines](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)
- [x] I checked the [documentation](https://speckle.guide/) and found no answer.
- [x] I checked [existing issues](../issues?q=is%3Aissue) and found no similar issue. 
- [x] I checked the [community forum](https://speckle.community/) for related discussions and found no answer.
- [x] I'm reporting the issue to the correct repository (see also [speckle-server](https://github.com/specklesystems/speckle-server), [speckle-sharp](https://github.com/specklesystems/speckle-sharp), [specklepy](https://github.com/specklesystems/specklepy), [speckle-docs](https://github.com/specklesystems/speckle-docs), and [others](https://github.com/orgs/specklesystems/repositories))

Navisworks Connector

## Describe the bug

Selecting Saved Viewpoints as a selection source in DUI then fails to complete a Send. Leaves the UI in an indeterminate hanging state.

## Expected behaviour

Send completes, and the commit includes the referenced scene.

## Failure Logs

image

## Additional context

This error doesn't exhibit in all versions of Navisworks.

jsdbroughton:
## Description & motivation

This fixes #2231. The error was caused by differing implementations of the SavedItem class with different versions of Navisworks.

The original code assumed that the reference object was always sent as a string where the context of the reference and the member were separated with a ':'.

The reconstruction of the selected objects was failing when it was assumed this would always be the case. Instead the `string.Split` would return a single item array instead of the two parts. 

This fix also includes an alternative of resolving the selected Saved Viewpoint by Guid. This also has different implementations across versions. Prior to Navisworks 2022, a GUID was produced when saving a Viewpoint, but it was always the default Guid i.e. `00000-00000-00000-00000`, as such rendering it useless as a lookup.

## Changes:

The changes to defensively handle the differences between versions made this a bigger fix than anticipated.

ConnectorNavisworksBindings.Filters:
- feature detection on the host application will use a Guid as the reference if it is not a default Guid. This is the ideal as Guids persist even when the ordering or naming changes.
- The fallback to the View Name is resilient to order changes but will break if the naming changes. This will be used as a reference if no good Guid is present.
- The optional fallback of the path index in the tree of saved viewpoints is rejected as, unlike the ModelItems, where this is used as a pseudoId, and the ordering is static, Viewpoints can be reorganised much more frequently

ConnectorNavisworksBindings.Objects:
- An extensive rewrite of the GetObjectsFromSavedViewpoint method, which now handles the validity of save reference before handing to ResolveSavedViewpoint, which uses feature testing rather than version testing as branching code.
- Includes more robust handling of nulls resulting from the conversions of references to viewpoints
- No changes to retrieval of objects from the saved viewpoint

ConnectorNavisworksBindings.Send:
- Now sends the commit view alongside the saved view if the advanced setting option is checked.

ConverterNavisworks.ToSpeckle:
- Handles the two ways a viewpoint is referenced when converting the Viewpoint object itself.

## To-do before merge:

- [x] Ensure PR #XX change to DUI is merged

## Validation of changes:

Manual testing with multiple versions.

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.cs
+++ b/Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Reflection;
 using Autodesk.Navisworks.Api;
--- a/ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
+++ b/ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Linq;
 using Autodesk.Navisworks.Api;
 using Autodesk.Navisworks.Api.Clash;
@@ -20,7 +20,7 @@ namespace Speckle.ConnectorNavisworks.Bindings
 
       var selectSetsRootItem = Doc.SelectionSets.RootItem;
 
-      var savedSelectionSets = selectSetsRootItem.Children.Select(GetSets)?.OfType<TreeNode>().ToList();
+      var savedSelectionSets = selectSetsRootItem.Children.Select(GetSets).ToList();
 
       if (savedSelectionSets.Count > 0)
       {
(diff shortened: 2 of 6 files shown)


jsdbroughton:
Requires #2234 to satisfy failing CI

Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.cs
```diff
@@ -47,7 +47,7 @@ public ConverterNavisworks()
     public ReceiveMode ReceiveMode { get; set; }
 
 
-    public Document Doc { get; private set; }
+    public static Document Doc { get; private set; }
```
teocomi, line 50: Does this need to be static? It can usually be just accessed in the Converter via partial classes...

teocomi (approved):

PR #2239, opened 2023-03-05
Fix(revit): scale analytical properties when sending to speckle

connorivy:
## Description & motivation

Analytical section properties are being sent to speckle using Revit internal units (feet). Scale to speckle before assigning these values.

## Changes:

- compute scale factor and then multiply required values by scale factor. Previous experiences profiling with Jedd has shown that computing the scale factor once and then multiplying it to values is much more performant than calling scaleToNative every time.

## Screenshots:

From Revit
image

Before this PR
image

After this PR
image

## Validation of changes:

See above screenshots. The types now send the correct sizes to Speckle

## Checklist:

- [x] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x] My commits are related to the pull request and do not amend unrelated code or documentation.
- [x] My code follows a similar style to existing code.
- [ ] I have added appropriate tests.
- [ ] I have updated or added relevant documentation.

## References
Issue reported on forum when receiving to ETABS https://speckle.community/t/etabs-20-3-and-revit-2022-interoperability-issue/5065/4

PR #2244, opened 2023-03-06
Fix(Revit): add appropriate scale values to analytical props

connorivy:
## Changes:

- instead of getting all section properties with reflection, we are explicitly converting them so that we can apply the proper scale to each value.

## Screenshots:

Here we have our good ol' Revit house that we're going to send to ETABS
image

Before this PR, receiving in ETABS and extruding the frames looks like this. The sections that are giant have a report error saying that the section was unable to be created, so ETABS just picked an existing one.
image

Zooming in on the circled area shows that the members are in fact extruded, they are just extremely small.
image

Doing the same process after this PR shows all of the sections have generally the correct shape. Further inspection of the created sections will show that the values are completly correct.
image

This is also the case for SAP
image

## Validation of changes:

See above screenshots. Inspection of each of the created frames in ETABS show that the values are set accurately based off of what the values were in Revit.

## Checklist:

## References

Issue reported on forum when receiving to ETABS https://speckle.community/t/etabs-20-3-and-revit-2022-interoperability-issue/5065/4

Previous PR https://github.com/specklesystems/speckle-sharp/pull/2239

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertAnalyticalStick.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertAnalyticalStick.cs
@@ -422,37 +422,7 @@ namespace Objects.Converter.Revit
           break;
       }
 
-      var propNamesMap = new Dictionary<string, string>()
-      {
-        { "Height", "depth" },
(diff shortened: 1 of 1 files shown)


AlanRynne (approved):
Nice! Merging it in so we can release the alpha

PR #2250, opened 2023-03-07
fix(core): ignore abstract types in `Base.speckle_type` serialization

### LINKED ISSUE #2249: core: ignore abstract class types in `base.speckle_type` serialization
With the new `Instance` abstract class in Objects, any implementations of that class has the full type serialized in `speckle_type` which is v ugly and presents issues for non-c# connectors that need to type match on deserialize.

abstract types should be ignored during serialization of `Base.speckle_type`

clairekuang:
## Description & motivation
Removes abstract types from speckle type serialization in `Base`.
Closes #2249

## Changes:

Core/Models/Base

## Screenshots:
Before:
image

After:
image

## Validation of changes:

Sent and received block instances

### DIFF
--- a/Core/Core/Models/Base.cs
+++ b/Core/Core/Models/Base.cs
@@ -283,7 +283,10 @@ namespace Speckle.Core.Models
 
           while (myType.Name != nameof(Base))
           {
-            bases.Add(myType.FullName);
+            if (!myType.IsAbstract)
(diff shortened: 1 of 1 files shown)


AlanRynne (approved):
Wow that was quick 👏🏼 Thx @clairekuang

PR #2256, opened 2023-03-07
feat: forces the use of english names when grouping objects on send

teocomi:
This PR aims to solve some of the issues reported in [this thread](https://speckle.community/t/mapping-tool-for-cad-bim-workflows/4086/24) in regard to the localization of categories used for grouping objects.

The idea is to derive an English name from the OST_BuiltInCategoryName name, note that there is not always an exact match here. For example, `Levels` in the UI corresponds to `OST_Levels`, but `Air Terminal Tags` to `OST_DuctTerminalTags`.

**NOTE: this PR will be breaking change for any code/app relying on a specific language for the category names.**

The category property on each element is kept as it is. 

Sample output, where `Soffitti` has been renamed to `Ceilings`:

image

### DIFF
--- a/ConnectorRevit/ConnectorRevit/ConnectorRevitUtils.cs
+++ b/ConnectorRevit/ConnectorRevit/ConnectorRevitUtils.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Autodesk.Revit.DB;
 using Speckle.Core.Kits;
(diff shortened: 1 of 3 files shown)


teocomi:
@martinromby would this affect you in some ways?

martinromby:
> @martinromby would this affect you in some ways?

Thanks for bringing our attention to this @teocomi 

Even though that we have a hardcoded whitelist of categories, I believe that we’re good on this one. I actually believe that this implementation will be beneficial to us when we start to onboard people with localised versions of Revit. 

If you do see some changes to the category names in the image below - please let us know 🙏

6744F30B-7ED5-49FD-99BD-0E40F616F29E

Issue #2187, opened 2023-02-23
feat(objects/rhino/revit): implement collections for layers and categories


clairekuang:
## What package are you referring to?

- Objects kit
- Rhino Connector
- Revit Connector

See notion scoping doc for more details: https://www.notion.so/speckle/Model-Collections-c8cd6ef6a541499ea88e0bae0bd96000

## Is your feature request related to a problem? Please describe.

Currently, commits from Rhino are structured with a flat list of `Base` representing each layer by it's full path. This is difficult to traverse in the viewer and also on receive.

Commits from Revit are structured with a list of `Base` representing categories by name.

## Describe the solution you'd like

- [ ] Refactor `Container` class in Objects

Implement collections for:

- [ ] Rhino send
- [ ] Rhino receive
- [ ] Revit send
- [ ] Revit receive

Also this should be implemented in other repos for:

- [ ] Blender
- [ ] QGIS

## Additional context

This needs to consider backwards compatibility on receive, since old commits will not have `Collection` objects.

PR #2259, opened 2023-03-08
feat(rhino/objects): sends and receives layers as `Collection`

clairekuang:
## Description & motivation

Sends and receives Rhino layers as `Collection` objects. Includes the following changes:

- Minor refactor of `Container` class to `Collection`
- Sends commits as nested `Collection` instead of a flat list of `Base` representing layers
- Sends the commit `Base` as a `Collection` representing the model
- Adds new feature: send just layers from Rhino using the ProjectInfo filter
- Receives `Collection` as layers
Part of #2187

## Changes:

- Objects
- Rhino Converter and Connector

## To-do before merge:

- [ ] Get feedback on layer updating behavior
- [ ] test block layer handling
- [ ] test receiving layers on create mode

## Validation of changes:

Tested sending and receiving new commits with collections, and old commits for backwards compatibility


AlanRynne:
Thx for the conflict fix @clairekuang 🙌🏼

PR #2260, opened 2023-03-08
chore: Change `Material.Icons.Avalonia` for our own `Speckle.Material.Icons.Avalonia` package

teocomi:
Updates some dependencies and introduces our fork of Material.Icons.Avalonia to fully remove a dependency on Newtonsoft.Json.


AlanRynne (approved):
All looks good!

I'll accept the SchemaBuilderGen.cs changes since they're needed but would be nice to have them in a separate PR for better tracking next time

PR #2264, opened 2023-03-09
feat(dui2): Improved receive exception handling

JR-Morgan:
This PR is a prelude towards better error reporting in our connectors.

Its goals are to:
- Ensure all connector errors are meaningfully logged
- Improve the user facing error messaging in DUI2 (no more "Something went wrong...")
- Clarify the way connector bindings should handle exceptions
- Remove exception swallowing from a few places in core (kit loading)

It also fixes a few bugs:
 - https://github.com/specklesystems/speckle-sharp/issues/2265
 - Bug with stream search spamming cancellation messages reported by @bimgeek 
 - User cancel, leading to meaningless errors that get logged.
 - Several exceptions being swallowed

TODO:
- [X] Add better exception handling to DUI2 bindings.
- [X] Reduce duplicate code snippets between connector bindings by extracting functions to a common place.
- [x] Update connector bindings.
    - [X] Rhino
    - [X] Revit
    - [x] AutoCad
    - [x] Tekla
    - [x] Navisworks
    - [ ] Archicad
    - [ ] CSI
    - [ ] Bentley

### New Guidelines for ConnectorBindings (receive + send + preview) functions
- DON'T return `null`, instead throw an exception! (ideally, one that DUI2 has a case for prettifying)
- DON'T cancel the cancellation token (that's for the user to do)
- DO call `cancellationToken.ThrowIfCancellationRequested();` between async operations, IF IT'S SAFE TO DO SO! (i.e. won't lead to a partial receive job, or invalid host app state)
- DO wrap exceptions to give a more specific/user readable exception message, if it makes sense to. However, IMPORTANT: For Async/Operations inside `try catch` blocks, DON'T swallow or wrap `OperationCancelationException`. You MUST let this bubble up higher without wrapping.
- Do use the functions in `ConnectorHelpers` when possible to ensure consistency between connectors.
- Avoid empty catch statements, at the very least do a `Log.Warning(ex, "Swallowing exception in {methodName} {exceptionMessage}", ... , ... )`


ConnectorTeklaStructures/ConnectorTeklaStructuresShared/UI/ConnectorBindingsTeklaStructures.cs
```diff
@@ -21,7 +21,7 @@ public void OpenTeklaStructures()
         UpdateSavedStreams(streams);
     }
     public static Model Model { get; set; }
-    public List<Exception> Exceptions { get; set; } = new List<Exception>();
+    public List<Exception> Exceptions { get; set; } = new List<Exception>(); //What is this? and can we get rid of it?
```
AlanRynne, line 24: I think this is one for @connorivy to answer 😄

AlanRynne (approved):
Left a comment about one of your code comments regarding Tekla.

Other than that it's looking great! Should require extensive testing this coming days.

PR #2267, opened 2023-03-09
feat(Navisworks): Implement the Collections class and revise the tree conversions

jsdbroughton:
## Description & motivation
The first alpha release handled the hierarchical tree as nested Base Objects. Adopting Collections as a Core Kit Object allows for better interaction with the Speckle Viewer and, more generally, a better semantical description.

Also included is better feedback to users for conversion progress. This removes a bunch of code falling between two stools for tree traversal. Traversal is now only performed Depth First once. Object resolution from selections and saved selections assess the tree for visibility but no longer prevalidates all the objects.

## Changes:

ConnectorNavisworksBindings.Objects:
- Object resolution from saved stream states now returns Tuples of pseudoId and children count
- Selection resolution methods all adopt the same return signature to suit

ConnectorNavisworksBindings.Selections:
- New helper method IsElementHidden this is in addition to IsElementVisible. The logic for both of these is similar, but they aren't the inverse of each other in all circumstances - this is a speed increase

ConnectorNavisworksBindings.Send:
- Tuple signature of selctected objects described above
- Code consistency with other Connectors @elements not @Elements (useful to FE and Viewer
- UX messaging improvements

ConnectorNavisworksBindings.Settings:
- Defaults to including the active viewpoint on all commits

ConnectorNavisworks.Geometry:
- A Navisworks specific Object class `Geometry` which is a stub representing a Geometry collection

ConnectorNavisworks.Properties:
- With Collections adopted and all the previous additional Node conversions added, there is less code branching to handle unknown nodes
- Refactored Adding properties to an extracted method in anticipation of other development tickets

ConnectorNavisworks.ToSpeckle:
- Speed improvement by extracting the DescendantsAndSelf SDK method and including fallouts. Whole tree traversal is no longer required when any node isn't visible.
- View geometry conversion improved
- Geometry conversion refactored for speed improvements

## To-do before merge:
- [x] Merge Claires Collections branch
- [x] Potentially further commits to strip the duplicate class out if CI fails for ambiguous reference reasons 

## Validation of changes:
- Extensive example models tested for consistence
- Manual testing of UI improvements

## Checklist:


Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.Geometry.cs
```diff
@@ -485,3 +485,13 @@ private static IEnumerable<double> MoveAndScaleVertices(Vector3D vertex1, Vector
     }
   }
 }
+
+namespace Objects.Organization
+{
+  public class Geometry : Base
```
AlanRynne, line 491: What is this class for? If it's empty why not use a `Base` instead?

Since this is only available in Navis, it will become a `Base` when received, since the Kit doesn't know about this type, and the deserialisation picks from the Kit classes only.

Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.Properties.cs
```diff
@@ -12,40 +12,27 @@ namespace Objects.Converter.Navisworks
 {
   public partial class ConverterNavisworks
   {
-    internal static Base GetPropertiesBase(ModelItem element, ref Base @base)
```
AlanRynne, line 15: I know this was discussed in Discord birefly... (maybe I should have weighed in)

This is not to stop you from merging this at all... But if this method has _side-effects_ on `@base`... the `ref` was a good indicator that _something_ could happen to that object.

@JR-Morgan may have some opinions on this.

Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.Properties.cs
```diff
@@ -55,12 +42,12 @@ public static string SanitizePropertyName(string name)
     {
       // Regex pattern from speckle-sharp/Core/Core/Models/DynamicBase.cs IsPropNameValid
       return name == "Item"
-        ? "$Item"
+        ? "Item"
         : Regex.Replace(name, @"[\.\/]", "_");
     }
 
     public static void BuildPropertyCategory(PropertyCategory propertyCategory, DataProperty property,
-      ref Base propertyCategoryBase)
```
AlanRynne, line 63: Same as my above comment about `refs`

AlanRynne (commented):
Just a couple of comments!

Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.Geometry.cs
AlanRynne: Placeholder! Will be moved to Objects when it's ready.

Since Navis doesn't receive, deserialisation to Base is a non-issue now.

Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.Properties.cs
AlanRynne: To be discussed with the team what our "ref stance" is. No particular opinions other than it seems like a nice hint to add it when there are side-effects, but the reality is that code-wise it does nothing.

AlanRynne (approved):
All good!

Objects/Converters/ConverterNavisworks/ConverterNavisworks/ConverterNavisworks.Properties.cs
JR-Morgan: As for communicating side effects. I agree, C# language does a poor job of allowing you to enforce this.
Other languages have `const` correctness, or object immutability rules
Some would say the way to go is to have getter only interfaces,
Some would say that you should just document when you mutate objects and/or make it obvious by the method name.

I don't have much wisdom to offer on this, other than, I don't think the `ref` keyword is the way to go.

PR #2275, opened 2023-03-12
fix(gh): Extract parameter position is now consistent

AlanRynne:
Fixes an issue reported by SOM where our `ExtractParameter` logic would place the resulting node in the wrong location.

This seemed to be an issue with HighDPI screens and how Grasshopper computes the canvas coordinates based on that.

Using `Global_Proc.AdjustUI` now to fix that.


teocomi (approved):
Haven't tested this, but looks safe enough!

PR #2277, opened 2023-03-13
feat(objects): RevitWall by Face now stores BREP instead of Surface

### LINKED ISSUE #2132: FaceWall conversion does not support trimmed Surfaces
Our current FaceWall conversion logic from Rhino/GH > Revit does not support trimmed surfaces as we're only picking the first Surface available in the input Brep; see https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Objects/BuiltElements/Wall.cs#L153

We should change this to actually store the full Brep input, and in Revit change the conversion logic accordingly, see https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial%20Classes/ConvertFaceWall.cs#L44

Input test files: https://drive.google.com/drive/folders/1nn8f9ciim4Om3GRfxiXyn2b-yVe2Zlfg?usp=share_link
Forum thread: https://speckle.community/t/revit-wall-by-face/4827/6

AlanRynne:
Fixes #2132 

Changes the surface storage property in a `RevitWall by Face` to instead store a BREP.

This aligns with how Revit works, as it expects a "solid with a single face" as   opposed to a "Nurbs surface".

This means you an now have weird trims and holes in your Face by Wall instances.

Screenshot 2023-03-13 at 00 10 37


teocomi (approved):
I can't build the T4 locally so it'll be a straight approve on my end :D

AlanRynne:
I'll open an issue for the TT file failing on local to check it out. Jedd told me the same a couple days ago.

PR #2282, opened 2023-03-13
feat(AdvanceSteel): Specializing Beam to Polybeam

phliberato:
## Description & motivation

- Specializing Beam to Polybeam structure
- Improved coordinateSystem Aligned using centroid as the first point
- User Attributes added in all Advance Steel Objects
- Filtering features of Advance Steel objects to don't show in selected objects
- Descriptor of Advance Steel objects simplified.

## Changes:

- BaseLine with many segments and UserAttributes for all AS objects
https://speckle.xyz/streams/1a0090e6fc

 Descriptor of Advance Steel objects simplified.

Screenshot 2023-03-13 at 10 49 19

Autocad objects have also been simplified, but the Speckle object description has not.

Screenshot 2023-03-13 at 10 50 02

## Checklist:


Objects/Objects/BuiltElements/AdvanceSteel/AdvanceSteelObject.cs
```diff
@@ -0,0 +1,31 @@
...
+  public class AdvanceSteelObject : Base
+  {
+    public AdvanceSteelObject()
+    {
+        
+    }
+
+    private string __typeAdvanceSteel;
+
+    public override string speckle_type
```
teocomi, line 17: Can you please explain this? I think we should not touch the `speckle_type` as it would affect deserialization

Objects/Objects/BuiltElements/AdvanceSteel/AdvanceSteelStraightBeam.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Speckle.Core.Kits;
+
+namespace Objects.BuiltElements.AdvanceSteel
+{
+  public class AdvanceSteelStraightBeam : AdvanceSteelBeam
+  {
+    [SchemaInfo("AdvanceSteelStraightBeam", "Creates a Advance Steel straightBeam.", "Advance Steel", "Structure")]
```
teocomi, line 10: The schema info props are used in Grasshopper to dynamically generate nodes via which the users can create "native elements" for Revit, AdvanceSteel etc.. Since this class/constructor doesn't have any props to then create a native AdvanceSteelStraightBeam we should omit it for now.
See: https://speckle.guide/user/grasshopper.html#grasshopper-bim

phliberato: ok. I will do that

Objects/Objects/BuiltElements/AdvanceSteel/AdvanceSteelObject.cs
phliberato: Because of this call in ConnectorBindingsAutocad.
Screenshot 2023-03-13 at 14 43 21

I think changing just the AdvanceSteelObject won't affect the deserialization. The AdvanceSteelObject objects have specific names.
BTW, speckle_type is virtual in Base class. But if you want, we can create another property in Base Class to use in Report Object.

Core/Core/Models/Base.cs
```diff
@@ -276,6 +276,20 @@ public virtual string speckle_type
         return __type;
       }
     }
+
+    private string __simpleType;
+
+    [JsonIgnore]
+    public string simple_Type
```
JR-Morgan, line 283: I would prefer not to add this as a property to `Base`

Calling `myObject.GetType().Name` in the report seems fine, no?

Adding new properties to `Base` requires a fair bit of commitment from us to maintain them, ensure don't break, unit test, align across SDKs, and document.
And in this case, I don't think this property brings enough value to justify that.

JR-Morgan (changes requested):

Core/Core/Models/Base.cs
phliberato: ok! I got it! I changed the report call.

teocomi (approved):

PR #2300, opened 2023-03-17
fix(Navisworks): Capture off thread Exceptions bubbling up to cause host app failure

jsdbroughton:
It also corrects for a progress bar glitch introduced with the Exception handling improvements

Allows the onErrorProgress action to populate and then handle once the Async has completed.


AlanRynne (approved):
🚀

PR #2307, opened 2023-03-20
fix(Revit): don't send unassigned parameters

connorivy:
## Description & motivation

Edge case reported on forum where object was being sent to Speckle but was unable to be received back. The issue was that the object had a parameter which would not accept a 0 value.

Trying to set the value to 0 on the frontend would results in the following error
image

The issue occurs when there is no value assigned to the parameter, the "AsInteger" property evaluates to 0 because that is the default value of integers.
image

Therefore we want to check if there is actually a value assigned to a parameter before we send the parameter to Speckle

## Changes:

- add check for param.HasValue

## Validation of changes:

this object can now be sent and received successfully

## Checklist:

## References

https://speckle.community/t/revit-connector-issue-receiving-lighting-fixtures/5129/2


teocomi (approved):
Txs Revit API! :)

Issue #2258, opened 2023-03-08
Our logging effort clashes with RhinoCompute


gjedlicska:
Rhino compute [uses](https://github.com/search?q=repo%3Amcneel%2Fcompute.rhino3d%20serilog&type=code) `Serilog` too as their logging platform.

If a user has the `Rhino / Grasshopper Connector` installed and runs Rhino Compute on that machine, our current logs are merged with their logs resulting in:
* us potentially overriding version info in their log
* us leaking the hashed userId to their log
* rhino compute errors making their way into our logging systems

The fix for this is to not use the Serilog global static logger, but to store our own instantiated logger in our own static property.
This is mostly a brainless refactor, but it has to happen across all connectors for consistency.

gjedlicska:
hey @JR-Morgan can i drop this ball on you? The core part is done in #2313, we need to make sure connectors use the `SpeckleLog.Logger` global static instance instead of the `Serilog.Log` convenience methods, to avoid hijacking other plugin devs logs.

Once we're done with the refactor in sharp and manager, we need to remove this [line](https://github.com/specklesystems/speckle-sharp/pull/2313/files#diff-e8ffa496d274af4e06de148c7c03c4dd580f54508d47c7438a1b833d74de0571R105)

JR-Morgan:
Cool! on it

teocomi:
Closing as temporarily fixed.
Related to https://github.com/specklesystems/speckle-sharp/issues/2477

PR #2313, opened 2023-03-21
fix(core logging): make sure core logger uses our own SpeckleLog logger, not the global serilog instance

gjedlicska:
This is a non breaking change, but enables the fixing of #2258 

Also changes the seq api key to a new one.


AlanRynne:
I'm assuming we still need to update any other connector references to `Log.XXX` right?

gjedlicska:
You are correct @AlanRynne,

That's why i left the Serilog `Log.Logger` assignment in for the time being, to not break things.

On the connectors side I've passed the baton to @JR-Morgan.

AlanRynne (approved):
Nice work! I'm approving and merging to do the connectors change on top.

PR #2315, opened 2023-03-22
feat: Use `SpeckleLog.Logger` instead of global `Log`

AlanRynne:
Follows #2313

Updated all references I found.

I skipped Triangles as it seemed to be using something different


gjedlicska:
git is very confused about the reformatting in some places. but looks good to me

PR #2317, opened 2023-03-22
fix(Revit): don't modify points when flattening floor outline

connorivy:
## Description & motivation

in order to make a floor, Revit needs to be passed a flat floor outline, even if the floor is actually sloped. Previous we were editing the floor outline to be flat, but this ended up messing up the import of analytical panels which do not need to be passed in as a flat outline.

## Changes:

- return a copy of of the floor outline with flattened points instead of changing the values of the object

## Screenshots:

Receiving [this commit](https://latest.speckle.dev/streams/85bc4f61c6/commits/ef41f8402c) results in many object baking failures including one that causes the entire transaction to be rolled back
image

After this PR, the commit bakes with no failures
image

## Checklist:

## References

Forum post that brough this issue to my attention
https://speckle.community/t/errors-in-receiving-slabs-and-walls-from-etabs-to-revit/5183/4

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
@@ -47,8 +47,8 @@ namespace Objects.Converter.Revit
         level = ConvertLevelToRevit(LevelFromCurve(CurveToNative(speckleFloor.outline).get_Item(0)), out ApplicationObject.State state);
       }
 
-      ChangeZValuesForCurves(speckleFloor.outline, level.Elevation);
-      var outline = CurveToNative(speckleFloor.outline, true);
+      var flattenedOutline = GetFlattenedCurve(speckleFloor.outline, level.Elevation);
+      var outline = CurveToNative(flattenedOutline, true);
       UnboundCurveIfSingle(outline);
 
       if (!GetElementType<FloorType>(speckleFloor, appObj, out FloorType floorType))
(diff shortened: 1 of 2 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
```diff
@@ -196,38 +196,68 @@ private List<ICurve> GetProfiles(DB.CeilingAndFloor floor)
...
+              line.start.y,
+              z * Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, line.start.units),
+              line.start.units),
+            new OG.Point(
+              line.end.x,
+              line.end.y,
+              z * Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, line.end.units),
+              line.end.units),
+            line.units
+            );
```
AlanRynne, line 220: I suggest getting the conversion factor just once for the line (point units should always be the same as the line units anyway)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
```diff
@@ -196,38 +196,68 @@ private List<ICurve> GetProfiles(DB.CeilingAndFloor floor)
...
+          for (var i = 0; i < nurbs.points.Count; i++)
+          {
+            if (i % 3 == 2)
+              curvePoints.Add(z * Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, nurbs.units));
+            else
+              curvePoints.Add(nurbs.points[i]);
+          }
+          var newCurve = OG.Curve.FromList(curvePoints);
+          newCurve.units = nurbs.units;
+          return newCurve;
```
AlanRynne, line 238: Same as the line. Computing the conversion factor just once (outside of the loop) will have some performance improvement when curves have a high number of points.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
```diff
@@ -196,38 +196,68 @@ private List<ICurve> GetProfiles(DB.CeilingAndFloor floor)
...
+          for (var i = 0; i < originalPolylinePoints.Count; i++)
+          {
+            if (i % 3 == 2)
+              polylinePonts.Add(z * Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, poly.units));
+            else
+              polylinePonts.Add(originalPolylinePoints[i]);
+          }
+          var newPolyline = OG.Polyline.FromList(polylinePonts);
+          newPolyline.units = poly.units;
+          return newPolyline;
```
AlanRynne, line 252: Polyline also needs to compute it's conversion factor just once😉

AlanRynne (changes requested):
Just a couple of comments based on @JR-Morgan's performance workshop (which addressed a similar issue with calculating the conversion factor)

AlanRynne (approved):
Yay! 🚀 Into dev it goes.

PR #2323, opened 2023-03-23
fix(core/rhino/gh/revit/navis): Adds enhancedLogContext config option and disables it on mac

AlanRynne:
Fixes logger initialisation issues in:

- Rhino for mac
- Grasshopper
- Revit
- Navisworks


AlanRynne:
Merging in. Discussed changes with @gjedlicska, and verified connectors locally.

PR #2324, opened 2023-03-23
Jrm/objects/instance serilisation fix

JR-Morgan:
Fixed issue with serializing block instances


AlanRynne (approved):
Good to go! Very deep rabbit hole you went down through to find this...🚀

PR #2325, opened 2023-03-23
fix(gh): TreeBuilder was checking for List<object> when it should be IList instead

AlanRynne:
Fixes issue where tree builder would output a `List` object, instead of a tree with a single branch in it.

This was caused because we expected a `List<object>` but that doesn't cover all cases. Instead we should be checking for `IList`

The DSO node already did this so this was an oversight on my behalf


AlanRynne:
I'm approving my own change for the sake of making `beta5` already... 🙇🏼‍♂️

PR #2336, opened 2023-03-25
fix(autocad): Fixed issue with transforms that aren't perfectly perpendicular 

JR-Morgan:
It seems autocad is really picky about the precision of transform matrices.

It expects basis vectors to be perfectly perpendicular, within DOUBLE precision tolerances.
We recently switched to using `System.Numerics.Matrix4x4`, which uses 32bit floats.

This was leading to many cases when receiving instances, Autocad would raise an exception when setting the native blockDefinition.Transform

It seems that double -> float rounding can lead to autocad considering a transform non-perpendicular (`!IsScaledOrtho`)

---

What I've done:

<strike>1. Swapped `System.Numerics` out for `System.DoubleNumerics` https://www.nuget.org/packages/System.DoubleNumerics/</strike> (reverted)
2. Added `MakePerpendicular` function to Autocad converter utils, that will re-calculate the basis vectors to ensure they are perpendicualr.

<strike>Change no.1 fixed receiving most commits, but it seems that a few blender commits (blender sends 32 bit float)</strike> (reverted)
Change no.2 fixed receiving all commits I've tested with, from different host apps.

<strike>Potentially, Change no.2 is all that is required to get receiving working in autocad.
Using 32bit float matrices may lead to slight loss of precision, but would be a smaller change for this PR.
So we should consider are options here./</strike> (reverted)

### DIFF
--- a/Core/Core/Core.csproj
+++ b/Core/Core/Core.csproj
@@ -53,7 +53,6 @@
       <Aliases></Aliases>
     </PackageReference>
     <PackageReference Include="Microsoft.Data.Sqlite" Version="7.0.3" />
-    <PackageReference Include="System.DoubleNumerics" Version="3.1.3" />
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Serialisation\" />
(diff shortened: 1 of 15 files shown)


AlanRynne (approved):
This looks good on my end ;)

PR #2344, opened 2023-03-27
fix(Autocad): MText height property shows 0

phliberato:
## Description & motivation

- Use TextHeight property instead of height property of MText object when converting to Speckle obj
https://help.autodesk.com/view/OARX/2023/ENU/?guid=OARX-ManagedRefGuide-Autodesk_AutoCAD_DatabaseServices_MText_TextHeight

## Checklist:


clairekuang:
Hi @phliberato , what's the reasoning behind using TextHeight instead of Height for multiline text? Full height information is still useful for other connectors, but if the textheight property is needed it would be better to add that as a dynamic property

clairekuang (changes requested):
I'd add `TextHeight` as a dynamic prop instead of replacing `Height`, since `Height` is still a useful prop (`TextHeight` is the height of the text per line, `Height` is the total height of the Mtext object including all lines of text and spacing between lines). Not sure why some MText `Height` props return 0, in this case it may be stored in the `ActualHeight` prop and you could add a check for that instead.

phliberato:
> I'd add `TextHeight` as a dynamic prop instead of replacing `Height`, since `Height` is still a useful prop (`TextHeight` is the height of the text per line, `Height` is the total height of the Mtext object including all lines of text and spacing between lines). Not sure why some MText `Height` props return 0, in this case it may be stored in the `ActualHeight` prop and you could add a check for that instead.

phliberato:
@clairekuang  I've attached an example that shows the different behaviors between DBText and MText. In MText I think it's better to use the Text property instead of Content. Take a look at the results.

Screenshot 2023-04-14 103652

Screenshot 2023-04-14 103858

[TextsIssue.zip](https://github.com/specklesystems/speckle-sharp/files/11233516/TextsIssue.zip)

clairekuang (approved):

PR #2345, opened 2023-03-27
fix: Legacy Ellipse curve checks + missing return on ellipse clause in CurveToSpeckle

AlanRynne:
Fixes issue reported by Pavol where a specific brep was not converting properlly.

Turned out it was a clash of a legacy edge case condition in the nurbs curve conversion + a bug in the actual conversion of ellipses inside `ConvertToSpeckle`


AlanRynne:
Live reviewed by Jedd and myself.

Both the test BREP from notion and ellipses are coming in correctly

PR #2346, opened 2023-03-27
fix(Navisworks): avoid the `Item` reserved property name for Indexed data structures

jsdbroughton:
- [x] Is @jsdbroughton a forgetful idiot


AlanRynne (approved):
Nice self realisation, and a good reminder to everyone else! 😅 had forgotten about this edge case too

PR #2347, opened 2023-03-27
feat(Autocad):  Curve contour extractor to displayvalue in MText and DBText

phliberato:
## Description & motivation

- Curve contour extractor to displayvalue in MText and DBText

Known issues:

- The displaced matrix is not perfect
- Speckle viewer creates an extra point(plane property in Text speckle object?)
- Can't find yet the font name of SHX file
- It will be better if we could reduce the number of curves
- It's not handling multiline strings
- Find a better way to insert PresentationCore and WindowsBase dlls in a project
- Check for other segment types in System.Windows.Media to convert (PolyQuadraticBezierSegment, QuadraticBezierSegment and ArcSegment)

## Checklist:


Objects/Converters/ConverterAutocadCivil/ConverterAutocad2021/ConverterAutocad2021.csproj
```diff
@@ -29,6 +29,8 @@
   </Target>
 
   <ItemGroup>
+    <PackageReference Include="WindowsBase_Core.STW" Version="4.0.30319.1" />
+    <PackageReference Include="EPP.PresentationCore" Version="1.0.0" />
```
teocomi, line 33: Are these needed in the converters?
Even if they were, we should probably be using the `System.Windows` ones, no?

phliberato: Maybe, but I didn’t get. The purpose of this PR is a initiative of converting CAD texts to speckle. It’s a good thing not merge with Main or Dev yet…. I think so

phliberato: It’s working in a reasonable way.

teocomi: Ok! Thanks, Pedro!
Our main concern is around performance implications, we will do some tests to see how it might affect sending, if it's not significant we'll be happy to include this in.

Let us know if you have done any tests already.

In the future, ideally, the viewer will be able to natively display text without us converting it to lines :)

phliberato: Good news! I also think the right way to do this is to make the viewer able to display texts natively (using normal vector, box, font and text).

phliberato: I did few tests, so I don't have benchmarks of it. But I think the main pain of this approach is the connector processing. Perhaps this can slow down the process.

teocomi:
Hey @phliberato , thanks again for this PR! 🙏

We've tested it quite a bit, and unfortunately, the performance implications outweigh the benefits for now, so we'll rather wait for proper text support in the viewer.

One of the sample files I used is this, which is quite text heavy, probably more than most drawings, but the send process was just taking too long (minutes on every text object), and I had to stop it entirely: [1344464051.zip](https://github.com/specklesystems/speckle-sharp/files/11224499/1344464051.zip)

Let us know if you'd like to discuss some more, otherwise, I'll close it for now.

phliberato:
> I agree with you. The performance of this approach is a bottleneck in the conversion process. I think it's not necessary to discuss further, but if the text support in the viewer is lagging and if you want that too, we can try to replace Windows and presentation libraries with Skia.
https://medium.com/medialesson/generating-a-curved-text-in-net-d8a38209c5e2
But at this moment you can close this PR.

PR #2352, opened 2023-03-28
fix(revit): correct transforms when sending linked models and with a reference point setting

clairekuang:
## Description & motivation

Fixes how display value meshes were being retrieved on instance definitions, as well as taking into consideration linked models and reference point settings when computing transforms.

Now revit instance display values are always retreived relative to the internal origin, and the linked model and reference point setting transforms are baked into the instance transform.

Also includes a small fix for autocad, which now correctly traverses revit instances on receive.

## Changes:

- Revit converter
- Autocad connector and converter


JR-Morgan (approved):

JR-Morgan:
I ended up creating a test file to help test this.

image

Everything looks correct for walls + generic instances when sending/receiving using the different coordinate options.
image

Nested instances ...

Rotations on the base point also seem to work fine!
image

PR #2358, opened 2023-03-30
fix(gh): Lower AccountList "no accounts" message to be warning

AlanRynne:
Requested by Arup folks.

This was causing their scripts to fail on Rhino compute and return a 500 error.

A warning is all that is necessary for a user that has no accounts to be aware of what to do about it, and will allow any Compute run that uses this node to succeed.


AlanRynne:
Its a one-liner so I'm approving my own change. Had a chat with Matteo about this already.

PR #2377, opened 2023-04-03
fix(Navisworks): Adds <UseWpf> to projects to enable CICD builds on linux/mac

jsdbroughton:
Fails to build without


AlanRynne (approved):
Beautiful! Thx! This makes me so happy

PR #2387, opened 2023-04-04
fix(revit): SpeckleSchema logic for DirectShapes

### LINKED ISSUE #2378: fix(rhino): Mapping tool does not send BREPs with material
When geometry is sent to Speckle and received in Revit, it comes in the `Generic Models` category with materials applied✅. 

But, this is not the case when sent via the Mapping tool❌. Geometry is received without any materials.

https://file.notion.so/f/s/2d40bc75-669b-40fb-8be2-f1a7878d146f/vWd2uNHTMX.mp4?id=0e78368f-eee7-44f7-bbf5-6e6360bc50da&table=block&spaceId=2e89d096-7145-4cc5-93d3-57fdedb465c0&expirationTimestamp=1680600950628&signature=6_z8_nyHDSeMTIIDJbi1rgBfwr3JUEKqzPiblStAOaw

AlanRynne:
Fixes #2378 

The old logic (from 2021!!! 😱) of our `@speckleSchema` swapping mechanism in the Revit converter was totally broken for `DirectShapes`

A full fix is necessary which may require the addition of an extra attribute on the main geometry property we need to swap.

To go from `geometry->SpeckleSchema` to `SpeckleSchema->geometry` in a consistent way that does not require guesswork.

This fixes the issue at hand for now, but we should re-engineer a better solution once we get to re-engineering mappings

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
@@ -80,7 +80,7 @@ namespace Objects.Converter.Revit
     /// Used to cache already converted family instance FamilyType deifnitions
     /// </summary>
     public Dictionary<string, Objects.BuiltElements.Revit.RevitSymbolElementType> Symbols { get; private set; } = new Dictionary<string, Objects.BuiltElements.Revit.RevitSymbolElementType>();
-    
+
     public Dictionary<string, SectionProfile> SectionProfiles { get; private set; } = new Dictionary<string, SectionProfile>();
 
     public ReceiveMode ReceiveMode { get; set; }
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -408,9 +408,26 @@ public object ConvertToNative(Base @object)
         // find self referential prop and set value to @object if it is null (happens when sent from gh)
         if (CanConvertToNative(speckleSchema))
         {
-          var prop = speckleSchema.GetInstanceMembers().Where(o => speckleSchema[o.Name] == null)?.Where(o => o.PropertyType.IsAssignableFrom(@object.GetType()))?.FirstOrDefault();
-          if (prop != null)
-            speckleSchema[prop.Name] = @object;
+          //if you can find a "MainParamProperty" get that
```
clairekuang, line 411: minor, but should remove this comment

clairekuang (approved):

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
AlanRynne: oopsies! thanks!

AlanRynne: did one better and moved the whole thing to it's own function, plus moved around comments so that they'd make sense

PR #2389, opened 2023-04-04
Revert "Fix(CSiBridge): remove objs ref"

AlanRynne:
This has been fixed in the latest `dev` branch while doing the monorepo cleanup and the change now conflicts with the `dev` branch.

If this is the way CSIBridge has always been shipped, I saw we just wait 1 release cycle and it will fix itself with no conflicts 🙌🏼

That is, unless it was removed for a specific reason on `2.13.0-rc3`


connorivy (approved):

AlanRynne:
I'm merging this in now that we've released 2.13 stable.

PR #2391, opened 2023-04-05
feat(revit): DirectShape conversion is now capable of updating existing DS

### LINKED ISSUE #1674: Support updating DirectShape elements
## Description

Reported **Gaby Pena** of [Studio NA](https://www.studiona-mx.com/).

Common workflow users follow is they use Revit to document their designs from Rhino. Studio NA is using Speckle for this integration. They are mostly involved in Interior Design projects.

Since Revit is used for documentation, dimensions/annotations play an important role here. However, our Update method deletes all the received geometry and creates new ones in their place, instead of updating unchanged parts and creating only the changed parts. Because we are deleting and creating, all the added dimensions get deleted. To explain it through an example:

> Gaby sends two design alternatives to Speckle. The only difference between the two is that the handles on the second one are rotated 30 degrees. Other than that everything is the same, the second commit is created by using the source geometry of the first commit. 
Gaby receives the first commit and adds dimensions and when she receives the second commit, all dimensions are deleted.
> 

`📌 Received geometry creates **Direct-Shapes** in Revit. This can be seen in the GIF below👇:`

Revit_zh8ycxyv5l

## Expected Behaviour

When we receive the geometry in Revit, the parts that do not change should not be deleted and recreated, they should not be touched.

### Commit 1️⃣:

[https://latest.speckle.dev/streams/6b24f9dc21/commits/893a9c8083](https://latest.speckle.dev/streams/6b24f9dc21/commits/893a9c8083)

### Commit 2️⃣:

[https://latest.speckle.dev/streams/6b24f9dc21/commits/eee69a452a](https://latest.speckle.dev/streams/6b24f9dc21/commits/eee69a452a)

## Steps to reproduce

1. Activate Revit connector.
2. Receive first commit.
3. Add dimensions to geometry in plan view.
4. Receive second commit.
5. You’ll see that all geometries have been created from scratch and dimensions have been deleted.

## Files Used

**Revit File** 

[https://drive.google.com/open?id=12obi8EF2YsEcvJxL4LusWElElnMZhrJd&authuser=bilal%40speckle.systems&usp=drive_fs](https://drive.google.com/open?id=12obi8EF2YsEcvJxL4LusWElElnMZhrJd&authuser=bilal%40speckle.systems&usp=drive_fs)

**Rhino File**

[https://drive.google.com/open?id=12qpfDXE5mHAB5xVTXESUvuMyU8QISWAN&authuser=bilal%40speckle.systems&usp=drive_fs](https://drive.google.com/open?id=12qpfDXE5mHAB5xVTXESUvuMyU8QISWAN&authuser=bilal%40speckle.systems&usp=drive_fs)

AlanRynne:
Fixes #1674 

I refactored the entire direct shape logic so that:
- DirectShapeToNative can only take a `DirectShape` object, and has no fallback DXF handling
- TryDirectShapeToNative calls `DirectShapeToNative` and handles DXF fallback
- Other `TryDirectShapeToNative` calls accept the existing options (Mesh, Brep, List<Mesh>) but now consistently handle DXF fallback mode too.
- Removed the previous `DirectShapeToNative` overloads
- Updated any use of the removed functions throughout the converter

DirectShape now checks if the object exists in the doc and, if so, will no longer delete it. Instead it will grab that existing DirectShape and call `SetShape` on it.

The previous logic would remove the existing direct shape and create a new one. This will still happen in create mode, only update mode affects this as expected.

This means it will keep references to any part of the DS that remains unchanged. This is managed by revit in an obscure way so we have no way of controlling how it determines if geometry is new or not.

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
@@ -388,7 +388,7 @@ namespace Objects.Converter.Revit
       }
       return speckleSchema;
     }
-    
+
     public object ConvertToNative(Base @object)
     {
       // Get setting for if the user is only trying to preview the geometry
@@ -408,7 +408,7 @@ namespace Objects.Converter.Revit
           //List<GE.Mesh> meshes = (List<GE.Mesh>)property;
           var cat = GetObjectCategory(@object);
           var speckleCat = Categories.GetSchemaBuilderCategoryFromBuiltIn(cat.ToString());
-          return DirectShapeToNative(new ApplicationObject(@object.id, @object.speckle_type), meshes, speckleCat);
+          return TryDirectShapeToNative(new ApplicationObject(@object.id, @object.speckle_type), meshes, ToNativeMeshSetting, speckleCat);
         }
         catch
         {
@@ -432,7 +432,7 @@ namespace Objects.Converter.Revit
             return null;
         }
       }
-      
+
       // Check if object has inner `SpeckleSchema` prop and swap if appropriate
       @object = SwapGeometrySchemaObject(@object);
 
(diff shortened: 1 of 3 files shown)


teocomi:
yassss

connorivy (approved):
Added working functionality and got rid of 30 lines of code. It's beautiful.

PR #2415, opened 2023-04-12
fix(revit): floor outline flattening fix

teocomi:
Fixes possibly two forum reported bugs:

- https://speckle.community/t/revit-errors-importing-a-floor/5420
- maybe this one too (not tested): https://speckle.community/t/revit-to-etabs-groups/5419

Test files (revit 2023):
[floor bug.zip](https://github.com/specklesystems/speckle-sharp/files/11213994/floor.bug.zip)

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
@@ -240,17 +240,16 @@ namespace Objects.Converter.Revit
 
         case OG.Polyline poly:
           var polylinePonts = new List<double>();
-          var originalPolylinePoints = poly.ToList();
+          var originalPolylinePoints = poly.points;
           var polyConversionFactor = z * Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, poly.units);
           for (var i = 0; i < originalPolylinePoints.Count; i++)
           {
-            if (i % 3 == 2)
-              polylinePonts.Add(z * polyConversionFactor);
-            else
-              polylinePonts.Add(originalPolylinePoints[i]);
+            polylinePonts.Add(originalPolylinePoints[i].x);
+            polylinePonts.Add(originalPolylinePoints[i].y);
+            polylinePonts.Add(originalPolylinePoints[i].z * polyConversionFactor);
           }
-          var newPolyline = OG.Polyline.FromList(polylinePonts);
-          newPolyline.units = poly.units;
+          var newPolyline = new Polyline(polylinePonts, poly.units);
+          newPolyline.closed = poly.closed;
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
```diff
@@ -240,17 +240,16 @@ private ICurve GetFlattenedCurve(ICurve curve, double z)
 
         case OG.Polyline poly:
           var polylinePonts = new List<double>();
-          var originalPolylinePoints = poly.ToList();
+          var originalPolylinePoints = poly.points;
           var polyConversionFactor = z * Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, poly.units);
```
connorivy, line 244: I realize that you didn't write this (I did) but this line is wrong. We shouldn't be multiplying this value by z. It should just be ```var polyConversionFactor = Speckle.Core.Kits.Units.GetConversionFactor(ModelUnits, poly.units);```

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
```diff
@@ -240,17 +240,16 @@ private ICurve GetFlattenedCurve(ICurve curve, double z)
...
           for (var i = 0; i < originalPolylinePoints.Count; i++)
           {
-            if (i % 3 == 2)
-              polylinePonts.Add(z * polyConversionFactor);
-            else
-              polylinePonts.Add(originalPolylinePoints[i]);
+            polylinePonts.Add(originalPolylinePoints[i].x);
+            polylinePonts.Add(originalPolylinePoints[i].y);
+            polylinePonts.Add(originalPolylinePoints[i].z * polyConversionFactor);
```
connorivy, line 249: This last line should be ```polylinePonts.Add(z * polyConversionFactor);```. We're changing all z values to the same value, not scaling each individual value.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFloor.cs
```diff
@@ -240,17 +240,16 @@ private ICurve GetFlattenedCurve(ICurve curve, double z)
 
         case OG.Polyline poly:
           var polylinePonts = new List<double>();
-          var originalPolylinePoints = poly.ToList();
+          var originalPolylinePoints = poly.points;
```
connorivy, line 243: The compiler warns that Polyline.points is obsolete and we should use Polyline.GetPoints(). Not sure if this actually matters though because they both do the same thing

connorivy (changes requested):

PR #2416, opened 2023-04-12
reformat(core): Core reformatted

JR-Morgan:
Core has had a makeover and is now super sexy 🔥 

Applied `Format Code` + `Cleanup Code` + a tiny bit of manual intervention where needed


AlanRynne (approved):
I think we're ready to go!!!!

PR #2417, opened 2023-04-12
fix(Navisworks): Conversion fails due to single unconverted property

jsdbroughton:
Not equal comparison fail for non equal type operands fixed with `!x.Equals(y)`


jsdbroughton (commented):
Format cleanup hides the single line fix :/

teocomi (approved):

PR #2422, opened 2023-04-13
build(revit): Launch config

JR-Morgan:
Removed ConverterRevit launch configurations and added ConnectorRevit launch configurations.

Added ConverterRevitXXXX as a implicit build dependency of ConnectorRevitXXXX
Such that building the connector will trigger a rebuild of the converter!


connorivy (approved):
It's great to have the connectors as the default projects instead of the converters

PR #2423, opened 2023-04-13
fix(revit): fixed traversal issue of displayValues

JR-Morgan:
As reported by @KatKatKateryna 

The update to new traversal logic introduced a small regression where raw `Base` objects with `displayValue`s would not have their `displayValue`s converted. 

I've tweaked the rules to traverse into displayValues. And this works fine for now. But we may need to look at implementing a 
`DisplayValueToNative` and/or a `MeshesToNative` function to allow for 1-1 object conversion


Core/Core/Models/GraphTraversal/DefaultTraversal.cs
```diff
@@ -21,7 +24,6 @@ public static GraphTraversal CreateTraverseFunc(ISpeckleConverter converter)
       .ContinueTraversing(b =>
       {
         var membersToTraverse = b.GetDynamicMembers()
-          .Concat(displayValueAliases)
```
jsdbroughton, line 24: Removing this from the general default traversal - was it here to address what was intended for Revit but isn't valid for others?

Was this the QGIS data being replicated issue?

jsdbroughton (approved):

PR #2429, opened 2023-04-14
Replaced triangle net source with nuget

JR-Morgan:
@phliberato


JR-Morgan:
Good point @phliberato, there are also some unknowns on our end with packaging so let's leave this for now.
I'll ask @teocomi what he thinks of us publishing the triangle net on our nugets

teocomi:
> Good point @phliberato, there are also some unknowns on our end with packaging so let's leave this for now. I'll ask @teocomi what he thinks of us publishing the triangle net on our nugets

We already have several nugets, so would be happy to do so for this one too :)

AlanRynne:
Reopened this now that we have our own NuGet published!

AlanRynne (approved):
I'm blessing my own changes 🙌🏼

Issue #2436, opened 2023-04-14
enhancement(dui): Allow for access to branches greater than 100


jsdbroughton:
The server query allows for 100 branches to be returned. We have received requests from teams with complex branch management that more should be possible to be returned.

This would require DUI to make paged queries to the server.

Issue #2437, opened 2023-04-14
enhancement(dui): Allow access to branches (& streams, commits) via combobox where number of options exceed `n`


jsdbroughton:
Related to #2436 but less critical.

Selecting branches in DUI would be improved where there is a large number if instead the interface to do so would be from a combo box (combination of search and drop-down) in place of a plain drop-down.

teocomi:
Definitely will come in DUI3!

PR #2440, opened 2023-04-14
feat(core): Adds Pagination to StreamGetBranches and increases limit requested by DUI

jsdbroughton:
Closes: #2436 

## Description & motivation

Allows for use cases heard from Speckle users requiring more than 100 branches as they address cataloguing or complex team information management. The server allows a maximum of 100 results per query; the current solution takes this as a hard limit.

### Core / API / GraphQL / Client.BranchOperations :: StreamGetBranches
- Adds pagination to the query

### DUI2 / ViewModels / StreamViewModel
- increases the Branches requested from 100 to 1000.

## Validation of changes:

Rhino and Navisworks both tested against https://speckle.xyz/streams/01fca0eab7

## Checklist:

### DIFF
--- a/Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.BranchOperations.cs
+++ b/Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.BranchOperations.cs
@@ -72,6 +72,71 @@ public partial class Client
     var res = await ExecuteGraphQLRequest<StreamData>(request, cancellationToken).ConfigureAwait(false);
     return res.stream.branches.items;
   }
+/// <summary>
+  /// Get branches from a given stream
+  /// </summary>
+  /// <param name="streamId">Id of the stream to get the branches from</param>
+  /// <param name="branchesLimit">Max number of branches to retrieve</param>
+  /// <param name="commitsLimit">Max number of commits to retrieve</param>
+  /// <returns></returns>
+  public Task<Branches> PagedStreamGetBranches(string streamId, int branchesLimit = 10, int commitsLimit = 10, string? nextCursor = null)
+  {
+    return PagedStreamGetBranches(CancellationToken.None, streamId, branchesLimit, commitsLimit);
+  }
+
+  /// <summary>
+  /// Get paginated branches from a given stream
+  /// </summary>
+  /// <param name="cancellationToken"></param>
+  /// <param name="streamId">Id of the stream to get the branches from</param>
+  /// <param name="branchesLimit">Max number of branches to retrieve</param>
+  /// <param name="commitsLimit">Max number of commits to retrieve</param>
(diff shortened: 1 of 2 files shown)


Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.BranchOperations.cs
```diff
@@ -67,10 +74,14 @@ public Task<List<Branch>> StreamGetBranches(string streamId, int branchesLimit =
...
-      Variables = new { streamId }
-    };
-    var res = await ExecuteGraphQLRequest<StreamData>(request, cancellationToken).ConfigureAwait(false);
-    return res.stream.branches.items;
+        Variables = new { streamId }
+      };
+      var res = await ExecuteGraphQLRequest<StreamData>(request, cancellationToken).ConfigureAwait(false);
+      branches.AddRange(res.stream.branches.items);
+      nextCursor = res.stream.branches.cursor;
+    } while (branches.Count < branchesLimit && nextCursor != null);
```
AlanRynne, line 82: This while loop is questionable, depending on the inputs it could hit our rate-limit quite easily.

AlanRynne (changes requested):
I would advise against this solution.

The api call should be "pure", meaning it does 1 call when you run it, not potentially 100s. It should be the responsibility of the project using this method to implement pagination on their end.

☝🏼 This would require **adding a cursor input to this call**, which I'd be happy to have.

If this is something commonly done in many connectors, we should create a separate function to get N branches while safely handling all other server considerations, but leave `StreamGetBranches` untouched.

jsdbroughton:
That makes sense. I can review. The idea here would be to do that call recursion in DUI. Thanks

teocomi:
Agree with Alan :)

It also sounds more of a UI/UX problem; how would people deal with 100+ branches in the current DUI?
Or better, how are they planning to use them? This could be pretty informative for DUI3...

jsdbroughton:
> It also sounds more like a UI/UX problem; how would people deal with 100+ branches in the current DUI?
> Or better, how are they planning to use them? This could be pretty informative for DUI3...

This is a UI/UX problem that exists in both Connectors and the web. The specific use case I have referred to in Notion cc-ing @teocomi 

Aside from the reporting user, we see increasing numbers of users with catalogue-like approaches to their Speckle data management. cc @paloknapo 

Where users have greater than 100 branches to a stream, these will be accessible online by URL at least until this may be supported. In connectors, you cannot reach your data. #2437 Suggests a further enhancement to DUI as a UX solution for many branches.

An alternative consideration is to have a "fetch more branches" button.

teocomi:
Closing as currently blocked by https://github.com/specklesystems/speckle-server/issues/1549 - will reopen/revisit when the right time comes.

Issue #586, opened 2021-06-24
Feat (revit): support face hosted family instances


teocomi:
![image](https://user-images.githubusercontent.com/2679513/123244945-47f40480-d4dc-11eb-8fb1-be58450e9477.png)
![image](https://user-images.githubusercontent.com/2679513/123244958-4a565e80-d4dc-11eb-925e-713d6b475a03.png)

Sample files: https://drive.google.com/drive/folders/1QAMO6ytdYEzQpPkhQoqmrk77LRX0TY8c?usp=sharing

Reporter: Marco (Poldo)

clairekuang:
This may have been resolved by #2312, checking now

teocomi:
This issue seems to be still present and will require additional investigations when tackled:
- https://speckle.community/t/speckle-2-12-2-generic-model-with-void-or-solid-face-based-family-as-cutout/5071
- https://github.com/specklesystems/speckle-sharp/pull/2441

clairekuang:
reverting the face-hosted instances to use the old method is re-introducing some problems that sending them as instances fixed:
![image](https://github.com/specklesystems/speckle-sharp/assets/16748799/b6fd29cc-8387-420f-bbc2-6c7babdaa3f2)

It seems like instances are working well for some cases of face hosting, and not others...we'll need to continue investigating this

paloknapo:
Closing as stale.

PR #2441, opened 2023-04-17
fix: exclude face hosted families from new instancing logic

teocomi:
Following [this user-reported bug](https://speckle.community/t/speckle-2-12-2-generic-model-with-void-or-solid-face-based-family-as-cutout/5071) it seems that face-hosted family instances are not converting well when sending from Revit.
This PR excludes then and also improves the exclusion for "Structural Framing" making it language agnostic.
If we want to look at it again it might be tackled together with: https://github.com/specklesystems/speckle-sharp/issues/586

Source:
image

Before:
image

After:
image

Sample file:
`G:\Shared drives\All Company\07 Sample Models\Revit\Generic Model with cutout`


clairekuang (approved):

clairekuang:
reverting the face-hosted instances to use the old method is re-introducing some problems that sending them as instances fixed:
image

PR #2443, opened 2023-04-17
fix(husky): Fixed install on restore in Core target

AlanRynne:
Fixes the new target that runs in Core for every restore. Condition for IsDesktopBuild was wrong.


AlanRynne:
Skipping CI as this is not a code change at all.

PR #2448, opened 2023-04-17
fix(ci): Installer property is no longer needed as we always build them

AlanRynne:


AlanRynne:
Verified deploy job starts and is valid

PR #2454, opened 2023-04-18
Revit/context objects dic

teocomi:
Make `ContextObjects` a Dictionary so that lookup is faster during conversion. Initial tests show a 5x at lest improvement on ModelGroups.


clairekuang (approved):
looks good to me! thanks for this change 👍

PR #2456, opened 2023-04-19
feat(revit): ensures hosted elements are sent as a detached list, fix…

### LINKED ISSUE #2446: Detach hosted elements
After investigating a [user-reported bug](https://speckle.community/t/revit-to-etabs-groups/5419/10), I found that hosted elements (which include items inside a model groups) are not being detached this leads the converter to max out on very large groups, [see example file (confidential)](https://drive.google.com/file/d/1SbzlZe-rMgJr4GLBOjGC6MFQc-Sfnwjg/view).

I'd like to propose changing every "elements" prop instance to be detached with "@elements", with some extra code for backward compatibility. We would need to verify how this affects other connectors (and viewer) who are consuming Revit data (PowerBI, Excel etc).

Initial tests indicate that this would overcome the "Object is too large" issue reported by the user: `revit/host-face-instances`

These are some of the responsible lines: https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs#L145

teocomi:
Fixes #2446 in Revit

Hosted elements are now sent as detached, but on receive we still check for both "elements" and "@elements" for backward compatibility.

Tagging @AlanRynne @didimitrie to check if this might require changes in the Viewer, PowerBi, or other connectors too.


didimitrie:
For viewer testing, is there a quick commit link with a new model group to check it out on our end? Afaik, it should be ok.

teocomi:
> For viewer testing, is there a quick commit link with a new model group to check it out on our end? Afaik, it should be ok.

https://speckle.xyz/streams/af7b11f6e9/commits/377d156dca

PR #2457, opened 2023-04-19
fix(revit): disable getting floor slope on send

teocomi:
We're disabling getting the slope of floors on send as this is causing significant performance issues on large models.
Quick tests made 1h+ conversion time go down to ~4-7min.


AlanRynne (changes requested):
Can't we just delete this whole thing instead of commenting it out? If we need it back at some point, we can always roll back this commit.

teocomi:
Closing in favour https://github.com/specklesystems/speckle-sharp/pull/2466

PR #2469, opened 2023-04-21
add default initialization to logger with Core as the app name

gjedlicska:
initialize logger by default as Core logger instead of throwing


teocomi (approved):

gjedlicska:
Is this still an issue @clairekuang ?

JR-Morgan:
> Is this still an issue @clairekuang ?

No this is 100% fixed, we were just adding tags to older commits to properly categorise our issues.

PR #2476, opened 2023-04-21
Fix(Revit): Recent change to context objects broke groups

connorivy:
## Changes:

- converter should skip an object if it is not in context objects. Context objects is the list of all objects that are going to be sent. If an element is missing from context objects, then we assume it has already converted itself, therefore the parent should not convert it
- wrap subelement conversion in try catch so that if a single hosted element fails to convert, the rest of the hosted elements will continue to convert.

## Checklist:

## References

This is the PR that originally broke this behavior https://github.com/specklesystems/speckle-sharp/pull/2454


teocomi (approved):
Ouch, seems like this was my bad!
Txs

PR #2480, opened 2023-04-24
fix(acadcivil): better handling of application ids

### LINKED ISSUE #2383: papercut(autocad/civil): null applicationId causes exception preventing conversion
Issue encountered receiving data into AutoCAD 22 in 'update' receive mode. We've seen an edge case where an object's applicationId was null, which caused an exception on [checking for an existing object in the model](https://github.com/specklesystems/speckle-sharp/blob/main/ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs#L536). This exception is caught on receiving entire commit obj, so no other objects receive successfully.

I think more ideal behaviour could be either of the following:
- Add a null or empty check on applicationId. If null, return empty list of existing objects and allow the object to be created as new.
- Catch the exception in a lower level of the conversion so just the specific object conversion fails ([somewhere here](https://github.com/specklesystems/speckle-sharp/blob/main/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs#L403)), and other objects will convert/receive successfully.

These are just my initial thoughts, so happy to continue discussion!

### LINKED ISSUE #2462: bug(civil3d): surfaces with the same name are not updating correctly
## Describe the bug

When receiving surfaces in Civil3D in update mode, a duplicate surface is being created with an incremented name `{surface name}(1)` etc instead of the original surface being deleted/recreated.

## To Reproduce
Create a surface with a name
Send and receive in update mode.
## Expected behavior
The existing surface should be updated with the name unchanged, and no duplication.

## Additional context
 See  [notion page](https://www.notion.so/speckle/b7cf9c6a6b2446ac9d42cd721e9b9598?v=356279396980430cb286ea6a0533d864&p=3dddf739109141dba4886f3705a4d44a&pm=s)

clairekuang:
## Description & motivation

Includes two minor improvements to application id handling:
- old acadcivil files may have the older appllication ids saved, which results in unreliable updating. This pr will selectively overwrite saved applicationIds to include the new improved version, which has a file hash appended to the object id.
- applicationId string null or empty handling (previously only checking for null)

Also includes an improvement to civil3d corridor featurelines to include featureline points and break points when sending to speckle. This requires 1) changing the displayvalue of the featureline class to be a list of curves and adding a points property. This was a quick fix requested by Jerome.

Fixes #2462
Fixes #2383
## Changes:

Autocad connector
Civil converter
Objects (Featureline class)


ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
```diff
@@ -1001,18 +1136,42 @@ private void ConvertSendCommit(Base commitObject, ISpeckleConverter converter, S
...
-                "Blocks" :
-                Utils.RemoveInvalidDynamicPropChars(layer); // remove invalid chars from layer name
+              string containerName = obj is BlockReference ? "Blocks" : Utils.RemoveInvalidDynamicPropChars(layer); // remove invalid chars from layer name
 
               if (commitObject[$"@{containerName}"] == null)
                 commitObject[$"@{containerName}"] = new List<Base>();
               ((List<Base>)commitObject[$"@{containerName}"]).Add(converted);
 
               // set application id
+              #region backwards compatibility
```
clairekuang, line 1146: @teocomi this is the main change for handling old acad application ids which have already been saved on the obj

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -486,19 +503,25 @@ public static bool SetObjectCustomApplicationId(DBObject obj, string id, out str
...
+      public static List<ObjectId> GetObjectsByApplicationId(
+        Document doc,
+        Transaction tr,
+        string appId,
+        string fileNameHash
+      )
       {
         var foundObjects = new List<ObjectId>();
-        if (appId == null) return foundObjects;
+        if (string.IsNullOrEmpty(appId))
```
clairekuang, line 523: @teocomi this was the other change for #2383

teocomi:
Txs, looks good!

PR #2481, opened 2023-04-24
fix(acad): exclude adsk nugets from being copied to the output dir

teocomi:
Exclude adsk nugets from output dir


AlanRynne (approved):
🙇🏼‍♂️ How did that not get picked up while we were doing the cleanup?!

This is super! Thx @teocomi

Issue #2392, opened 2023-04-05
Transport Cleanup


JR-Morgan:
I would quite like to clean up some problems in `Operations.Receive`, `Operations.Send`, and our `Transports`

Problems I'm facing:

### 1 Exception handling in Transports is done poorly:
Transports are not consistent in the way they communicate **Non-recoverable errors**

In some cases, **exceptions are being thrown**, other times exceptions are caught and passed to `OnErrorAction`.
[for example](https://github.com/specklesystems/speckle-sharp/blob/0ad7ff3f2211f6d54a15ff49eb80d2575b0cce87/Core/Core/Transports/SQLite.cs#L81-L108)

This has several problems
- This leads to inconsistent error handling, and, since `OnErrorAction` is optional, important exceptions are being swallowed.
- Using `OnErrorAction` for non-recoverable exceptions is bad, as important stack trace info is lossed, and if `OnErrorAction` is null, failures are completely silent.

    
Additionally, Transports are not disposed by Operations.Receive (EVEN WITH `disposeTransports = true`!!!!) if an exception is thrown anywhere that prevents Receive from returning.
         
### 2 Cancellation in Transports is done poorly:

Frequent `null` returns on cancellation make handling cancellation impossible!

Returning null everywhere just leads to random other bits of code exploding `NullReferenceException`s!
`ThrowIfCancelationIsRequested` should be used in most cases instead.

### 3 Continuation in many parts of Core is incorrect, and leads to deadlocks.

`ConfigureAwait(true)` is the default behaviour for awaiting async calls, and will attempt to marshal back to the calling thread once complete. This is "feature" is designed for UI async calls, where you need to sync back to a UI thread.
`ConfigureAwait(false)` allows worker threads to continue execution between async calls. 
This is the "recommend" way to do async in Library code, where you don't need to sync to a main or UI thread.

For most (maybe all) parts of `Core`, we want to use `ConfigureAwait(false)`, and we are not!

1. This leads to deadlocks, e.g. if the treadpool gets exhausted, or a custom Synchronisation context is used, or the awaited thread is waiting to continue on a blocked thread.
    -  This is a major blocker for Unity Connector and leads to frequent deadlocks!
1. There's also performance overhead to using `ConfigureAwait(true)`, and it's totally unnecessary, since we don't care about our Receive operations running on an applications `Main` or `UI` thread.

### 4. `Operations.Send` and `Operations.Receive` are unreadable

[...]

PR #2489, opened 2023-04-26
refactor(revit): small cleanup of connector converter

JR-Morgan:
In preparation for implementing sending Collections, I have attempted a small cleanup of the `SendStream` function in the revit connector.

I have extracted functions and made some tweaks to the exception handling to reduce the number of failure branches, and make the error handling more readable.


DesktopUI2/DesktopUI2/ConnectorHelpers.cs
```diff
@@ -61,12 +59,12 @@ public static async Task<Base> ReceiveCommit(Commit commit, StreamState state, P
...
 
           //Treat all operation errors as fatal
-          throw new SpeckleException($"Failed to receive commit: {commit.id} objects from server", ex);
+          throw new SpeckleException($"Failed to receive commit: {commit.id} objects from server: {s}", ex);
         },
         onTotalChildrenCountKnown: c => progress.Max = c,
-        disposeTransports: true
+        disposeTransports: false
       )
-      .ConfigureAwait(true);
```
teocomi, line 69: Hey @JR-Morgan , any specific reason for these changes?

JR-Morgan: The `ConfigureAwait(true)` was added during the reformat for all usages in DUI2 because we wanted to be extra safe about introducing breaking threading changes. In core we are using `ConfigureAwait(false)` everywhere since we don't care about what thread we execute on.

In this case, we are not doing any UI work that would necessitate us being on the UI thread, so (`ConfigureAwait(false)`) is desirable to avoid the costly synchronization of `ConfigureAwait(true)`

JR-Morgan: And the case of `disposeTransports`, 
There are some issues with our `Operations.Recieve` and `Operations.Send` that mean that transports won't always be disposed (even when passing `disposeTransports: true`

imo, I don't really like how we even expose this as an option, it's far better to let the caller handle disposal, since they are the ones creating and storing the transport.

I've noted in #2392 a few improvements I think we could make to both Operations and transports, since these nitty-gritty details are not self-evident, nor are they a great DX.

teocomi (approved):
Fine to go once the conflict is resolved :)

Issue #2281, opened 2023-03-13
bug(DUI): Amending stream data while Connector is active may trigger unhandled exception


jsdbroughton:
## What package are you referring to?

DUI

## Describe the bug

Sometimes (unfortunately not always) if a Stream State is active in DUI in a connector host application, changes made at the server end can trigger an unhandled exception. Depending on the host application's handling of these final throws, this can lead to fatal crashes.

The exception is consistent:
`Collection was modified; enumeration operation may not execute.`

But manifests in different areas of the Avalonia library:
 
<img width="900" alt="Screenshot 2023-03-13 102837" src="https://user-images.githubusercontent.com/760691/224693649-6a4985a3-b6e0-403e-9d74-ef18b2a54106.png">

or 

<img width="665" alt="Screenshot 2023-03-09 234118" src="https://user-images.githubusercontent.com/760691/224693982-32ab9df5-c728-4435-8a3a-362982ba57b2.png">

## To Reproduce

1. Operate with DUI Stream states in the host application (tested in Navisworks)
2. Amend stream name, branch name, a prior commit name in web UI
3. Host Application deals with the unhandled exception in an unspecified manner

## Expected behaviour

If the observables point to a changed resource and this exception is inevitable, DUI should handle the error and report to the user they did a silly thing. e.g. The references Speckl resource is no longer available; refreshing to start the process again."

Very often, DUI handles this with no problem. For example, if the renamed branch is selected, the UI will switch back to `main` and update the list to include the name.

teocomi:
Closing for now, hopefully handled by https://github.com/specklesystems/speckle-sharp/pull/2495

PR #2495, opened 2023-04-27
fix(dui): adds some try catches

teocomi:
Adds some safety guards in DUI, potentially fixes: #2281

### DIFF
--- a/DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
+++ b/DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
@@ -161,7 +161,7 @@ public class CollaboratorsViewModel : ReactiveObject, IRoutableViewModel
     }
     catch (Exception ex)
     {
-      SpeckleLog.Logger.Warning(
+      SpeckleLog.Logger.Error(
         ex,
         "Swallowing exception in {methodName}: {exceptionMessage}",
         nameof(CanSaveCommand),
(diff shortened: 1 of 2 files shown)


DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
```diff
@@ -134,189 +134,243 @@ private async void SearchUsers()
...
-        return true;
+    catch (Exception ex)
+    {
+      SpeckleLog.Logger.Warning(
+        ex,
+        "Swallowing exception in {methodName}: {exceptionMessage}",
+        nameof(CanSaveCommand),
+        ex.Message
+      );
+    }
```
AlanRynne, line 170: Shouldn't this be a `Fatal` log, since it would indicate something unexpected went wrong that prevented us from fulfilling the request.

DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
```diff
@@ -134,189 +134,243 @@ private async void SearchUsers()
...
-        .ConfigureAwait(true);
-      _stream.Stream.pendingCollaborators = pc.pendingCollaborators;
-      _stream.StreamState.CachedStream = _stream.Stream;
-
-      ReloadUsers();
+      if (IsDialog)
+        MainViewModel.RouterInstance.NavigateBack.Execute();
     }
+
     catch (Exception ex)
```
AlanRynne, line 374: Same here, maybe we should make this one `Fatal`

DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
```diff
@@ -134,189 +134,243 @@ private async void SearchUsers()
...
     {
       SpeckleLog.Logger.Warning(
```
AlanRynne, line 376: Same here, maybe we should make this one `Fatal`

AlanRynne (commented):
This looks good! Just once question regarding Logging

PR #2504, opened 2023-05-01
ca1062 supress

JR-Morgan:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062

CA1062 does not play nicely with nullability syntax, I suggest we revert to it being disabled.


AlanRynne (approved):
Perfect

PR #2505, opened 2023-05-02
feat(gh): First implementation of Flatten Collection component

AlanRynne:
Adds a new "Flatten Collection" component to Grasshopper connector.

This component will accept a Speckle collection, and return a flat list of all speckle collections included within it (including itself).

To prevent any side-effects, the output collections are new objects that contain
- The elements inside that collection (but no inner collections)
- The original name of the collection and its type
- A dynamic property called "fullName" that contains all the names of all the parent collections and itself in Rhino style ("::" separator)

Additionally, a new GH_Goo data type has been created called GH_SpeckleCollection, and conversions between GH_SpeckleBase and GH_SpeckleCollection have been implemented to allow for conversion casting both ways.

Also added temporary icons for both GH_SpeckleCollection and "Flatten Collection" component

Screenshot 2023-05-02 at 08 44 17


teocomi (changes requested):
I'm getting some errors with a commit coming from Revit; maybe the data structure changed since this was written?

`1. Data conversion failed from Speckle Model to GH_SpeckleCollection`


Test commit: https://latest.speckle.dev/streams/ff85c554cc/commits/714d6e0a81
image

AlanRynne:
This is due to the fact that `Model` does not currently inherit from `Collection`, and my logic in GH checks specifically to verify if an object is a `Collection` class, so this will fail and it is currently "by design".

Weirdly, your commit `Model` does contain all the info as if it were a collection (collectionType, name, elements list)

What I have improved is the text on the conversion error to be more understandable by a normal user.

image

AlanRynne:
Scratch what I said, I was missing a commit from `dev` and a clean rebuild
Screenshot 2023-05-10 at 11 36 14

teocomi (approved):

PR #2508, opened 2023-05-03
Log: Ensured log templates are all compile time constants

JR-Morgan:


AlanRynne (approved):
Browsed through the changes and they all seem legit 🙌🏼

Nice work on this Jedd 🙌🏼

PR #2510, opened 2023-05-04
feat(revit): Implemented sending collections

JR-Morgan:
see #2187

# Functional Changes (all regarding sending)

- Revit `Model` class (root commit object) now inherits from `Collection`
- Revit commit object will send `Collection` objects by category, instead of the old prop name (i.e. `@Walls` `@Doors`)
- `Network` objects will now be nested under their host (e.g. lighting fixture will be nested under its ceiling object), previously we were ignoring this hosting relationship.
- Enforcing that every ToSpeckle conversion result will have its application id set. (means that both `Groups` and `Networks` will have their app ids set). NOTE: I'm not sure the implications of this change, and we can reconsider it.

# Code changes (all regarding sending)

- Introduced an abstract `CommitObjectBuilder<T>` in Core.
- Introduced a `RevitCommitObjectBuilder` class in the revit connector (inheriting from ^)
- Revit hosting is now done by querying the child elements for their host, not the host for their children (more reliable)
- Revit elements now get converted one by one, rather than the previous behaviour of host converting its children if the host is converted first. This change removes the need for the `SpeckleHost` dummy objects.

# Things that need testing
- Are all types of hosting working as expected?
- Are network elements working properly?
- Does the new Collection structure receive in other connectors properly?


teocomi:
I think it works quite great!

Tested with this model containing a few edge cases: 
[Nested-instance test.zip](https://github.com/specklesystems/speckle-sharp/files/11436131/Nested-instance.test.zip)

Output: https://latest.speckle.dev/streams/ff85c554cc/commits/96c9852ae3

PR #2511, opened 2023-05-04
feat(core): adds an extension method to get a collection's elements i…

teocomi:
Adds an extension method to get a collection's elements in a backwards compatible way.

Quick attempt, suggestions welcome!


AlanRynne (approved):
Looks good!

I'm thinking of moving my "Flatten Collection" logic to core to now that I see this

PR #2533, opened 2023-05-10
fix(Revit): Fixed regression of Networks

JR-Morgan:
#2510 introduced a regression that meant that `Network` objects were being hosted under the host of their network elements.

@clairekuang and I agreed that this is not the behaviour that we want, since our `Network` class was not designed with hosting in mind.

This PR aligns the new commit building with the older behaviour of having all `Network` objects nested under their category.

Todo:
- We also mentioned the possibility that some objects (e.g. Lighting fixtures) that are currently being converted as `Network`s, perhaps should simply be converted as a `RevitFamilyInstance` and could support hosting.

Additionally, `Views`, `Materials`, and `ProjectInformation` will be nested directly under the model collection, under unique props, rather than as a collection


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertCeiling.cs
```diff
@@ -28,7 +28,7 @@ private RevitCeiling CeilingToSpeckle(DB.Ceiling revitCeiling, out List<string>
       GetAllRevitParamsAndIds(speckleCeiling, revitCeiling, new List<string> { "LEVEL_PARAM" });
 
       GetHostedElements(speckleCeiling, revitCeiling, out List<string> hostedNotes);
-      if (hostedNotes.Any()) notes.AddRange(hostedNotes);
+      if (hostedNotes.Any()) notes.AddRange(hostedNotes); //TODO: what are we doing here?
```
AlanRynne, line 31: I'm also having the same question right now...

AlanRynne (approved):
Haven't actually tested it, but changes look good to me.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertCeiling.cs
clairekuang: This was supposed to be used for reporting in DUI, if retrieving hosted elements went wrong for some reason

PR #2550, opened 2023-05-12
Jrm/core/nan deserializer

JR-Morgan:
Added support for deserialziation of json String -> Double for "Nan", "Infinity" and "-Infinity" values


teocomi (approved):
Txs @JR-Morgan !!

PR #2560, opened 2023-05-15
revit(fix): mapping levels & types

JR-Morgan:


clairekuang (approved):
added some refactor notes, but tested and it works!

PR #2564, opened 2023-05-16
Core/update deps2

teocomi:
Updates avalonia, sentry, sqlite to their next minor versions and gql to next major (minimal changes needed).
Had to tweak slightly GQL bc of breaking changes: https://github.com/graphql-dotnet/graphql-client/releases


JR-Morgan (approved):

teocomi:
As this does not directly fix the gql issue, we can wait to merge it in 2.15...

PR #2576, opened 2023-05-17
bug(rhino): temporarily disables doc layer change event handler for performance

### LINKED ISSUE #2570: Rhino: Layer updated event should only be registered when `Layer` filter is active
HUGE PERFORMANCE HIT, and will likely hit Rate Limits on the server if there are too many layers on the commit, which would cause it to be even slower, as it will retry and fail multiple times.

This line in CollectionToNative causes a LayerChanged event in Rhino, which in turns causes a refresh of the Stream page, which does a QGL query to the server to keep branches up to date.

This happens once per layer, meaning that if we have 10000 layers, the stream page will be updated 10000 times even if the current view is the main page (and hence, has no need to update the branch list of the stream at all)

image

## Solution

The `LayerChanged` event should only be activated whenever the `Layer filter view` is active.
`Layer filter view` should refresh the layers as soon as it becomes active, setup the event to keep the list up to date and finally disable the event when it's hidden (user changes filter or closes stream page)

I'm assigning @teocomi to this one too bc I'm unaware of the consequences of not calling this update stream method on the binding level.

clairekuang:
## Description & motivation

Disables the doc layer change event handler on layer creation during receive operations, as this was causing a huge performance issue when receiving commits that result in many layers being created.

Not the ideal solution, but should fix #2570

## Changes:

- rhino connector

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
@@ -44,7 +44,7 @@ public class ConnectorBindingsRhino : ConnectorBindings
   public ConnectorBindingsRhino()
   {
     RhinoDoc.EndOpenDocument += RhinoDoc_EndOpenDocument;
-    RhinoDoc.LayerTableEvent += RhinoDoc_LayerChange;
+    RhinoDoc.LayerTableEvent += RhinoDoc_LayerChange; // used to update the DUI2 layer filter to reflect layer changes
   }
 
   public RhinoDoc Doc => RhinoDoc.ActiveDoc;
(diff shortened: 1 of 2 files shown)


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.cs
```diff
@@ -602,6 +604,7 @@ public override async Task<StreamState> ReceiveStream(StreamState state, Progres
 
             layers.Add(layer.FullPath, layer);
           }
+          RhinoDoc.LayerTableEvent += RhinoDoc_LayerChange;
           #endregion
```
AlanRynne, line 608: Just wondering if we should delay adding this event even further down the method to prevent possible mishaps if a user modifies a layer while baking?

AlanRynne (approved):
This is looking good!

Not the most SOLID solution, but that would require a massive refactor we're not up to doing right now. :)

I just have a single comment 👇🏻 which doesn't really block merging if you don't deem it important, so I'm approving this already.

DesktopUI2/DesktopUI2/ViewModels/HomeViewModel.cs
```diff
@@ -114,7 +115,7 @@ internal void UpdateSelectedStream()
   {
     try
     {
-      if (_selectedSavedStream != null)
```
teocomi, line 117: Took the freedom to add this @clairekuang @AlanRynne - we do not need to update a Stream while it's currently sending or receiving.

Maybe a debouncer would be needed on the event-triggering side too.

PR #2587, opened 2023-05-18
fix(archicad): Fixed issue where logger was used before initialising

JR-Morgan:
I noticed while testing that the archicad connector was logging to console/file/seq under "Coreunknown"

After some investigation, it seems that the kit loading done by `App` uses the logger before its initialised by DUI2.

This fix adds an extra log in the entry point of the Archicad connector to ensure Setup is initialised before DUI2.
This is similar to how we do things in Rhino and Revit connectors.

@AlanRynne please could you review the changes made in Core
@jozseflkiss , I didn't figure out how to run the archicad connector locally, if you are available tommorow to give this a quick test for me. 
Simply running the connector from this branch. I expect the first console line to read "Archicad26" and not "Coreunknown"
image

### DIFF
--- a/ConnectorArchicad/ConnectorArchicad/Program.cs
+++ b/ConnectorArchicad/ConnectorArchicad/Program.cs
@@ -6,6 +6,7 @@ using Avalonia.Controls;
 using Avalonia.ReactiveUI;
 using DesktopUI2.ViewModels;
 using DesktopUI2.Views;
+using Speckle.Core.Logging;
 
 namespace Archicad.Launcher
 {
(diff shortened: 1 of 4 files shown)


jozseflkiss:
The log seems to be ok:
`[20:35:26 INF] Initialized logger inside ArchicadArchicad 26/2.0.999-local/2.0.999.0000 for user `

AlanRynne (approved):
♥️ This looks legit, pending @jozseflkiss confirmation that it works as intended.

PR #2590, opened 2023-05-19
Revert "Rhino/mapper tweaks"

AlanRynne:
Reverts specklesystems/speckle-sharp#2586

Fixes an incorrect revert with Revit issues.


AlanRynne:
Merging in as this is being Live verified by @teocomi and @clairekuang via video call

PR #2596, opened 2023-05-22
Refactor(Revit): Abstract previously received objects logic into interface

connorivy:
## Description & motivation

Currently we have some tight coupling of DUI, the Revit connector, and the Revit converter. This makes it difficult to make changes to the connector behavior without having adverse effects on other parts of the system. A prime example of this tight coupling is the way that we keep track of previously received objects.

The way that we currently keep track of previously received objects is by a list that is stored on a DUI object, converted to a dictionary, read by the converter, and then written by the connector. All of these projects access the direct data storage of the previously received objects which is a violation of the dependency inversion principle, and ultimately it means that making changes to the way we cache previous recieved objects will result in changes to either hacky fixes or changes to DUI, the connector, and the converter.

This PR looks to abstract all functionality dealing with previously received objects into an interface that is accessible by both the connector and the converter. For the time being, this object is still being stored on the StreamState, but this is only to maintain the current functionality we have now. Soon we will completely remove any DUI objects from this process.

## Changes:

- Added new netstandard 2.0 project as a dependency of both the connector and converter
  - Currently this project only holds the IReceivedObjectsCache and the IConvertedObjectsCache interfaces. I think it will eventually need to have a IRevitConversion interface for when we split the converter into multiple classes, but mostly likely not much more than that.
  - Open to other suggestions if we don't want to add a new project, but to me this seems much cleaner than the way we are currently doing things
- Created IReceivedObjectsCache cache
  - This interface exposes the essential functionality of a cache of previously received objects. When we want to change to storing this information on a per-object scope instead of a global scope, all we need to do is implement this interface on the new storage class
  - Removed the PreviousContextObjects dictionary in the converter and replaced it with an object of this interface's type
- Created IConvertedObjectsCache 
  - This interface is used to expose the elements in the converted objects cache to allow the IReceivedObjectsCache to add the newly converted elements.
  - We could eventually pass this to the converter instead of using the confusing "ContextObjects" property that is a list of all elements to be converted except that we sometimes remove objects that we convert. This would involve building in some thread safety though.
- Create the StreamStateCache object as the current implementation of IReceivedObjectsCache which implements our current previously received objects logic

## Validation of changes:

-tests are passing to the same extent they are passing on dev

## Checklist:


Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -127,14 +134,14 @@ public void SetContextObjects(List<ApplicationObject> objects)
 
     public void SetPreviousContextObjects(List<ApplicationObject> objects)
```
teocomi, line 135: Are `PreviousContextObjects ` not needed anymore?

ConnectorRevit/ConnectorRevit/Storage/ConvertedObjectsCache.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+      convertedObjects[@base.applicationId] = (@base, elements);
+    }
+    public IEnumerable<string> GetApplicationIds()
+    {
+      return convertedObjects.Keys;
+    }
+    public IEnumerable<Base> GetConvertedBaseObjects()
+    {
+      foreach (var tuple in convertedObjects.Values)
```
teocomi, line 20: Can we just use a Linq Select?

ConnectorRevit/ConnectorRevit/Storage/StreamStateCache.cs
```diff
@@ -0,0 +1,86 @@
...
+      }
+
+      //element was not cached in a PreviousContex but might exist in the model
+      //eg: user sends some objects, moves them, receives them 
+      yield return doc.GetElement(applicationId);
+    }
+
+    public void RemoveSpeckleId(string applicationId)
+    {
+      //previousContextObjects.Remove(applicationId);
```
teocomi, line 83: Do you think this should be implemented?

teocomi (approved):

Issue #2612, opened 2023-05-29
Dynamo: 2.14.0 seems to be broken


AlanRynne:
Dynamo Connector in 2.14.0 is broken:

- Nodes that have custom UI can't be placed in the canvas (send, receive, account list)
- All nodes are missing their icons

PR #2613, opened 2023-05-29
fix(dynamo): broken project after build step

teocomi:
Fixes: https://github.com/specklesystems/speckle-sharp/issues/2612

Not sure why these after-build steps were modified during the SDK project migration, but that broke the Connector.
This PR restores the after-build step to what it was before.


AlanRynne (approved):
Tested locally on my machine and seems to have solved the issue:

Icons appear
Custom ui components work

PR #2620, opened 2023-05-31
feat(dyn): Changes resource building step so it works with dotnet build

AlanRynne:
Modifies Dynamo project so that the resources file is compiled with CSC instead of AL.

This is the recomended way in `dotnet core`


AlanRynne:
Thx for checking! I was sure I had that one on my clean local build... let me check if it's not being copied over properly.

teocomi (approved):

PR #2624, opened 2023-06-01
chore(core): updates sentry, avalonia and other packages minor verisons

teocomi:
updates sentry, avalonia and other packages minor verisons


teocomi:
Alright, finally all passing 🎉

AlanRynne (approved):
Nice!

Merging in so we can have the second half of 2.15 dev cycle to catch any potential issues.

PR #2625, opened 2023-06-01
feat(dui): check if stream exists

teocomi:
- adds a method in core to check if a stream exists
- in DUI, checks if a stream exists before opening its card
- 
I tagged you Jedd as it couches Core, but it should be pretty straightforward.


Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
```diff
@@ -9,6 +9,48 @@ namespace Speckle.Core.Api;
...
+  {
+    return StreamExists(CancellationToken.None, id);
+  }
+
+  /// <summary>
+  /// Cheks if a stream exists by id.
+  /// </summary>
+  /// <param name="id">Id of the stream to get</param>
+  /// <returns></returns>
+  public async Task<bool> StreamExists(CancellationToken cancellationToken, string id)
```
JR-Morgan, line 28: The method signature you have done does match what we use currently.

But I would like us to consider switching to the more C# recommended system of having the cancellation token as
an optional parameter, instead of having a overload function that uses CancellationToken.None

```cs
public async Task<bool> StreamExists(string id, CancellationToken cancellationToken = default)
```

JR-Morgan: I've been waiting for an opportunity to switch our current functions over to use this. so address some compiler warnings that it generates + reduce the boilerplate in the Client class.

Potentially we want to make new API functions comply?
But there is an argument to keep them consistant

Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
```diff
@@ -9,6 +9,48 @@ namespace Speckle.Core.Api;
...
+                    }}",
+        Variables = new { id }
+      };
+      var stream = (await ExecuteGraphQLRequest<StreamData>(request, cancellationToken).ConfigureAwait(false)).stream;
+
+      return stream.id == id;
+    }
+    catch
+    {
+      //ignore
```
JR-Morgan, line 48: Do we need to swallow exceptions here?
The other client functions let Exceptions bubble up, no?

JR-Morgan (changes requested):

Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
JR-Morgan: This is a only a recommendation from .NET, not a strictly necessarily change 
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068 

So happy to ignore this for now if you'd like

Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
teocomi: makes sense, I gave it a stab!

Issue #2382, opened 2023-04-03
feat(objects/core): move `Collection` class to Core


clairekuang:
Currently the `Collection:Base` class is in `Objects`, but we should migrate it to `Core` due to:
- traversing the base commit in connectors on receive
- handling the data structure in Grasshopper, as per any organizational class

This would be a prerequisite for:
- refactoring the Rhino connector on send and receive
- unblocking implementation in SpecklePy

Note: serialization and deserialization testing would be needed if these Collections are stored in the streamstate

PR #2626, opened 2023-06-01
hotfix(collections): Deprecated object models for backwards compatibility.

JR-Morgan:
To allow us to have backwards compatibility saving object models, as targets for deserialization, I have changed the way the deserialiser looks for types to also check for a ".Deprecated" namespace that otherwise matches an incoming speckle type.

This provides us a convenient way to change/move the namespace/type name of our object models, without having to change converters in anyway.

This was motivated by a change made by #2382 that breaks older commits from 2.13

### DIFF
--- a/Core/Core/Serialisation/BaseObjectSerialzerUtilities.cs
+++ b/Core/Core/Serialisation/BaseObjectSerialzerUtilities.cs
@@ -217,9 +217,23 @@ internal static class SerializationUtilities
       var type = KitManager.Types.FirstOrDefault(tp => tp.FullName == typeName);
       if (type != null)
         return type;
+
+      //To allow for backwards compatibility saving deserialization target types.
+      //We also check a ".Deprecated" prefixed namespace
+      string deprecatedTypeName = GetDeprecatedTypeName(typeName);
+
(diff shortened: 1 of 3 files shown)


Core/Core/Serialisation/BaseObjectSerialzerUtilities.cs
```diff
@@ -217,11 +217,25 @@ internal static Type GetAtomicType(string objFullType)
...
+
+      var deprecatedType = KitManager.Types.FirstOrDefault(tp => tp.FullName == deprecatedTypeName);
+      if (deprecatedType != null)
+        return deprecatedType;
     }
 
     return typeof(Base);
   }
 
+  internal static string GetDeprecatedTypeName(string typeName, string deprecatedSubstring = "Deprecated.")
```
AlanRynne, line 233: Is it `.Deprecated` or `Deprecated.`? The comment above seems to be conflicting with the default value

AlanRynne (commented):
Looks good but I do have a couple questions. A quick call will be faster! :D

clairekuang (approved):
Tested on older 2.13 commits, looks good to me 👍

PR #2634, opened 2023-06-02
feat(Archicad): Time profiler

jozseflkiss:
## Description & motivation

Long-term goal is to ensure reached performance during development (regression check). As a first step time measurements has been done in this PR.

## Changes:

Archicad.Helpers.Timer introduced, it creates a scope where time measurements are used.
In the Core: SerilogTimings nuget package is used to measure times and to do the reporting to Serilog. Additional CumulativeTimer class is added (based on the implementation pattern of SerilogTimings, but timer values are cumulatively collected. At the end EnrichSerilogOperation is invoked to pass the sub timing to the main SerilogTiming entry.

For passing the context to the measurement point across deep stack calls, a State class is introduced in Speckle.Core.Helpers. It uses global static variables as context (thread locals variable unfortunately are not enough because the conversion runs in multiple threads).

Usage:
```
using (var timer = Archicad.Helpers.Timer.CreateReceive(state.StreamId))
{
  ...
      var context = Archicad.Helpers.Timer.Context.Peek;
      using (context?.cumulativeTimer?.Begin(ConnectorArchicad.Properties.OperationNameTemplates.ConvertToNative, Type.Name))
    {
      ...
    }

}
```

## Screenshots:

image

## Checklist:

- [x] I have updated or added relevant documentation.


Core/Core/Helpers/State.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+
+namespace Speckle.Core.Helpers;
+
+public class State<T> : IDisposable where T : State<T>, new()
+{
+  static T root;
```
AlanRynne, line 7: I'd love to get more info on why `ThreadStatic` wasn't working for your use-case.

AlanRynne (approved):
This is looking good! I just have one question about the `State<T>` class, not really a blocker 👇🏼 See below

Core/Core/Helpers/State.cs
jozseflkiss: I removed the `ThreadStatic` and use static variables here, because the conversion runs on multiple threads thanks to `async` functions. With `ThreadStatic` the first Peek call in a specific thread would return a new context.

AlanRynne: Ok this sounds legit, I suspected as much but I hadn't actually done much testing in this regard 🙌🏼

I'm ok to merge this in 🚀

PR #2646, opened 2023-06-07
Feat(Revit) : load missing types on receive

connorivy:
## Changes:

- added elementTypeMapper class whose job is to either create or update a type map between incoming "types" and element types that exist in the project
  - The TypeMap's job is just to store the data
- added familyImporter class whose job is to walk the user through importing family symbols into the active project
- created interfaces IHostTypeAsStringContainer and ITypeMap that get passed to the DUI viewModel and allow DUI to access the data that is in the document
- created IRevitElementTypeRetriever and implemented it on the converter. This allows the connector to pass objects to the converter and allows the converter (who knows about Objects) to query the objects for their types. It also exposes a couple element type retreival methods that cache the retrieved values in a new cache class
- create ElementTypeInfo class to consolidate a bit of information about different categories and IElementTypeInfoExposer to let the converter expose that category info to the mapper (in the connector)
- created new missingIncomingType dialog that pops up by default when there are incoming types that don't exist in the project
- created conversionCache object to store elementTypes (and eventually other things) during mapping so we can retrieve them faster during conversion
- deleted a lot of old code that did the same thing but worse

## Validation of changes:

manual sending and receiving. The tests project is no longer in the speckle sharp repo so we'll definitely need to figure out what to do about that and run / create some tests

## Checklist:


ConnectorRevit/ConnectorRevit/TypeMapping/FamilyImporter.cs
```diff
@@ -0,0 +1,236 @@
...
+      //close current dialog body
+      MainViewModel.CloseDialog();
+    }
+
+    private static void AddSymbolToAllSymbols(Dictionary<string, List<Symbol>> allSymbols, XmlDocument xmlDoc, XmlNamespaceManager nsman, string familyName, IEnumerable<ElementType> elementTypes)
+    {
+      var familyRoot = xmlDoc.GetElementsByTagName("A:family");
+      if (familyRoot.Count != 1)
+      {
+        // TODO: logging
```
teocomi, line 168: let's log it before we forget :)

DesktopUI2/DesktopUI2/Models/TypeMappingOnReceive/MappingValue.cs
```diff
@@ -0,0 +1,48 @@
...
+      set => this.RaiseAndSetIfChanged(ref _initialGuess, value);
+    }
+
+    [DataMember]
+    public string OutgoingType
+    {
+      get => _outgoingType;
+      set => this.RaiseAndSetIfChanged(ref _outgoingType, value);
+    }
+
```
teocomi, line 41: Is this not being serialized on purpose?

teocomi:
As discussed on discord, we need to ensure that loadable family instances are supported too. Since the family name is missing from the UI it could be confusing for the user to know what is missing and what to map to (some families of the same category would have the same type name eg window1-50x100cm, window2-50x100cm).

image

PR #2650, opened 2023-06-08
Feat(Revit): send cropped meshes from view filter

connorivy:
## Changes:

- if the user sends 1 selected view, then pass that view into the converter. The converter will then save that view in a ViewSpecificOptions property
- if the ViewSpecificOptions property is not null then it trumps any option value that may have been passed into the GetElementDisplayValue method. This ensures the view specific geometry will be sent
- Changes to CurtainWall mesh retrieval
  - Because a wall retrieves it's own curtain wall panels and mullions, there is no guarentee that those panels are actually present in the view. This will cause them to be sent as normal when they shouldn't be present at all. Here is an example of this behavior
  
image

  - Therefore we need to actually check if each panel is present in the view (if the ViewSpecificOptions isn't null)

## Screenshots:

From Revit
image

Sending 3D view before this PR
image

Sending 3D view after this PR
image

## Checklist:

### DIFF
--- a/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
+++ b/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
@@ -5,6 +5,7 @@ using Autodesk.Revit.DB;
 using ConnectorRevit;
 using DesktopUI2.Models.Filters;
 using DesktopUI2.Models.Settings;
+using Speckle.Core.Kits;
 using Speckle.Core.Logging;
 
 namespace Speckle.ConnectorRevit.UI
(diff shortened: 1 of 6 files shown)


ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
```diff
@@ -243,38 +244,37 @@ private List<Element> GetSelectionFilterObjects(ISelectionFilter filter)
...
+                .Where(x => viewFilter.Selection.Contains(x.Name))
+                .ToList();
 
-            var views = new FilteredElementCollector(currentDoc)
-              .WhereElementIsNotElementType()
-              .OfClass(typeof(View))
-              .Where(x => viewFilter.Selection.Contains(x.Name));
+              if (views.Count == 1)
+              {
+                converter.SetContextDocument(views[0]);
```
teocomi, line 258: Could you please just add a note here that we're using this to crop the objects according to the view?

teocomi (approved):

PR #2651, opened 2023-06-08
feat(core): Traversal optimisation by avoiding unnecessary traversal of dynamic properties

JR-Morgan:
While creating the traversal documentation, I noticed that there is one source of inconsistencies between some of our traversal functions.
And it raises a great question that potentially could simplify one of the rules.

Can anyone think, are there any Object models that
1. `CanConvertToNative` returns `true` for,
2. AND we expect dynamic properties to be traversed,

IF NOT, then we can 
1. increse travesral performance by avoiding unnessesry traversal!
2. Align better the rhino/revit traversal rules with Archicad/Blender rules which already have this behaviour (with no reported problems)

### DIFF
--- a/Core/Core/Models/GraphTraversal/DefaultTraversal.cs
+++ b/Core/Core/Models/GraphTraversal/DefaultTraversal.cs
@@ -21,20 +21,9 @@ public static class DefaultTraversal
       .NewTraversalRule()
       .When(converter.CanConvertToNative)
       .When(HasDisplayValue)
-      .ContinueTraversing(b =>
-      {
-        var membersToTraverse = b.GetDynamicMembers().Concat(elementsAliases).Except(ignoreProps);
-        return membersToTraverse;
-      });
+      .ContinueTraversing(ElementsAliases);
 
-    var ignoreResultsRule = TraversalRule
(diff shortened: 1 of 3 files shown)


Core/Core/Models/GraphTraversal/DefaultTraversal.cs
```diff
@@ -86,35 +57,28 @@ public static GraphTraversal CreateBIMTraverseFunc(ISpeckleConverter converter)
...
-
   #region Helper Functions
 
-  internal static readonly string[] elementsAliases = { "elements", "@elements" };
-
-  internal static IEnumerable<string> ElementsAliases(Base _)
-  {
-    return elementsAliases;
-  }
+  //WORKAROUND: ideally, traversal rules would not have Objects specific rules.
```
AlanRynne, line 66: Agreed! We can live with this for now, but if some Objects need to be ignored we may want to create a specific class attribute for it.

I'll open a separate issue with this.

AlanRynne (approved):
So much cleaner!

This is great @JR-Morgan. I added a comment on the `Objects`-specific rule, but will be tracked separately so I'm approving this!

PR #2656, opened 2023-06-12
feat(rvt): Refactored categories logic into shared project

### LINKED ISSUE #2442: Better way to handle Revit categories
Revit categories are handled in our connector and converter in a bit of a messy way, this is partly due to their separation of context.
Here are some use cases:

1. a user in Grasshopper BIM wants to specify the category input of some nodes, and should be able to pick it from a dropdown, not have to type it
2. the DUI interface needs to expose a list of categories for the user to select to filter elements to send
3. we're also using an internally defined list of categories to filter out elements not supported
4. we need to match user friendly strings (like the ones passed by Grasshopper BIM) with actual built-in names

As a consequence, we have the following mess:
- List of categories in the connector for (2) https://github.com/specklesystems/speckle-sharp/blob/main/ConnectorRevit/ConnectorRevit/ConnectorRevitUtils.cs#L264
- List of categories in the converter for (3) https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/Categories.cs#L44
- Enum in Objects for (1) (serialized as `Int`!!) https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Objects/BuiltElements/Revit/Enums.cs#L7
- Another list of categories for (4) https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Converters/ConverterRevit/ConverterRevitShared/Categories.cs#L158

The big problem is that the category enum used by the schema builder is currently serialized to `int` which means that every change is breaking.

A solution is to use a dictionary structure to store all this information in one place, ideally inside Objects/Converter - but a question remains on how to move this from the converter to the connector. A solution would be to add an additional binding.

AlanRynne:
Fixes #2442 

Removes duplication of categories lists:
- Between Converter and Connector
- Between Objects and Converter

All shared logic has been moved into `RevitSharedResources`
The shared project has also been split to allow for specific Revit version support.


teocomi (approved):
Looks good to me!

I tested:
- GH > Revit Directshapes
- Revit > Speckle filters

All seemed to work well.

Next time we should make a separate PR with style changes to make review simpler.

One thing to watch for are the nugets, will the RevitSharedResources dll be included there automatically?

AlanRynne:
> One thing to watch for are the nugets, will the RevitSharedResources dll be included there automatically?

I think the best thing to do here would be to also release it as a NuGet, so that dependent nugets get linked by the build/pack process

I'll double check this can be done safely and reliably, and that it's set up to do so 🙌🏼

Our nuget pack CI step builds `dotnet pack All.sln` so any project that has been flagged with `IsPackable` will have it's nuget pushed.

Maybe one thing to take into account before hitting a release here is going through the necessary nuget info for this new package, ensuring its all correct.

Issue #2580, opened 2023-05-18
feat(civil3d): add profile to speckle conversions


clairekuang:
## What package are you referring to?
 Civil3D connector

## Describe the solution you'd like

Add Profile To Speckle and Profile to Native conversions, taking into consideration:

- [x] sending individual profiles
- [x] sending profiles as part of featurelines and corridors
- [x] re-evaluating the current profile class

## Additional context

Requested by community members (Jacobs civil engineering)

teocomi:
@clairekuang please link the matching PR and close if done :)

PR #2660, opened 2023-06-12
feat(civil3d): adds `CivilProfile:Profile` class and updates ProfileToSpeckle conversion

clairekuang:
## Description & motivation
This pr is part of #2580 and includes:
- a new `CivilProfile` class that has civil3d specific properties
- updated `ProfileToSpeckle` conversion to use the new Civil class
- small refactor of CivilDoc handling and retrieving
- adds profile conversions to alignments

## Changes:

- Civil3D converter


AlanRynne (approved):
Looks good, and worked on my end send/receive from civil. At least to the best of my knowledge :D no errors were reported on each operation.

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -505,6 +542,97 @@ private Arc ProfileArcToSpeckle(CivilDB.ProfileCircular circular)
       return LineToSpeckle(new LineSegment2d(start, end));
     }
 
+    /*
```
AlanRynne, line 545: Do we need this huge comment chunk here though? :) If it can be removed it would be nice!

I'll leave the decision to you though

PR #2666, opened 2023-06-13
fix(Navisworks): Division by zero errors on small models. Fixes #2665

### LINKED ISSUE #2665: bug(Navisworks): Small models (with few objects) may trigger Divide by zero errors
To reduce the UI drag of the Progress Bar running on the main thread, we use an interval that is a constant of 1% of the total objects for conversion or traversal. Small models containing fewer than 100 elements were triggering divided by zero errors on the Modulus check for updating progress.

jsdbroughton:
Fixes #2665.

The code changes prevent division by zero errors in three different files. The changes ensure that the progress bar is updated correctly even when there are no items to process or fewer than the anticipated interval, resulting in a more accurate representation of progress.

Code changes wrap the interval assignment to a default value where the total is less than the default interval chunk. For increment values, the divisor is zero-checked.

This is an alternative to checking on the Progress update loop. However, that would be more DRY.


AlanRynne (approved):
Interesting.... would love to know more about this on a quick chat! Mostly out of curiosity on why it happens :D progress bars are hard 😓

PR #2667, opened 2023-06-13
fix(Navisworks): Reset ConverterContext on each send. Fixes #2430

### LINKED ISSUE #2430: regression(Navisworks): Basepoint transforms not applying
Unclear when this stopped working. 

This is a key interoperability feature for host applications unable to handle real-world coordinates.

jsdbroughton:
The previous code set Converter on Init and Document change did not reflect the settings change for Basepoint options.

Change to call SetConverterContext on each send.


AlanRynne (approved):
Nice! Not sure if it's needed on every send but in Grasshopper we also do this for safety (never assume you have the latest doc when an operation starts seems like a safe assumption) 🙂

PR #2669, opened 2023-06-13
Feat(Revit) : make receive operations testable and make tests pass for new family instance tests files

connorivy:
## Description & motivation

This PR seeks to make our integration tests reuse a lot more code that our connectors use. It changes both the Recieve and Send command to just call a different method where all of the connector logic lives because the new method is testable.

##### SendStreamTestable

- changed the method that is called by DUI, SendStream(streamState, progressViewModel) to only contain a single line which calls the new method, SendStreamTestable
- SendStreamTestable takes the streamstate and progressVM as well as a Type (converterType), UIDocument, and an ISpeckleObjectSender
- The ISpeckleObjectSender faciliates testing of this method.  When called by a typical user, this object will send the user's selection to the server associated with their account, as before. When this method is called by the integration tests, the ISpeckleObjectSender will be an object that just sends the commit to a local transport.
- Made several changes to the selection logic that make it static and accessible by the converterRevitTests project.

##### ReceiveStreamTestable

- Added RecieveStreamTestable which returns an IConvertedObjectsCache and takes a StreamState, ProgressViewModel, Type (converterType), UIDocument, and an ISpeckleObjectReceiver
  - The ISpeckleObjectReceiver either receives from a server, or if running tests, it will receive from a local transport.
  - Many global objects were being used by this method. CurrentSettings, Preview, and StoredObjects. I removed these in order to make this method static in order to be used by the tests.
  - Many of the changes to ConnectorBindingsRevit.Preview and Mapping are related to these class-level properties no-longer existing.

##### ConvertedObjectsCache in Converter

-previously we were populating a cache of converted objects in the connector after each Base object returned from the converter. The issue with that was that some objects are doing their own conversions in the converter. 
- an example of the above is that walls are responsible from converting their children. Therefore the wall will get passed to the converter, the wall will get converted, all of it's children will get converted (doors, windows, ect.), and then the converter will return the wall's conversion result. Only the wall will get cached which means that nested elements will not be able to be updated.
- While we potentially want to remove the need to host elements under an "elements" prop of the host in the future, for now the fix is to have the convertToNative method add the converted objects to the cache after the conversion finishes. This way every item gets added to the cache.
- The "familyInstance" test in particular now shows the proper updating of nested elements

The rest of the changes were either conversion changes that I needed to make so that tests would pass, or changes to the test project in order to properly validate the results of testing using the connector logic 

## Validation of changes:

All 56 integration tests pass for revit 2021
Added a few tests for family instances in 2022. To test them yourself, download them from [my google drive](https://drive.google.com/drive/folders/1rp5CWJPaDecjP91SUM5o1iUpTE4y_z99?usp=sharing) and then save them in the testModels/2022/familyInstance folder.

## Checklist:


teocomi (commented):
Let's discuss this when you're back!

Issue #2668, opened 2023-06-13
feat(Navisworks): A Resend last commit Action


jsdbroughton:
Issue #2360 is longstanding and related to enormous model sends. The problem is resolved when sending the same data again as more chunks get through before the magical cutoff.

However, long sends come from long conversions, and it would be helpful to retry just the send part.

If the Connector caches the last conversion and commit-object, then a call could be made outside of the normal Send UI, which could reinitiate the Send.  This feature proposes an additional Action on the Ribbon. It's hacky, but a convenient addition.

Successful Sends should Dispose of and not cache the conversion. Likewise, a new Send or DocumentChange should also Dispose of the cached objects.

### Alternatives considered:
If it is possible to highlight the exact cause of Sends being cancelled and returning null by caching and parsing the Logs sent from the Connector, then on matching the fail pattern, a Send retry could be initiated from the original send command.

PR #2670, opened 2023-06-13
feat(Navisworks): Retry Sending already converted objects. Implements #2668

jsdbroughton:
## This is a highly speculative but functional adaptation. If long sends become possible from the server, this may be moot. I am open to rejections on this.

This PR adds a new Ribbon button that resends a commit that has failed to send to completion post successful Conversion. Motivation is centred around long sends being cancelled, which loses potentially long conversions—wasting users' time. 

The converted and draft Commit objects are cached on completing the conversion. A successful Transport and Commit nullify these caches.

The Retry Send button is enabled if the Main thread becomes idle and the cached items still exist, indicating either Transport or the Commit Send failed. 

When Retry Send completes, it nullifies the caches, as does changing the document or starting a new Send.

It is a little hacky and isn't completely tied into the DUI just yet - the failed send Sets the CancellationToken as cancelled - can research if this can be uncancelled (A newly created `ProgressViewModel` is divorced from DUI)


AlanRynne (approved):
This is looking good! After your demo yesterday I'm super keen on having this... EVERYWHERE (included as a DUI feature).

Anyhow, in it goes!

AlanRynne:
Cudos for the detailed message history 🚀

PR #2678, opened 2023-06-16
fix(rhino): mapper wall logic for extrusions

clairekuang:
Improves the wall conversion method for mapped extrusions in Rhino. Previously, the logic assumed all `Wall` mappings were extrusion geometry which have been created by extruding a line upwards. The new logic should take into account:

- `Wall` mapped breps
- `Wall` mapped extrusions that were created from top to bottom direction, or a side to side direction.

Fixes an issue reported by @bimgeek where revit could not successfully receive mapped walls due to incorrect wall baseline and heights.

### DIFF
--- a/Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Mappings.cs
+++ b/Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Mappings.cs
@@ -60,9 +60,24 @@ public partial class ConverterRhinoGh
 
         //NOTE: this works for BOTH the Wall.cs class and RevitWall.cs class etc :)
         case Wall o:
-          var extrusion = (RH.Extrusion)@object.Geometry;
-          var bottomCrv = extrusion.Profile3d(new RH.ComponentIndex(RH.ComponentIndexType.ExtrusionBottomProfile, 0));
-          var topCrv = extrusion.Profile3d(new RH.ComponentIndex(RH.ComponentIndexType.ExtrusionTopProfile, 0));
+          var wallBrep = @object.Geometry is RH.Brep wallB ? wallB : ((RH.Extrusion)@object.Geometry)?.ToBrep();
+          if (wallBrep == null)
+          {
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Mappings.cs
```diff
@@ -60,9 +60,24 @@ private Base MappingToSpeckle(string mapping, RhinoObject @object, List<string>
 
         //NOTE: this works for BOTH the Wall.cs class and RevitWall.cs class etc :)
         case Wall o:
-          var extrusion = (RH.Extrusion)@object.Geometry;
-          var bottomCrv = extrusion.Profile3d(new RH.ComponentIndex(RH.ComponentIndexType.ExtrusionBottomProfile, 0));
-          var topCrv = extrusion.Profile3d(new RH.ComponentIndex(RH.ComponentIndexType.ExtrusionTopProfile, 0));
+          var wallBrep = @object.Geometry is RH.Brep wallB ? wallB : ((RH.Extrusion)@object.Geometry)?.ToBrep();
```
AlanRynne, line 63: Working with BREP instead is a nice move, but I'm guessing that this is intended for "Single Surface BREPs", not for complex multi face ones.

Maybe we should throw a meaningful message if the BREP has more than 2 surfaces? Or are we doing this somewhere else in the mapping tool?

AlanRynne (changes requested):
Just one suggestion

Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Mappings.cs
AlanRynne: Ok then we can go ahead, as any other weird brep will fall under "invalid" because it will have more than 2 edges for sure.

AlanRynne (approved):
Based on the previous review conversation I'm approving this

PR #2679, opened 2023-06-16
fix(core): AccountManager.GetUserInfo was still using old user query

AlanRynne:
Fixes issue reported in the forum, where Account.ToString would throw because it would be missing the `User` information after validation.

This was caused by the fact that, even though the queries were updated to use `activeUser` a while back, **the responses were not**


AlanRynne:
We should maybe check to ensure this is not happening anywhere else.

Response will not contain an error, instead it will just have `UserInfo` set to `null`

PR #2681, opened 2023-06-16
Fix(Revit): properly switch between sending new instances and old instances

connorivy:
## Description & motivation

Previously, we've been converting all revit family instances as instances except for structural foundations or generic type elements. We knew this was the incorrect criteria for determining when instances wouldn't work, but we didn't know when. Now, glory to God, I believe I found the correct indicator for determining this.

When a Revit element has the "HasModifiedGeometry" flag as true, then it won't give you and symbol geometry in the mesh retrieval. Instead it gives you the inplace geometry. If we want to in the future, we could apply the inverse of the element's transform to this mesh to send everything as an instance, but for now I am just happy to get rid of everything floating in space.

## To-do before merge:

Because of the similar changes I was making, I built this PR on top of this PR https://github.com/specklesystems/speckle-sharp/pull/2669. Which also allowed me to make sure that my tests passed. Must merge the above PR before we can merge this one.

## Screenshots:
Before
image

After
image

## Checklist:


clairekuang:
I *think* I found a fix for the displayvalues of all family instances with modified geometry, allowing us to continue sending them as `RevitInstance` objects and deprecate the old `PointBasedFamilyInstanceToSpeckle()` method. See:[https://github.com/specklesystems/speckle-sharp/tree/revit/claire/revit-instance-fix]( https://github.com/specklesystems/speckle-sharp/tree/revit/claire/revit-instance-fix)

clairekuang:
image

PR #2682, opened 2023-06-17
feat(Navisworks): Conversion Cache

jsdbroughton:
This is a developer feature that runs only under Debugging.

If the Cache Persistence is set to on, any Send will proceed as normal, but subsequent sends will use the Converted Elements from the cache.

This feature allows for the same set of Converted Elements to be sent to multiple streams and branches with no additional conversions. 

Any changes to the Send Settings do not get applied to the converted objects, this is purely enabling sending to new streams where the diffing will always return as new objects. This is to test server resistance. Failed sends can still be sent with the Retry Send which uses the same caching mechanism. The difference here is that the cache will persist even after successful sends.


AlanRynne:
Oh, just saw we're tracking Caching implementations in a notion page https://www.notion.so/speckle/Connectors-Caching-e082240762964324ad46abb1c3dbdd4e

Merging in!

PR #2691, opened 2023-06-22
feat(revit): design options support

### LINKED ISSUE #1546: Add support for design options and worksets
When sending models to the Revit Connector; users might want to specify a phase, design option, and workset.
Especially if this happens from Grasshopper, see context: https://speckle.community/t/send-model-variants-from-gh-to-revit/3458

The match would probably have to happen by name. We would need to decide where these props live, if it's the base Revit element or if they should be defined as parameters instead.

teocomi:
Partially fixes #1546 

Adds support for design options: https://speckle.community/t/revit-design-option-settings-are-ignored-in-everything-stream/3182/8

- by default, they are excluded, and only the main model is sent
- if any design option is selected, then also all relevant elements are sent

No Design Option
image

Design Option Selected
image

### DIFF
--- a/All.sln
+++ b/All.sln
@@ -237,9 +237,6 @@ EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ConverterNavisworks", "Objects\Converters\ConverterNavisworks\ConverterNavisworks\ConverterNavisworks.shproj", "{C3232EF3-2000-44C6-A330-B94531C9CC83}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Navisworks", "Navisworks", "{B6887DDC-B9B9-4B00-95DC-1DD930A1E901}"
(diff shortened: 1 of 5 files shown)


AlanRynne:
There's an old issue for this 🙌 #1546

Linking it to this PR so it gets tracked :)

connorivy (changes requested):
I would say that this behavior needs a bit more fine tuning before merging.

A couple notes that I have on the behavior:
- Elements in design options can be selected while in main model mode (though you have to make them selectable). Currently if a user sends a selection with objects that live in a design option that isn't currently active, then those elements will not be sent.
- All design option sets have a "primary" option (I think. Please let me know if that is not the case) which is on in every view by default. I think I would expect those elements to be sent in all sending modes, but they are currently not sent in any.

teocomi:
@connorivy, as discussed, now it's sending the Main Model objects and those part of a Primary Design Option

PR #2692, opened 2023-06-22
feat(core): Allow for traversal context inheritance

JR-Morgan:
Added support for subclassing the TraversalContext object to allow for custom information to be attached while traversing.

This might be useful for
1. Keeping track of matrix transforms.
2. Propagating displayStyles / layer properties to children.
3. Keeping track of collection/layer info.

Example (untested) of what this might look like:
```cs
public sealed class ArchicadTraversalContext : TraversalContext<ArchicadTraversalContext>
{
  public Matrix4x4 matrix { get; private set; }
  public ArchicadTraversalContext(Base current, string? propName = null, ArchicadTraversalContext? parent = default)
    : base(current, propName, parent)
  {
    var parentMatrix = parent?.matrix ?? Matrix4x4.Identity;
    if(current is Instance instance)
    {
        matrix  = parentMatrix  * instance.Transform.matrix; //TODO: I haven't checked this maths, and this doesn't take into account if the parent has different units to the child.
    }
      
  }
}

public sealed class ArchicadTraversal: GraphTraversal<ArchicadTraversalContext>
{
  protected override ArchicadTraversalContext NewContext(Base current, string? propName, ArchicadTraversalContext? parent)
  {
    return new ArchicadTraversalContext(current, propName, parent);
  }
}
```

### DIFF
--- a/Core/Core/Models/GraphTraversal/GraphTraversal.cs
+++ b/Core/Core/Models/GraphTraversal/GraphTraversal.cs
@@ -4,27 +4,53 @@ using System.Collections.Generic;
 
 namespace Speckle.Core.Models.GraphTraversal;
 
-public sealed class TraversalContext
+
+public class TraversalContext
 {
-  public readonly Base current;
-  public readonly TraversalContext? parent;
-  public readonly string? propName;
-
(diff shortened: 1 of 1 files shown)


jozseflkiss (approved):
Tested with Archicad connection, works fine.

PR #2693, opened 2023-06-22
fix(core): Fix for requests for invalid auth tokens failing 

JR-Morgan:
There has been a change on our GraphQL api that means that specifying an invalid `Authorization` header (e.g. an empty token leading to a  `"Bearer "` header value) would  lead to a 403 response with the following error 
```json
{
  "error": "Your token is not valid."
}
```

---

I have replaced the few places in sharp where we were adding auth headers to HTTP/GQL clients with a call to new `public static void Helpers.HTTP.AddAuthHeader(HttpClient client, string? authToken)`  function


Core/Core/Credentials/AccountManager.cs
```diff
@@ -67,7 +67,7 @@ public static async Task<ServerInfo> GetServerInfo(string server)
   public static async Task<UserInfo> GetUserInfo(string token, string server)
   {
     using var httpClient = Http.GetHttpProxyClient();
-    httpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer {token}");
+    Http.AddAuthHeader(httpClient, token);
```
AlanRynne, line 70: Much nicer 😄

AlanRynne (approved):
🙌🏼 Great job @JR-Morgan. Pending @gjedlicska approval this is good to go on my behalf

gjedlicska (approved):

PR #2695, opened 2023-06-23
fix(revit): correct display meshes for instances with modified geometry

clairekuang:
## Description & motivation

This pr fixes display mesh retrieval for revit instances that have modified geometry, where previously some fo these instances (like structural foundations) were using the old family instance method or were not accounted for.

Also includes full deprecation of the `PointBasedFamilyInstanceToSpeckle` method, so all point based instances are now sent as `RevitInstance` elements.

## Changes:

-Revit converter

## Screenshots:

Tested with structural foundations and a sample file that a user provided.
image

## Checklist:


connorivy:
looks like when you merged dev into this branch, you overwrote your changes, so this isn't working at the moment. Also, when we're transforming the solids / meshes, I would prefer to either log the exception that is thrown or to just not wrap those in a try catch at all.

clairekuang:
Ah shoot yeah, just fixed the issues from the dev merge as well as the try catches. Should be good to test again!

connorivy:
Awesome, approved. Also, I did some profiling and found that the SolidUtils.CreateTransformed method doesn't seem to be expensive at all.

connorivy (approved):

PR #2702, opened 2023-06-26
feat(acadcivil): add 2024 support for autocad and civil3d

### LINKED ISSUE #2543: Civil and AutoCAD 2024 support

clairekuang:
Also published 2024 autocad and civil3d nugets.

Closes #2543


AlanRynne (approved):
Looks good!

Found nothing missing on the csproj side so I think we can go ahead with this

PR #2703, opened 2023-06-27
AutoCAD Connector: Add Feature name block anonymous

chuongmep:
## Description & motivation

Add Support send cad block include name and anonymous name, at the moment just show name.

## Changes:

- Change function to return two name : Block Name and Anonymous Name of Block.
- Stream Demo: https://speckle.xyz/streams/088f1b9f0d/branches/main
## Screenshots:

Before Change: 
- Send Block just show definition Block Name
image

After Change: 
- Support for show definition Block Name and Block Anonymous Name
image
image

## Checklist:

### DIFF
--- a/Objects/Objects/Other/Block.cs
+++ b/Objects/Objects/Other/Block.cs
@@ -21,7 +21,7 @@ public class BlockDefinition : Base
   }
 
   public string name { get; set; }
-
+  public string anonymousName { get; set; }
   /// <summary>
   /// The definition base point of the block
   /// </summary>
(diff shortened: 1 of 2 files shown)


Objects/Objects/Other/Block.cs
```diff
@@ -21,7 +21,7 @@ public BlockDefinition(string name, List<Base> geometry, Point basePoint = null)
   }
 
   public string name { get; set; }
-
+  public string anonymousName { get; set; }
```
clairekuang, line 24: If you really need the anonymous name property, instead of adding it to the `BlockDefinition` class (which is used by other applications like Rhino and Sketchup that do not have anonymous names), please add it as a dynamic property attached to the block definition instead

clairekuang (changes requested):
As far as I'm aware, anonymous names are used for block definitions that serve as base definitions for dynamic blocks and other elements - the naming convention `*Unnn` for the anonymous name for a dynamic block base can have the last 3 `nnn` values differ depending on session. Not sure if this is really a needed property, but if it is it should be dynamically attached to the block definition instead of added as a property to the class!

PR #2705, opened 2023-06-27
Feat(Revit) : design option support

connorivy:
## Description & motivation

This PR is built on top of an existing PR here. https://github.com/specklesystems/speckle-sharp/pull/2691/files

The main goal is to recognize design options in Revit and only send what the user can see in the document.

## Changes:

- Split up 300 line method into many many different methods. This makes the PR look like more changes than it really is
- Changed logic from alway sending primary to checking if another option is active first, and then either sending the primary or the active design option.
- Added a few cheaper checks before looping through every element to check for the design option
  - if the selection method is manual, then don't check anything. I believe the user's selection should not be altered. 99.9% of the time, the user's selection won't contain an element in a hidden design option, so filtering wouldn't do anything. The .1% of the time when the user's selection would contain an element in a hidden design optionis when the user makes a selection, and then switches the visible design option. Again, I believe that we should not filter out the hidden elements in this scenario, but I'm happy to discuss otherwise.
  - filters the document for a design group. If no design options exist in the project, then don't do any filtering
- fixed issue with the element selection from views that has been reported many times as you can see here in seq https://seq.speckle.systems/#/events?filter=Failed%20to%20filter%20objects&signal=signal-m33301,signal-39&from=2023-05-01T05:00:00.000Z&to=2023-06-27T10:00:35.943Z
- changed the catch statement from an exception log to throwing an exception. Previously, we were just swallowing and logging the exception, and then returning an empty list of elements. This would result in the user getting a message that the send operation failed due to "no elements being selected" which is misleading and not helpful. Now the user will get an exception telling them (and us) exactly why the GetSelectionFilterObjects failed

## Checklist:

## References

https://github.com/specklesystems/speckle-sharp/pull/2691/files


AlanRynne:
So, is this dependent on #2691 or does it make it redundant? i.e do we need to merge 2691 first, or can we close and merge this instead?

PR #2706, opened 2023-06-27
fix(Navisworks): Handle Incomplete Viewpoint Definitions

jsdbroughton:
Testing revealed an obscure bug whereby the getter for `FocalDistance` throws an error if a viewpoint doesn't have one set.

The behaviour is inconsistent. No `FocalDistance` being set on a `SavedViewpoint` may produce a null, or throw either `NotSupportedException` or `InteropServices.COMException`.

This fix wraps the assignment in a try-catch and logs the incidence.

In addition, and this bug surfaced inside a main thread invocation, the SavedViewpoint is no resolved from reference in the same invocation Action as the assignment of that view to the Current Viewpoint.


AlanRynne (approved):
Beautiful! Approving so it can go into the next alpha.

PR #2708, opened 2023-06-28
feat(revit): set phases on rooms when sending to speckle

teocomi:
Mini bug-fix/enhancement sets the phase on Revit rooms: https://speckle.community/t/model-parameters-strategy/6158/8


connorivy (approved):
looks good, I just fixed some formatting that was messed up

PR #2712, opened 2023-06-29
fix(gh): StreamUpdate now uses nullable boolean

AlanRynne:
This prevents accidental toggling when the default value was different from the original stream value.

Reported in Notion testing page by Kat! 🙌🏼 https://www.notion.so/speckle/Grasshopper-1d0064e3d80d4589a4baf622bca13490?pvs=4#5fe81eaa01364dac960d11ad8f5050c1


AlanRynne:
Merging in so it's re-tested on the next alpha

PR #2714, opened 2023-06-29
Fix(CSi) : add polygon mesher

connorivy:
## Description & motivation

- polygonMesher is a dependency of the csi converter. However, we're only copying the converter itself to the objects kit. This was resulting in the conversions failing due to a missing dependency.

- I made a bit of a bandaid fix where I added a class that uses the converter dependency in the connector. Therefore the dependency will be there for the converter to use. We should probably make a policy for what to do with converter dependencies.

## Validation of changes:

- 2d objects used to fail to send, but now they do not

## Checklist:

## References

https://speckle.community/t/specklepy-receiving-commit-data-from-etabs/5831/12


AlanRynne (approved):
Oh yeah, we talked about this and it seemed to be the best way forward right now, although not ideal.

Thanks for the changes @connorivy.

PR #2718, opened 2023-06-30
Fix(Revit): Fixed issue with revit traversal of displayValues

JR-Morgan:
Fixed traversal regression with displayValues receiving in Revit


Core/Core/Models/GraphTraversal/GraphTraversal.cs
```diff
@@ -73,7 +74,10 @@ public IEnumerable<T> Traverse(Base root)
       Base current = head.current;
       var activeRule = GetActiveRuleOrDefault(current);
 
-      foreach (string childProp in activeRule.MembersToTraverse(current))
+      var toTraverse = activeRule.MembersToTraverse(current).ToList();
+
+      _ = 0;
```
jsdbroughton, line 79: ???

jsdbroughton: is this JBDD's breakpoint holder pattern?

jsdbroughton (approved):
I don't mind leaving debug code in

Core/Core/Models/GraphTraversal/GraphTraversal.cs
JR-Morgan: It is, good catch, I'll revert this 😁

PR #2719, opened 2023-06-30
Fix(Tekla) : initialize logger

connorivy:
## Changes:

- initialize logger in tekla to get better logs in seq


AlanRynne (approved):
Nice! Thx Connor! merging so we can get this in the next test release

PR #2720, opened 2023-06-30
Fix(Tekla) add polygon mesher dep

connorivy:
## Description & motivation

Same issue and resolution as https://github.com/specklesystems/speckle-sharp/pull/2714

### DIFF
--- a/ConnectorTeklaStructures/ConnectorTeklaStructuresShared/PolygonMesherReferencer.cs
+++ b/ConnectorTeklaStructures/ConnectorTeklaStructuresShared/PolygonMesherReferencer.cs
@@ -2,8 +2,10 @@ using StructuralUtilities.PolygonMesher;
 
 namespace ConnectorTeklaStructuresShared
 {
+  [Obsolete("See comment within implementation of UnusedMethod", true)]
   internal class PolygonMesherReferencer
   {
+    [Obsolete("See comment within implementation", true)]
     public PolygonMesher UnusedMethod()
     {
       // This class is only here to throw an error if the polygon mesher dependency is ever removed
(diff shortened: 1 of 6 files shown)


ConnectorTeklaStructures/ConnectorTeklaStructuresShared/PolygonMesherReferencer.cs
```diff
@@ -0,0 +1,16 @@
+using StructuralUtilities.PolygonMesher;
+
+namespace ConnectorTeklaStructuresShared
+{
+  internal class PolygonMesherReferencer
+  {
+    public PolygonMesher UnusedMethod()
```
AlanRynne, line 7: Hmm I saw the comment but this looks dodgy...

@JR-Morgan what do you think about this? The problem it describes is legit though

AlanRynne (commented):
I'd like to double check with Jedd if the `PolygonMesherReferencer` isn't something that would bite us back in the short term.

Other than that... all good

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/PolygonMesherReferencer.cs
AlanRynne: After a quick chat with Jedd we're assuming this is to introduce a forced usage of `PolygonMesher` to prevent it's accidental removal in the future?

Waiting on confirmation of this and I'll approve :)

AlanRynne (commented):
fix(rvt): PEBKAC... missing using statement

AlanRynne (approved):
Approving after chat with connor and minor modifications

Issue #2657, opened 2023-06-12
feature(Navisworks): Coalesce data from "FirstObject" to Geometry Node


jsdbroughton:
Currently, selecting objects in the viewer and filtering them in PowerBI leads to missed expectations as key properties are often held on a parental collection.

Consider offering a feature to coalesce these Object data down the inheritance tree.

As multiple "First Object" levels can exist, the suggested implementation is to ascend the tree from the Geometry Node only to the first qualifying node that satisfies the First Object criteria. Coalescing progressive properties back to the Geometry Node could exist from there.

Alternatives to consider:
Limit the propagated properties to those defined as Quick Properties. Potentially also going beyond the First Object ancestor.

PR #2724, opened 2023-07-03
feat(Navisworks): Option to coalesce all data from First Object to the Geometry Node

jsdbroughton:
Implements #2657 

This feature will populate a Geometry Node with all the data from itself up to the "First Object". The "First Object" is a special state related to what Navisworks understands as a significant Node for selection. Many First Objects can exist in an ancestral tree; logically, the reverse ancestral lookup makes the most sense.

Selecting this option will also include the reference nodes in the commit tree. In this way, it will be more referential than a simple node selection - it will provide context for where the data has been gleaned.

The performance in the viewer with this change is that Data that may be instantiated in a First Object will now propagate to the selected geometry object in the Speckle frontend. This also brings much greater Data Filtering capabilities.

### Connector
- ConnectorNavisworks/Bindings/**ConnectorNavisworksBindings.Send.cs**
  - Moves the Dictionary mapping to the method that makes use of it. Allows for back referencing within the data traversal
- ConnectorNavisworks/Bindings/**ConnectorNavisworksBindings.Settings.cs**
  - Adds a checkbox setting for this feature. Default Off 
- ConnectorNavisworks/Other/**Constants.cs**
  - Makes the COnversion State enum public for access in Elements
- ConnectorNavisworks/Other/**Elements.cs**
  - New method `AddPropertyStackToGeometryNode` gets all data props from Geometry Nodes to their ancestral FirstObject and adds them to a dictionary. This is collating only a single data property by key - closest node to Geometry or Self overwrites parental values
- ConnectorNavisworks/Other/**SelectionBuilder.cs**
  - Ensures the parent nodes are included as ancestors for conversion.
  
 ## Before
image

## After
image

### DIFF
--- a/ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Send.cs
+++ b/ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Send.cs
@@ -299,8 +299,8 @@ public partial class ConnectorBindingsNavisworks
     }
 
     _progressBar.StartNewSubOperation(0.66, "Building a family tree, data-style...");
-    var elements = converted.ToDictionary(x => x.Key.PseudoId, x => x.Value.Item2);
-    commitObject.elements = Element.BuildNestedObjectHierarchy(elements).ToList();
+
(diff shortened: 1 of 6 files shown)


AlanRynne (approved):
Good to go after our quick chat!

PR #2726, opened 2023-07-03
fix(rhino): correct layers on create receive mode

### LINKED ISSUE #2710: bug(rhino): objects not correctly placed on layers on create receive mode
## Describe the bug

When receiving commits from Rhino and revit on create mode, some objects are place on the `default` layer instead of the correct layer. This was found when investigating another gh>rhino layer issue during 2.15, and occurs in the 2.14 releases.

## To Reproduce

1. Send some elements from the sample revit house
2. Receive elements in rhino on `Create` mode
3. Objects are placed on the default layer, with the report exception error message `"Could not create layer [{path}]. Objects will be placed on default layer."`

clairekuang:
## Description & motivation
Creates correct layers on `create` receive mode in Rhino: issue was being caused by the base commit layer not being created before other layers.

Also contians refactor of connector bindings to split methods into separate files.

Fixes #2710

## Changes:

- rhino connector
- rhino converter

image


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -0,0 +1,644 @@
...
+          var layers = new Dictionary<string, Layer>();
+          var containers = Preview
+            .Select(o => o.Container)
+            .Distinct()
+            .ToList()
+            .OrderBy(path => path.Count(c => c == ':'))
+            .ToList();
+          // if on create mode, make sure the parent commit layer is created first
+          if (state.ReceiveMode == ReceiveMode.Create)
+          {
```
clairekuang, line 146: This code chunk is where the bug fix is, along with the change in the converter

AlanRynne (approved):
Very nice refactor ♥️ 

I don't see any gotchas here, but the diff makes it difficult to find anything so lets go for it and test it once it's out.

PR #2727, opened 2023-07-03
fix(rvt): View selection filter now uses `Title` instead of `Name`

AlanRynne:
Fixes ambiguous view names in filter view.

Reported in 2.15 testing in the Notion table.
Screenshot 2023-07-03 at 17 15 03


connorivy (approved):
Tested and approved

PR #2735, opened 2023-07-06
fix(tekla): properly load DUI after it's been closed, sets the window owner

teocomi:
This PR fixes 2 issues:

- it was not possible to re-open DUI after the window was closed (and there were some unnecessary extra calls to get streams etc) - the main Tekla window is now set as owner of DUI so that DUI doesn't hide back when unfocused
- 
tekla-win

### DIFF
--- a/ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainForm.cs
+++ b/ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainForm.cs
@@ -1,53 +1,75 @@
 using System;
 using System.IO;
-
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Threading;
 using Avalonia;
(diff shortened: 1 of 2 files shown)


ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainForm.cs
```diff
@@ -1,53 +1,75 @@
...
+          };
+
+
+
+          Bindings.OpenTeklaStructures();
 
+        }
+        catch (Exception ex)
+        {
+          SpeckleLog.Logger.Fatal(ex, "Failed to create main form");
```
AlanRynne, line 60: Nice extra log 😎

AlanRynne (approved):
Looks good! I don't have Tekla to verify everything works as intended though

PR #2750, opened 2023-07-11
enhancement(Navisworks): Add global unhandled exception handler...

jsdbroughton:


ConnectorNavisworks/ConnectorNavisworks/Entry/SpeckleNavisworksCommandPlugin.cs
```diff
@@ -23,24 +24,19 @@ namespace Speckle.ConnectorNavisworks.Entry;
...
-]
+  ), UsedImplicitly]
 internal sealed class SpeckleNavisworksCommandPlugin : DockPanePlugin
 {
   internal ConnectorBindingsNavisworks Bindings;
 
   public override Control CreateControlPane()
   {
     AppDomain.CurrentDomain.AssemblyResolve += OnAssemblyResolve;
+    AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
```
teocomi, line 35: Curious, does it work as expected?

jsdbroughton: So far, I've not been triggering enough errors to know.

This was an observed pattern across several other Navisworks plugins. As each plugin is in an isolated Worker Thread (until it isn't), the fact other plugins adopt this pattern, but still, we can trigger fatal is telling.

In this specific case, I'm looking to head off errors arising from the Main Thread invocation that can fail to capture with the InvokerHelper wrappers I have employed; these are almost exclusively Interop errors and get caught by the hostApp global error handler. 

I intend to see what this can learn. I could log the fatal and then rethrow to ensure no change in operation and then learn from the logging, which should be more contextual than relying on crash dumps.

TBD.

PR #2755, opened 2023-07-12
fix(revit): instance display mesh fix once and for all🤞

clairekuang:
## Description & motivation

Community-provided test file on the 2.15 announcement in the forum showed that some instances (with `.HasModifiedGeometry() = false` ) still have incorrect display value meshes on send.

This fix will always retreive the instance geometry rather than the symbol geometry when converting definitions, and then untransform the instance geomtry - hopefully this will be the most reliable approach to accurate display meshes + transforms.

## To-do before merge:

- [ ] test as many user-provided revit files as possible, especially those with previously out of place instances

## Screenshots:
Before:
image

After:
image

## Validation of changes:

Tested on Revit2023 & 2024:
- face based instances
- simple wall
- rool wall floor hosted instances
- sample architectural model
- sample structural model


connorivy:
This PR was superseded by https://github.com/specklesystems/speckle-sharp/pull/2771

PR #2759, opened 2023-07-12
fix(Navisworks): Handles lack of One-click-send implementation causing host app to be unresponsive

jsdbroughton:
As reported on the forum:

https://speckle.community/t/navisworks-speckle-greys-out-after-use-and-naviswork-need-to-be-restarted/6311?u=jonathon

And it is replicable.

Two situations occur: 

1. One Click Send is enabled in the DUI even with a completely empty document with no prospect of sending anything.
    - Code now guards by throwing a `FileNotFoundException` 
2. With the file present, I am still in the defensive mode about allowing Everything mode. 
    - The code now validates ahead of the progress bar being initiated.

TODO: Consider changes to DUI2 logic on Send so that it is possible to legitimate communicate that a new Stream should not be created under the circumstances that the Connector is in control of - like a `get` only property Binding. Might hold for DUI3 (@teocomi ?)


jsdbroughton:
this kinda sucks

image

teocomi (approved):

Issue #2762, opened 2023-07-13
bug(Navisworks): progress measurement causing conversion failure


jsdbroughton:
As reported on the forum:

[error when I tried to send a model from Navis to Speckle.](https://speckle.community/t/speckle-2-15-release-candidate-is-here/6276/9?u=jonathon)

![image](https://github.com/specklesystems/speckle-sharp/assets/760691/b6c7353c-af85-439e-8ce1-486684acdb7f)

PR #2763, opened 2023-07-13
fix:(Navisworks): progress bar should have a maxium value of 1.0

jsdbroughton:
closes: #2762 

An occasional rounding error was causing some updates to the progress bar of greater than 100% (or over 1.0). This throws an error from the host application causing a failed conversion)

Changes:
For this hotfix, the Invoker method updating the progress UI has a `Math.Max(fractiondone,1)` shield.

TODO: Will revisit this as there are three different methodologies currently used to make traversal progress reporting - should adopt a single pattern


jsdbroughton:
bound into https://github.com/specklesystems/speckle-sharp/pull/2766

PR #2773, opened 2023-07-14
feat(revit) : adds non-length units to parameters

chuongmep:
## Description & motivation

 - Resolve Problem with Parameter Data can use for data analyst
 - [x] Unit detail with parameter value - Before request `Unit` is null
 - [ ] Add Units data information for model 
## Screenshots:

- Resolve Parameter Unit Property detail with more information.

image

- Add Units general for model :

image

Detail : https://speckle.chuongmep.com/streams/b268a4f540/commits/a0b05bf882
## Checklist:


chuongmep:
Please don't Merge, It is still working on process.

chuongmep:
Hi, can give me response for this pull request, I don't think really need one month for review this. I don't want keep a pull request too long to resolve.

teocomi:
Hey @chuongmep, thanks for the ping!
We're currently reviewing this, it's taking us a bit of time as we need to be careful with any changes to the objects library - will post some updates soo.

chuongmep:
Thank you @teocomi, I have so many research to focus now, hope can back with team soon.

clairekuang:
Hello @chuongmep , I've reviewed this pr and decided that it would be best to resolve your issue (adding non-length units for data analytics) through a simpler implementation, and will be submitting our own pr as yours would require quite a few changes. This is for two main reasons:
1. We would like to eventually support non-length units in all of our connectors and converters, and want to scope out this feature properly when we do so.  
2. It's not good to introduce a new `Units:Base` class just for `BuiltElements`, since we already[ have one in Core](https://github.com/specklesystems/speckle-sharp/blob/main/Core/Core/Kits/Units.cs). We'd like to constrain the scope of this change to just the Revit converter, so I'm looking into dynamically attaching the non-length units information without the use of a new class
Happy to coordinate with you on the new pr to make sure we are extracting all of the unit information you need!

chuongmep:
@clairekuang maybe I should remove class unit first, and just commit parameter unit.
How you think about this idea ?

clairekuang:
That should be fine - I can make the cleanup directly to your branch if you'd prefer!

chuongmep:
Yes, please help me do it. I don't have more time to help for your team now. remove units class is good idea to go ahead 🙏

clairekuang (approved):

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -497,10 +553,11 @@ public void SetInstanceParameters(Element revitElement, Base speckleElement, Lis
 
         var rp = revitParameterById.ContainsKey(spk.Key) ? revitParameterById[spk.Key] : revitParameterByName[spk.Key];
 
-        TrySetParam(rp, sp.value, sp.units, sp.applicationUnit);
+        TrySetParam(rp, sp.value, applicationUnit: sp.applicationUnit);
       }
     }
 
+    [Obsolete("This method will be removed in feature")]
```
clairekuang, line 560: @chuongmep could you clarify why you added the `Obsolete` tag here?

clairekuang: Removed this flag as the method is still in use

PR #2782, opened 2023-07-18
fix(core): Fixed issue with blob storage folder being ignored in server transport

JR-Morgan:


didimitrie (approved):
LGTM - I don't remember these parts enough. If it's tested and passes ✅

PR #2796, opened 2023-07-21
core: hotfix for sqlite transport swallowing exception when failing t…

JR-Morgan:
This PR contains a small subset of the changes already merged in `dev` for 2.16 regarding transport exception handling - https://github.com/specklesystems/speckle-sharp/pull/2700

The problem with the SQLite transport, is if it fails to initialize a connection to the DB, it "calls" OnErrorAction. However, there is no way for OnErrorAction to be assigned, since this is performed in the constructor.

So in effect, this is a silent swallow of any exceptions that occur when initializing a connection to the DB.

This leaves the transport in an invalid state such that calls to its members will fail with no-descript error messages.

This PR replaces the false call to OnErrorAction with a throw - as per the changes already made in dev. This way callers are aware of their failed sqlite connection

### DIFF
--- a/Core/Core/Transports/Exceptions.cs
+++ b/Core/Core/Transports/Exceptions.cs
@@ -1,0 +1,21 @@
+﻿#nullable enable
+using System;
+
+namespace Speckle.Core.Transports;
+
+public class TransportException : Exception
+{
+  public ITransport? Transport { get; }
+
+  public TransportException(ITransport? transport, string? message, Exception? innerException = null)
+    : base(message, innerException)
+  {
+    Transport = transport;
+  }
+
+  public TransportException() { }
+
+  public TransportException(string message) : base(message) { }
+
+  public TransportException(string message, Exception innerException) : base(message, innerException) { }
+}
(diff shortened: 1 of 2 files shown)


JR-Morgan:
As @tsvilans is [no longer blocked](https://speckle.community/t/c-null-argument-when-receiving-from-sqltransport-with-known-object-id/6416/4) by this, we might not bother hot fixing this for 2.15.

The functional change is already in 2.16 `dev` branch.

didimitrie:
Happy with it, my only fears were re release. Merge away!

didimitrie (approved):

AlanRynne:
So is this to be merged?

> The functional change is already in 2.16 `dev` branch.

I can't find any commit in the `dev` branch that has this,

AlanRynne:
Sorry, closed PR by mistake while writing a comment 🙇🏼‍♂️

JR-Morgan:
The exact commit is not on dev, but the change is https://github.com/specklesystems/speckle-sharp/blob/d05449583efc8499ce25adaf4e7880bfa302baf4/Core/Core/Transports/SQLite.cs#L64

I've merged now

PR #2802, opened 2023-07-26
fix(Navisworks): Civil3D Models failing as geometries containing embedded primitives

jsdbroughton:
As Text Primitives are in fact decimated into lines, the addition of a Text Primitive was inconsequential.

TODO: Processing Text Primitives to the underlying text content may be possible and omitting the geometric


AlanRynne (approved):
Seems "harmless enough" FLW 🙌🏼

So, if I understand your explanation correctly, text is not sent as text but as geometry?

jsdbroughton:
> , text is not sent as text but as geometry?

Everything in Navis is geometry, even text. BUT there is the data to reconstruct a Speckle text object later. Right now this just allows for nodes with fragments that are described as Text Primitives to not block conversion.

PR #2803, opened 2023-07-27
fix: Add condition to 2022 Tekla converter post-build event

AlanRynne:
Fixes broken NUGET publish CI pipeline due to Tekla2022 converter having a post-build event that was not disabled with `IsDesktopBuild = false`.

I also added a Windows only restriction, but it may be overkill.


AlanRynne:
Going for this straight away as its not a code change and will unblock me merging #2804

teocomi (approved):

Issue #2800, opened 2023-07-25
Changes to Transform results in loss of precision


timglewis:
The conversion of the Transform class to use System.Numerics.Matrix4x4 results in a loss of precision as this uses floats instead of doubles.

The change was introduced in #2159.

Whilst this loss of precision may be fine when working with local coordinate transformations, it isn't if you want to place objects accurately in geospatial applications.

Request has been made to change the System.Numerics.Matrix4x4 API here, but doesn't appear to be being addressed anytime soon.

https://github.com/dotnet/runtime/issues/24168

Possible alternative solution is to use the Silk.NET library as mentioned in the comments of the issue above:

https://www.nuget.org/packages/Silk.NET.Maths

PR #2806, opened 2023-07-27
Fix: Transform using Matrix4X4 in Silk.NET

timglewis:
## Fix Summary
- Added the Silk.NET.Maths package
- Converted from Matrix4x4 to Matrix4X4<double>
- Modified TransformTests accordingly

## Description & motivation
Uses a Matrix4x4 with the Silk.NET package that has double precision instead of floats.  This is required for geospatially accurate transformations.

This fix is for #2800


timglewis:
Makes sense Claire.  Happy to help if you have other priorities.

PR #2811, opened 2023-07-27
Refactor(Revit) : Revit MEP elements

connorivy:
## Description & motivation

Our MEP support has been messy for quite a while. We even had to drop our very basic MEP element tests because they were failing and we didn't want to touch the "Networks". This refactor changes the behavior of the MEP elements to be similar to any other element that we support.

## Changes:

Changed all NetworkToSpeckle calls into the proper call for whichever object is being converted
- Aligns with the expectation of what should happen when I call DuctToSpeckle
- Keeps the UI refreshed by not getting stuck in one massive conversion
- Avoids duplicate traversal in the converter
- Any exceptions or unsuccessful revit api calls will not cause the entire system to fail, but only the piece of the system that encountered the issue
- All converted objects will have meaningful report items

Added a ConvertedObjectsCache to the converter
- We have been using the convertedObjectsCache to keep track of converted objects in the connector, but not the converter. However, converting these MEP elements (which all need references to each other for conversion) created a need in the converter to not only know if an element has already been converted, but also know the native object that was created from that object.
- Every object that does not throw an uncaught error in the converter will be added to the convertedObjectsCache

Created a new ConversionNotReady Exception
- When the converter throws this error, the connector will check how many times a certain object has been attempted to be converted. If that number is less than 3, the connector will move the object to the end of the list of objects to be converted and will try again. If not then the connector will treat it the same as any other error

Added IHasConnectors interface in Objects which is implemented on every MEP type
- I didn't know exactly how to structure the MEP elements because I am not an MEP engineer, so I just copied Revit's data structure. Each MEP object has a concept called connectors which are essencially the nodes at the end of the MEP element that attach to other nodes.
- I changed each MEP "ToSpeckle" conversion to capture the element's "connectors"
- Also changed each "ToNative" method to join the new native object's connectors with those around it.

## Screenshots:

Original:
image

[Network](https://latest.speckle.dev/streams/85bc4f61c6/commits/eabbf3c4bd):
image
- Many fittings baked incorrectly
- Incorrectly saying that 1 object was received
- Report is unhelpful

This PR:
image
- Less fittings baked incorrectly
- Correctly reporting number of objects received
- All weird looking fittings have useful error messages and are selectable from the report

Original
image

Networks:
image
- when the conversion throws an error, the entire network stops converting

This PR
image
- when conversion throws, other objects continue to convert

## Validation of changes:

All tests still pass. Sent several networks to make sure they look correct in the viewer. Did not test receiving THAT much

## Checklist:

## References

https://github.com/orgs/specklesystems/projects/18/views/6?pane=issue&itemId=29229165


clairekuang (approved):
Did some testing C3D > Revit and Revit > Revit, looks good ~

teocomi:
Same here, it seems to be improving noticeably the receiving as well:

image

teocomi (approved):

PR #2815, opened 2023-07-29
Fix: Using System.DoubleNumerics for Transforms (#2800)

timglewis:
Fix for issue https://github.com/specklesystems/speckle-sharp/issues/2800 to have transforms in double precision using System.DoubleNumerics

## Changes:
- Converted Transform.cs to use System.DoubleNumerics
- Added a new TransformTests.cs file to test the new transform class
- Changed Converters to use System.DoubleNumerics

## Validation of changes:
Unit tests for Transforms all pass, however testing of the converters have not been conducted. Unsure if the changes to the converters is necessary or whether other work is required here.


clairekuang (approved):
Did some build tests and generally looking good, waiting for @JR-Morgan to review some changes I added to core serialization 👍

JR-Morgan (approved):

PR #2825, opened 2023-08-02
Feat(Revit) : send performance enhancement

connorivy:
## Changes:

- replaced RevitAsync nuget package with custom APIContext class
  - This new version does almost the same thing as the nuget package except it does not subscribe to the revit [IdlingEvent](https://www.revitapidocs.com/2023/e233027b-ba8c-0bd1-37b7-93a066efa5a3.htm)
  - This only speeds up the conversion from .5-1s. However, the idling event documentation says that subscribing to the event in a certain way (that RevitAsync is using) "can result in performance degradation for the system on which Revit is running because the CPU remains fully engaged in serving Idling events during the Revit application's downtime." So it is recommended to avoid using this event

- Throttle Send operation
  - only refresh the UI every .15 seconds. This makes a small impact on the ui refreshing and a huge impact on performance.

- Optimizations for "ParameterToSpeckle"
  - This method is taking the most time of any method that we're sending because it is called so many times. Therefore small optimizations to this method can have a large performance gain.
  -  Broke the parameter value logic out into it's own method "GetParameterValue" to avoid calling the entire ParameterToSpeckle method just to get the value of the convertedParameter
  - Started passing the cache object into the method for more perfomant unit conversions and queries

- Units to Speckle
  - started converting "1" to the desired units and then caching the result as the conversion factor. This is because Revit's "ConvertFromInternalUnits" method is fairly expensive.

## Screenshots:
image

A previous PR lowered the conversion time for this tower from 100s to 70s. This PR lowers the conversion time even further to ~25 seconds.

## Validation of changes:

All unit tests still pass

## Checklist:

### DIFF
--- a/ConnectorRevit/RevitSharedResources/Models/APIContext.cs
+++ b/ConnectorRevit/RevitSharedResources/Models/APIContext.cs
@@ -7,24 +7,26 @@ namespace RevitSharedResources.Models
 {
   public static class APIContext
   {
+    private static UIControlledApplication uiApplication;
     private static ExternalEventHandler<IExternalEventHandler, ExternalEvent> factoryExternalEventHandler;
     private static ExternalEvent factoryExternalEvent;
     public static void Initialize(UIControlledApplication application)
     {
+      uiApplication = application; 
       factoryExternalEventHandler = new(ExternalEvent.Create);
       factoryExternalEvent = ExternalEvent.Create(factoryExternalEventHandler);
     }
 
-    public static async Task<TResult> Run<TResult>(Func<UIApplication, TResult> func)
+    public static async Task<TResult> Run<TResult>(Func<UIControlledApplication, TResult> func)
     {
-      var handler = new ExternalEventHandler<UIApplication, TResult>(func);
+      var handler = new ExternalEventHandler<UIControlledApplication, TResult>(func);
       using var externalEvent = await Run(factoryExternalEventHandler, handler, factoryExternalEvent)
         .ConfigureAwait(false);
 
-      return await Run(handler, null, externalEvent).ConfigureAwait(false);
+      return await Run(handler, uiApplication, externalEvent).ConfigureAwait(false);
     }
     
-    public static async Task Run(Action<UIApplication> action)
+    public static async Task Run(Action<UIControlledApplication> action)
     {
       await Run<object>(app =>
       {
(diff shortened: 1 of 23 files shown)


ConnectorRevit/RevitSharedResources/Models/APIContext.cs
teocomi, line 1: If this is borrowed from the RevitAsync library, we should credit them :)

teocomi (approved):
Looks good, great finds!
Does caching the units conversion ratio really speed things up? Have you benchmarked it on a large model?

ConnectorRevit/RevitSharedResources/Models/APIContext.cs
connorivy: Yes, much of it was borrowed. Does this just mean adding a comment in the class?

connorivy:
> 
Yes it does. All of my optimizations were the result of profiling a large send operation. Caching the scale factor makes a big difference because we are calling scaleToSpeckle multiple times for every object converted. In my case I was sending ~5000 objects which lets say results in calling scaleToSpeckle 10,000 times. If caching the scale factor and using that saves 1ms over using Revit's "ConvertFromInternalUnits" method like we were doing, then the compound time savings would be 10s

connorivy:
> Does caching the units conversion ratio really speed things up? Have you benchmarked it on a large model?

Yes it does. All of my optimizations were the result of profiling a large send operation. Caching the scale factor makes a big difference because we are calling scaleToSpeckle multiple times for every object converted. In my case I was sending ~5000 objects which lets say results in calling scaleToSpeckle 10,000 times. If caching the scale factor and using that saves 1ms over using Revit's "ConvertFromInternalUnits" method like we were doing, then the compound time savings would be 10s

ConnectorRevit/RevitSharedResources/Models/APIContext.cs
teocomi: Yep exactly!

teocomi:
Cool, then merge away after adding the needed credits on the class :)

jsdbroughton:
Already merged into dev - the conversion logic here assumes that all conversions are by a multiplication factor between 1 and desired units.

The internal Revit util ConvertFromInternalUnits for temperature for instance from internal (Kelvin) to target (e.g. Celcius) -272.15

therefore the returned value we are caching is not able to be reapplied by factor * value which for 30*C is returning -82000, say.

PR #2839, opened 2023-08-08
Feat (Rhino): OnIdle

oguzhankoral:
There were some solution fixes and reconsidering usage of `Rhino.Idle` event.


ConnectorRhino/ConnectorRhinoWebUI/Utils/RhinoIdleManager.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Subscribe deferred action to RhinoIdle event to run it whenever Rhino become idle.
+    /// </summary>
+    /// <param name="action"> Action to call whenever Rhino become Idle.</param>
+    public static void SubscribeToIdle(Action action)
+    {
+      calls[action.Method.Name ?? Guid.NewGuid().ToString()] = action; 
+
+      RhinoApp.Idle += RhinoAppOnIdle;
```
didimitrie, line 22: This can allow for multiple subscriptions if called in fast succession, and we might end up firing the idle event multiple times.

PR #2848, opened 2023-08-14
Jrm/core/security analysers

JR-Morgan:
Changes to editor Build Props
1. Added all default .NET7 analysers (minimal analysers) to `WarningsAsErrors` tag
2. Added all "security" category analysers to `WarningsAsErrors` tag (except a couple where I've stated a reason in build props)
3. Resolved Violations
 - Removed hardcoded TLS (we were doing this due to .net4.6.2, but I've tested with rhino 6, still works without hardcoding) 
 - Set `CheckCertificateRevocationList = true` for `SpeckleHTTPClientHandler`
 - Replaced 

Things to discuss
- I've enabled these rules globally in the Build.Props, imo this is a MUST for core, but potentially could relax for connectors?


AlanRynne (approved):
Gone through all the changes carefully and it's looking good!

CI is now passing except for Archicad build steps (which are not related to this issue).

Once this is solved we can merge

PR #2849, opened 2023-08-14
Fix(Revit) : fatal crash in Revit and speed upgrade for large receive operations

connorivy:
## Description & motivation

When sending the revit sample arch model to Speckle and then trying to receive it again, you'll probably end up with your revit having crashed. This is a result of a pesky face hosted element that is somehow corrupting the document (perhapse we are picking the wrong face for the element?). This is easily fixed by just rolling back the transaction, however our connector is currently only using a single transaction.

This PR wraps every attempted baking operation in a subtransaction and then either commits or rolls it back depending on if any exceptions were thrown.

It also commits the transaction (different than the subtransactions) for every 50 objects that are baked because this dramatically increases the speed of the receive operation. Unfortunately this does have a downside of making dialogues pop up in the middle of receive operations if something goes wrong with the transaction commiting. However, we can fix this in the future.

## Changes:

- Subscribe to the unhandled exceptions and log those
- Transactions are now handled by a TransactionManager object.
  - Every transaction happens within a Revit "SubTransaction" so that it can be rolled-back on failure
  - every 50 objects, we are commiting the open Transaction. This is what leads to the bulk of the receive operation speed enhancement
  - moved the "ExecuteInTemporaryTransaction" functionality from the schedule partial class to be part of the transaction manager

## Validation of changes:

all integration tests still pass, however no new ones have been added.

## Checklist:


teocomi (approved):
Nice one, please give us all a quick run through at SU :)

AlanRynne:
Updating branch from `dev` to have CI pass

PR #2852, opened 2023-08-15
Fix(Revit) : Remove cool retractable roof feature

connorivy:
## Description & motivation

Simple fix for importing extrusions roofs so that they come in correctly instead of like this 
image

## Changes:

AFAIK, when sending the roof to speckle, there is no way to retreive the orientation of the reference plane. Therefore the is a 50% chance that when receiving an extrusion roof, the side offsets will be flipped. Therefore we need to check if the startOffset is actually equal to the end offset and if so we need to delete the roof, flip the reference plane, and recreate it.

## Validation of changes:

Receiving sample house roof as well as roofs in the Revit integrations tests create the roofs in the correct location

## Checklist:


teocomi (approved):
Too bad, I liked the retractable roof! 🤣

PR #2860, opened 2023-08-18
fix(dui): do not break the get streams loop if one server times out

teocomi:
Reported here: https://speckle.community/t/connector-does-not-show-streams-special-case/6797/2

If one of the Servers in the user accounts times out, the `OperationCanceledException` try-catch catches an exception with inner exception of type `TimeoutException`.
This is currently breaking the `GetStreams` loop, meaning that the user would not see any streams, even if the other accounts/servers are fine.

Since we're not 100% sure what over events might trigger a `OperationCanceledException` and since there is no way to currently cancel the Refresh/GetStreams operation, we think removing the try-catch is enough to solve the issue.

This change introduced here: https://github.com/specklesystems/speckle-sharp/commit/899989937cbbee3190ed80a314158bedc23f14b7


DesktopUI2/DesktopUI2/ViewModels/HomeViewModel.cs
```diff
@@ -202,10 +202,6 @@ private async Task GetStreams()
 
           streams.AddRange(result.Select(x => new StreamAccountWrapper(x, account.Account)));
         }
-        catch (OperationCanceledException)
-        {
-          return;
```
JR-Morgan, line 207: The purpose of this return was to prevent user cancellation (through the cancellation token) from creating an error popup.

Testing your branch, it seems this is now broken.

https://github.com/specklesystems/speckle-sharp/assets/45512892/65425c7a-68ee-4ea4-902d-08a3cc3a24e2

JR-Morgan (changes requested):

DesktopUI2/DesktopUI2/ViewModels/HomeViewModel.cs
JR-Morgan: I've just tested the following code, thinking it could be a possible fix.
And while it does catch some of the cancellations, it looks like some are still originating from other cancellation tokens. Which is a little concerning.
```cs
        catch (OperationCanceledException) when (StreamGetCancelTokenSource.Token.IsCancellationRequested)
        {
          return;
        }
 ```
Some more investigation is required.

teocomi: Ahh damn, I didn't think about the search and that it can be canceled!
If we change the `return` with a `continue` would it work?

JR-Morgan: That seems to work, shall I commit to your branch?

JR-Morgan (approved):

PR #2862, opened 2023-08-18
feat(revit): Added support for nested family instances inside adaptive components

### LINKED ISSUE #2471: Curtain Panel Nested Families Issue
See forum reports: 
https://speckle.community/t/curtain-wall-panel-with-nested-shared-famillies/5508/2?u=teocomi

Sample file by [M.Tarabishy](https://speckle.community/u/M.Tarabishy) and [Thomas-Lcp](https://speckle.community/u/Thomas-Lcp):
https://drive.google.com/drive/folders/1b_xpPVGTUG4lb3ovLtSXCuJZXC8NZVY6?usp=share_link

Note from Thomas:

> If you want to reproduce my issues then go to “EXPORT IFC” view and select all “MGMF_PMR_Frigorifique” then open Speckle, select a steam, and click to preview to crash your Revit or send to see what happen in online view.

JR-Morgan:
Potential fix for #2471

[M.Tarabishy](https://speckle.community/u/M.Tarabishy)'s sample file contained Adaptive components that contained nested families. 

Originally, these nested families would not send (using selection filter), unless individually selected along with the parent AdaptiveComponent.

This change ensures that nested components will be converted as well. (and won't be duplicated if also selected manually, or with another filter mode)

closes #2471


clairekuang (changes requested):
Duplicated nested instances on receive:
image

JR-Morgan:
Pr parked, since we don't have a nice way to prevent duplicated receives with the current object model of `AdaptiveComponent` without some `ConversionSkippedException` hacks which we are hesitant to introduce. (related to https://github.com/specklesystems/speckle-sharp/pull/2915)

See discussion https://www.notion.so/speckle/Nested-Elements-and-Traversal-4bb6f8e9c6e14bd0823486dc09571554#55bdfddca94041fe9dfabcf545066a67

We should re-scope the requirements for this issue, determine if `AdaptiveComponent` should be re-written with inspiration from `Instance` or if other tooling should be created

Issue #2829, opened 2023-08-03
bug(Revit) Sending linked models from view doesn't work


paloknapo:
Error pops up and sending fails when sending linked models with view selection.
![image](https://github.com/specklesystems/speckle-sharp/assets/760452/91eb8115-7a8d-4ef5-a114-47f7d8341779)

## To replicate
1. Open [this test model](https://www.notion.so/speckle/Linked-Model-Project-1-2702975de7d943c09c4b58c0b53dd428?pvs=4) 
2. Set selection filter to "3D View 1"
3. Activate "Send linked models" in advanced settings 
4. Send - fails with error

[Error log in Seq](https://seq.speckle.systems/#/events/?filter=@Id%20%3D%20%27event-aebceab0941908db7940047300000000%27&showExpanded&permalinkId=permalink-300)

teocomi:
Related to #2823 , we're essentially not dealing with linked models well.

paloknapo:
It is possible to set visibility of linked models individually for each Revit view. Sending linked models doesn't need to be supported until the #2823 issue is resolved but user should be able to send a view without any errors. 


![Image](https://github.com/specklesystems/speckle-sharp/assets/760452/e1e4a977-4c23-4ea1-9c43-25449b9ee974)

PR #2870, opened 2023-08-23
feat(revit): support for linked models with the view filter

teocomi:
Closing: https://github.com/specklesystems/speckle-sharp/issues/2829

- adds support for sending only visible elements inside linked models when sending by view 
- NOTE: the above is only available for Revit 2024 onwards, for previous versions we're either sending everything inside the linked model or nothing (if the linked model instance is hidden in the current view)
- - cleans up a bit the existing LinkedRevitInstance logic across our selection filters
- removes the unused `param` selection filter

TODO: update the docs

Tested using our sample model for linked files:

The doors and walls are inside a linked model, and the connector successfully sent the linked model without doors.

image

image


connorivy (approved):
Seems to work as intended. I had some thrown exceptions when my Revit document referenced a linked model that couldn't be found on my machine, but I added some null checks which fixed that.

PR #2875, opened 2023-08-25
feat(revit): receive GIS topography

KatKatKateryna:
## Description & motivation
Receive Topography from QGIS as a correct Revit Topography category

## Changes:
- Adding GisTopography class
- Adding a conversion from GisTopography to RevitTopography
- Adding a function to clean Speckle Mesh from duplicate points as required for RevitTopography (only detects (x,y,z) duplicates, not (x,y) )


KatKatKateryna:
To test: https://speckle.xyz/streams/17b0b76d13/commits/d0dc6b87a3

teocomi:
Can we just use or extend the https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Objects/BuiltElements/Topography.cs class instead?

KatKatKateryna:
Class extension is ok, should the Partial class ConvertGisTopography also merge with ConvertTopography?

Objects/Objects/GIS/GisTopography.cs
```diff
@@ -0,0 +1,22 @@
...
+    public int band_count { get; set; }
+    public List<string> band_names { get; set; }
+    public float x_origin { get; set; }
+    public float y_origin { get; set; }
+    public int x_size { get; set; }
+    public int y_size { get; set; }
+    public float x_resolution { get; set; }
+    public float y_resolution { get; set; }
+    public List<float> noDataValue { get; set; }
+    public List<Mesh> displayValue { get; set; }
```
AlanRynne, line 20: The usage of detachment here is wrong ;) We should have a look at this carefully and pick which ones are chunkable/detachable

Objects/Objects/GIS/GisTopography.cs
```diff
@@ -0,0 +1,22 @@
+using System.Collections.Generic;
+using Objects.BuiltElements.Revit;
+using Objects.Geometry;
+using Speckle.Core.Models;
+
+namespace Objects.GIS
+{
+  public class GisTopography : Base
```
AlanRynne, line 8: As suggested by Matteo, this may be better to inherit from `Topography`, in which case there will be no need for a specific conversion of GisTopography right now.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertTopography.cs
```diff
@@ -29,10 +30,15 @@ public ApplicationObject TopographyToNative(Topography speckleSurface)
...
-          var point = new Geometry.Point(displayMesh.vertices[i], displayMesh.vertices[i + 1], displayMesh.vertices[i + 2], displayMesh.units);
-          pts.Add(PointToNative(point));
+          // add a check for duplicate points, if 'keepXYDuplicates' is false 
+          var ptTuple = (displayMesh.vertices[i], displayMesh.vertices[i + 1]);
+          if ( keepXYDuplicates || !pointTuplesList.Contains(ptTuple) )
+          {
+            pointTuplesList.Add(ptTuple);
+            var point = new Geometry.Point(displayMesh.vertices[i], displayMesh.vertices[i + 1], displayMesh.vertices[i + 2], displayMesh.units);
+            pts.Add(PointToNative(point));
+          }
```
AlanRynne, line 40: This works right now only because we're using [`Topography.Create`](https://www.revitapidocs.com/2022/824b6a30-f775-149f-2e6a-5198c5299d14.htm) with the overload that doesn't accept triangulation (see line 47 of this file).

There is an overload [that does accept triangulation](https://www.revitapidocs.com/2022/5fa47ddd-7dbd-ae8a-69bc-9e260ec56d99.htm), and has been brought up in the forum in the past, but removing duplicates will screw around with triangulation too.

Maybe out of scope for this PR but felt it necessary to mention this.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertTopography.cs
```diff
@@ -10,7 +10,7 @@ namespace Objects.Converter.Revit
 {
   public partial class ConverterRevit
   {
-    public ApplicationObject TopographyToNative(Topography speckleSurface)
+    public ApplicationObject TopographyToNative(Topography speckleSurface, bool keepXYDuplicates = true)
```
AlanRynne, line 13: Is there a reason why we should be "keeping XY duplicates" in the Revit Converter? Revit doesn't allow for multiple points to coincide in XY so this seems redundant.

Either we do allow for duplicates (and then Revit conversion will fail) or we clean up all meshes with duplicates always to ensure they succeed.

Modifying duplicate points will affect existing triangulation (see comment below)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertGisTopography.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+      var meshes = new List<Geometry.Mesh>();
+      foreach (Geometry.Mesh displayMesh in gisTopography.displayValue)
+      {
+        meshes.Add(displayMesh);
+      }
+
+      var speckleTopography = new Objects.BuiltElements.Topography()
+      {
+        applicationId = gisTopography.applicationId ??= Guid.NewGuid().ToString(),
```
AlanRynne, line 26: This seems odd, why do we need an invented `applicationId` here? This property can be null and is only intended to store the original id of the application this was sent from.

If it had no id, that is also ok. I think our code is prepared to use the `Base.id` if `Base.applicationId` is `null` for identification purposes.

AlanRynne (commented):
Added a couple comments to this 🙌🏼

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertTopography.cs
KatKatKateryna: The 'keepDuplicates' parameter was introduces to keep the current behavior: create Revit Topography from incoming object, fail if there are duplicates, and default to DirectShape. This parameter is set to False only for GisTopography, because I know for sure the points will be duplicated multiple times (for coloring purposes). And as Revit Topography doesn't care about vertex color, they can be removed. BUT for the rest of cases I would still keep the vertices without cleaning (like now), so if there is a complex uncompatible mesh, it will fallback to DirectShape. 

This distinction is also a reason to separate GisTopography, but we can potentially find a better way

Objects/Objects/BuiltElements/Topography.cs
```diff
@@ -46,3 +47,10 @@ public RevitTopography([SchemaMainParam] Mesh displayMesh, List<Parameter> param
     public Base parameters { get; set; }
   }
 }
+
+namespace Objects.GIS
+{
+  public class GisTopography : Topography
+  {
+  }
+}
```
AlanRynne, line 56: I'm aware some other files have mixed namespaces... I think what you had before (Objects/GIS/GisTopography) was great so no need to undo that, just change it's inheritance from `Base` to `Topography` like here.

We'd also want the typed properties in the class (band_count, band_names... etc) even if we're not currently using them for anything in revit. They should still exist ;)

Objects/Objects/GIS/GisTopography.cs
AlanRynne, line 1: I would argue we should bring this file back (see comment above) exactly as it was, just inheriting from `Topography` instead of `Base`.

If we inherit from `Topography` we can also remove the `displayValue` property (as topography already has it)

AlanRynne (changes requested):
Just added 2 quick comments! @KatKatKateryna if you need my help sorting out the triangulation thing we talked about let me know 😉

AlanRynne (approved):
This looks great! Thanks @KatKatKateryna!!

Issue #2872, opened 2023-08-24
feat(rhino): master issue for improving revit > rhino workflow (P1/P2)


clairekuang:
This is the master issue for tracking P1/P2 improvements to receiving Revit models in Rhino.

- [x] [Create Layers for Nested Elements](https://www.notion.so/speckle/Create-Layers-for-Nested-Elements-ad63fe14110d4967ba8c0bb979da8255?pvs=4) => #2876
- [x] [Confirm Instances are received with Parameters](https://www.notion.so/speckle/bug-rh-Instances-don-t-have-Parameters-22254d698ff444d38de06aa3f81ac2f7?pvs=4) and [Parameters have Meaningful Names](https://www.notion.so/speckle/User-Friendly-Parameter-Names-c39ae1545983484b8a1057f10a2fe956?pvs=4) =>  #2876
- [x] [Remove Duplicate Hidden Geometry](https://www.notion.so/speckle/No-Duplicate-Geometry-59af03adab93462eaf49f57971cd96be?pvs=4) => #2876
- [x] [Clean Mesh Setting](https://www.notion.so/speckle/Setting-Clean-Meshes-NGON-1990e96d34a0472dab402d501b973b74?pvs=4) => #2881
- [x] [Receiving Levels](https://www.notion.so/speckle/RESEARCH-Revit-Levels-in-Rhino-7118abcc7e1a4af1b6cd170364baa93e?pvs=4) => #2883

PR #2876, opened 2023-08-28
feat(rhino): nested elements layers and parameters improvements when receiving revit commits

clairekuang:
## Description & motivation

Includes the following improvements:
- When receiving commits with nested elements, the rhino connector now places those nested elements in a top-level layer named by their `category` property when previously they were placed under their host element's layer.
- Removes locked and/or hidden objects when receiving on update mode, which was previously resulting in duplicates
- Adds parameters to instances and nested instances, and uses the Parameter `name` property as part of the user string key for legibility

This is part of #2872 

## Changes:

## Screenshots:

image
image


AlanRynne (approved):
This is looking good! Great addition @clairekuang 🙌🏼

I found no gotchas on the implementation either.

PR #2881, opened 2023-08-31
feat(rhino): adds setting to merge all coplanar faces when receiving meshes

clairekuang:
## Description & motivation
Adds a setting on receive in `Rhino7` to merge all coplanar faces on meshes. This includes some refactoring of how we're handling settings in Rhino.

**Note** includes an update of Rhino 7 RhinoCommon and RhinoWindows packages from version `7.4.21078.1001` to `7.10.21256.17001` as the mesh `MergeAllCoplanarFaces()` method was introduced in 7.9.

Part of #2872

## Changes:

## To-do before merge:

- [ ] Update mesh settings in GH on send to use the renamed slug: `sendMeshSetting` and to take a string instead of enum value

## Screenshots:

Before:
image

After (with setting enabled):
image


Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.Geometry.cs
```diff
@@ -724,22 +724,19 @@ public RH.Mesh MeshToNative(Mesh mesh)
     }
     m.Faces.CullDegenerateFaces();
 
-    return m;
-  }
-
-  private bool HasInvalidMultiplicity(RH.NurbsCurve curve)
```
clairekuang, line 730: This method had no references, so I removed it

AlanRynne: Perfect🙌🏼

AlanRynne:
Ok, I just rolled back some changes that Grasshopper was depending on. The way our components are coded requires a bit of rethinking of the current settings in general, potentially the creation of a shared settings class between connector and converter.
For now, keeping the existing grasshopper logic as fallback is non-breaking with what Rhino expects, so I went for that to not keep this PR stuck.

I'll open a new issue for this particular change to be made in the Converter and align theGrasshopper connector with the Settings as a dictionary.

PR #2879, opened 2023-08-29
fix(Revit): Corrects Affine and Offset Unit Conversions. Closes #2878

jsdbroughton:
## Description & motivation

Closes #2878 

## Changes:

- Units.cs changes the cached index for Parameter Conversions to type-value from type

The Conversion is performed for all new type:value combinations as not all conversions are scale factored.

Nominally this is less efficient than keying by typeId alone, but my crude benchmarking suggests no real effect.

## Validation of changes:

Comparison of UI values with values within Speckle

- [x] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x] My commits are related to the pull request and do not amend unrelated code or documentation.
- [x] My code follows a similar style to existing code.
- [ ] I have added appropriate tests.
- [ ] I have updated or added relevant documentation.

PR #2882, opened 2023-08-31
Fix(Revit) : fix units that don't scale linearly

connorivy:
## Description & motivation

See description here https://github.com/specklesystems/speckle-sharp/pull/2879 for problem.

Benchmarking the above solution led to finding a very slight slow down. On small models it is unnoticable. Here are some average results that I found from benchmarking the solution in the above PR (this cached value solution) and this solution (the type check solution)

Cached value solution : ~ 57.7s
image

Type check solution : ~ 55.3s
image

It is interesting to me that the recorded time difference between these two functions is ~ 100 ms (or maybe ticks?), but the time difference of the send operations was consistantly ~2s. This may be because of the extra heap allocations and garbage collections, but that is just my guess. The difference is small enough that I wondered if this PR is really necessay, but it's an easy implementation so why not? I also figure that we can eventual add more unit types that scale linearly which would improve the performance even more.

EDIT:
Turns out caching the revit convertFromInternal method IN THIS LOCATION is not saving much time for us at all, so we're just reverting back to using the Revit method.

## Changes:

- check if units are length type
  - if so, use previous method
  - if not, use new method that caches each individual value

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ForgeTypeIdExtensions.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ForgeTypeIdExtensions.cs
@@ -1,5 +1,6 @@
 #nullable enable
 #if !REVIT2020
+using System.Collections.Generic;
 using System.Linq;
 using Autodesk.Revit.DB;
 using DB = Autodesk.Revit.DB;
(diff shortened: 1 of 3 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/Units.cs
```diff
@@ -202,9 +203,20 @@ public static double ScaleToSpeckleStatic(double value, ForgeTypeId forgeTypeId)
...
-        .GetOrAdd(forgeTypeId.TypeId, () => UnitUtils.ConvertFromInternalUnits(1, forgeTypeId), out _);
+      if (forgeTypeId.IsLengthType())
+      {
+        return value * cache
+          .GetOrInitializeEmptyCacheOfType<double>(out _)
+          .GetOrAdd(forgeTypeId.TypeId, () => UnitUtils.ConvertFromInternalUnits(1, forgeTypeId), out _);
+      }
+      else
+      {
+        var cacheKey = $"{forgeTypeId.TypeId}_{value}";
```
jsdbroughton, line 214: Could you benchmark the halfway house vs. not implementing a cache at all?

It's a shame that there isn't a check enum on the `forgeTypeId` that says whether ConvertFromInternals is linear. I was partly against the building yet another area where we have conditionals based on Revit internals - even more so when it relates to a black-box API call results.

Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ForgeTypeIdExtensions.cs
```diff
@@ -24,7 +25,24 @@ public static class ForgeTypeIdExtensions
...
+    private readonly static HashSet<string> lengthTypes = new()
+    {
+      UnitTypeId.Millimeters.ToUniqueString(),
+      UnitTypeId.Centimeters.ToUniqueString(),
+      UnitTypeId.Meters.ToUniqueString(),
+      UnitTypeId.Inches.ToUniqueString(),
+      UnitTypeId.Feet.ToUniqueString(),
+      UnitTypeId.FeetFractionalInches.ToUniqueString(),
+    };
+    public static bool IsLengthType(this ForgeTypeId forgeTypeId)
```
jsdbroughton, line 39: If this is the conditional flag to switch on I'd prefer to flag IsLinearConversion.

There are lots of non Length units that will be linear conversions.

jsdbroughton (approved):
Can confirm unusual unit conversions are accurately represented.

PR #2884, opened 2023-09-01
feat(rvt): Adds Block to Family conversion

### LINKED ISSUE #2827: Blocks > Revit family
Investigate and eventually implement support for receiving AutoCAD and Rhino blocks in Revit as families.

AlanRynne:
Fixes #2827 

This PR changes the way Blocks are received in Revit. Now, each top-level block that is found while traversing will be converted into a Revit Family instance of it's same name.

### Block conversion to family

The process works as follows:

- If a family already exists with the same name as the block, that family will be used (regardless of the geometry of the receiving block). i.e. updating a family geometry is not supported
- If a family with the same name does not exist in the document, one will be created. No sub-families are created at the moment, all the block geometry will be transformed to the top-level and converted into a single family.

The created families are `Work Plane-Based` families to allow for 3D positioning of the instance accurately, including mirroring. **Scaling of families is not supported as this is a design-limitation of the revit API**.

Sub-families, as well as family geometry updating and parameter creation/assignment, can be done in further cycles.

### New Block -> Family mapping

A new mapping is now also available for Rhino when a Block instance is selected. **New Family Instance** mapping allows to assign a category to the Block Definition that will be respected when creating the family (note that an existing family with the same name will always take precedence, category will only apply on the first family creation.

The mapping is done by introducing a new wrapper class that will be assigned to the `speckleSchema` property. This new wrapper class also has corresponding conversion routines in Revit. **This is not ideal**, but it was the best we can do with the current Mapping API, as it assumes all mappings will attach to `speckleSchema` and does not allow for customization of the mapped object itself.
The ideal solution would have been to add a `category` property to the BlockDefinition, which is in essence what this new wrapper conversion in Revit does.

### Transform bug

This PR fixes a bug in the `Transform.Decompose` method, where only positive scaling was being considered, so mirroring was not correctly reflected on certain conditions.

### Material bug

Fixes a bug in the fetching of Block instance geometry where it would be missing the materials
Materials are now correctly propagated.


Objects/Objects/Other/MappedBlockWrapper.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Objects.BuiltElements.Revit;
+using Speckle.Core.Models;
+
+namespace Objects.Other;
+
+public class MappedBlockWrapper : Base
```
AlanRynne, line 6: Name pending review 😅

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -468,11 +476,11 @@ public object ConvertToNative(Base @base)
         case ApplicationObject appObject:
           if (appObject.Converted.Cast<Element>().ToList() is List<Element> typedList && typedList.Count >= 1)
           {
-            receivedObjectsCache.AddConvertedObjects(@base, typedList);
+            receivedObjectsCache?.AddConvertedObjects(@base, typedList);
           }
           break;
         case Element element:
-          receivedObjectsCache.AddConvertedObjects(@base, new List<Element> { element });
+          receivedObjectsCache?.AddConvertedObjects(@base, new List<Element> { element });
```
AlanRynne, line 483: Had to null check these as there may not be a cache initialized

AlanRynne (commented):
Just added a couple of important things that are noteworthy

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertBlock.cs
```diff
@@ -1,190 +1,403 @@
...
+  /// <summary>
+  ///   Creates a new family document as a duplicate of the specified template.
+  /// </summary>
+  /// <param name="name">The name of the template to use to create the family document.</param>
+  /// <returns>A new family document based based on the specified template.</returns>
+  /// <exception cref="System.IO.FileNotFoundException">When the file corresponding to the provided template name could not be found</exception>
+  /// <exception cref="Autodesk.Revit.Exceptions.InvalidOperationException">When the document could not be opened</exception>
+  private DB.Document CreateNewFamilyTemplateDoc(string name = "Block")
+  {
+    var templatePath = GetTemplatePath(name);
```
teocomi, line 278: I think we need to add these new rvts to the installer as well!

AlanRynne: Oh! So true.... will do so now

Objects/Objects/Other/MappedBlockWrapper.cs
AlanRynne: I chose this as it was the more explicit of names i could find: It defines a block that has been mapped.

I thought about `NewRevitFamily` but decided against it as it will not **always** create a new family. A new family will only be created if no other family matches by name.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertBlock.cs
AlanRynne: Not necessary, all templates are grabbed with a wildcard
https://github.com/specklesystems/speckle-sharp-ci-tools/blob/fd65d9a7595cddafb63154e5af5dacfc5f2384b8/revit.iss#L82

Objects/Objects/Other/MappedBlockWrapper.cs
teocomi: No strong opinions, in my head something like `NewRevitFamily` `RevitFamilyBlock` better communicates the intent of that class, but I'll leave the final decision to you ;)

clairekuang: Can we subclass a `RevitMappedBlockWrapper` on the `MappedBlockWrapper` class, and move the `category` prop to this revit subclass with the revit category default? Just for clarity in our Objects kit, and in case we need to reuse the `MappedBlockWrapper` class in non-revit block mapping workflows

PR #2886, opened 2023-09-04
Revit/unnecessary levels

jozseflkiss:
## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
@@ -50,6 +50,8 @@ namespace Objects.Converter.Revit
 
       var speckleRevitColumn = speckleColumn as RevitColumn;
 
+      double elevationOffset = 0.0;
+      double elevationTopOffset = 0.0;
       var levelState = ApplicationObject.State.Unknown;
       if (speckleRevitColumn != null)
       {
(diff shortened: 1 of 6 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
```diff
@@ -151,10 +153,14 @@ public ApplicationObject ColumnToNative(Column speckleColumn)
...
         //don't change offset for slanted columns, it's automatic
         if (!isLineBased)
-          SetOffsets(revitColumn, speckleRevitColumn, level, topLevel);
+          SetOffsets(revitColumn, level, topLevel, ScaleToNative(speckleRevitColumn.baseOffset, speckleRevitColumn.units), ScaleToNative(speckleRevitColumn.topOffset, speckleRevitColumn.units));
 
         SetInstanceParameters(revitColumn, speckleRevitColumn);
       }
+      else
+      {
+        SetOffsets(revitColumn, level, topLevel, -baseOffset, -topOffset);
```
connorivy, line 162: I think we want to add a check here for if (!isLineBased) similar to how we're doing it on line 155. We don't want to call this method for lineBased (aka slanted) columns

connorivy (changes requested):

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertColumn.cs
jozseflkiss: Yes, you are right. The column import didn't work, because there were no baseLine-s exported from Archicad. Now it is added (but only the vertical case, I'm about to implement the slanted cases, but in a separate PR).
Because all the Columns coming from not Revit are base line defined, I just simple removed the code you are referring. I tested, the columns top and bottom elevations are set correctly.

connorivy (approved):
Thanks Jozsef, looks good

PR #2889, opened 2023-09-05
fix(rvt): Fix static usages of Doc inside converter

AlanRynne:
Minor clash between #2868 and #2884

Revit Converter.Doc was made static by accident, switched back by me but meanwhile Jonathon added some static functions that used the doc.
All static functions using Doc have now been reverted to non-static.


AlanRynne:
Side-stepping review. This is a necessary change

Issue #2468, opened 2023-04-21
bug(Revit): Receive objects as Direct Mesh not working as expected


paloknapo:
When importing as DirectShapes some elements are missing.

- nested elements
- groups
- trees
- furniture
- site elements

### Test Model 
Revit House loaded with Speckle Connector as DirectShape is not visually the same as in Speckle viewer. 

Test model: rac_basic_sample_project.rvt
Commit: https://speckle.xyz/streams/9ff64f51e6/commits/d925c44f18

Missing objects in the test model: 

- [planting](https://speckle.xyz/streams/9ff64f51e6/objects/09ec296fd9eba79efaee1adc5e638eeb) - family: RPC Tree
- [PV panels](https://speckle.xyz/streams/9ff64f51e6/objects/671173efcc176534b76e96f963cf2710) - nested groups with electrical equipment, family: Photovoltaic-Panel-…, type: SunModule SW 245 Silver Mono - 10 Deg. Angle
- [site](https://speckle.xyz/streams/9ff64f51e6/objects/f6065e382d02928e45134350045d95cd) - family: M_Wind Power Generator…
- [entourage](https://speckle.xyz/streams/9ff64f51e6/objects/0a9c0aef0370e652410d2c2ae82feea4) - family: RPC Male…
- [furniture](https://speckle.xyz/streams/9ff64f51e6/objects/4d3a29fce7d4c86627c6e4f8a77da58e) - family: Seating - Artemis …
- [generic](https://speckle.xyz/streams/9ff64f51e6/objects/e3fd9a4d9522a173452fda14249972c2) - fireplace…
- all windows missing
- curtain walls received as walls

![Untitled](https://github.com/specklesystems/speckle-sharp/assets/760452/d13b1938-3573-4c5c-927f-a727b1ed3759)

### Desired Behavior:
- when receiving a model with this option enabled, all the meshes also visible in the viewer should be created in Revit with the right category associated to them
- any received model has to match its appearance in Speckle viewer and connector preview

### Extra:
- Rename the option to "Receive objects as DirectShape"

paloknapo:
@clairekuang Some objects come as Generic models, for example:
- stairs
- railings
- windows
- furniture
When receiving as DirectShapes, objects should come with the correct category. 
![image](https://github.com/specklesystems/speckle-sharp/assets/760452/12ed2eff-dc49-4e1f-99ce-812e7f8db880)

Tested with 2.16 wip3

PR #2890, opened 2023-09-05
fix(revit): directshape fallback receive category assignment

clairekuang:
## Description & motivation
 When receiving an element as a `displayable` object, first test to see if that element has a valid `RevitCategory` prop before assigning default revit categories.

Part of #2468

Also includes minor refactor of rhino connector and mesh settings bug fix, and added 2024 projects to the acadcivil solution filter

## Screenshots:

Before:
image

After:
image


AlanRynne (approved):
🙌🏼 Great, thx!

PR #2898, opened 2023-09-07
fix(rhino): various bugs when receiving

clairekuang:
## Description & motivation

Fixes the following issues during receiving:
- Revit parameters with null values are now being handled correctly: previously this was throwing a null exception and preventing receives
- Receiving on a different mesh setting now updates all objects properly: previously, when re-receiving the same commit with a different mesh setting, the previously converted mesh objects were not being re-converted
- Now traverses and receives `Views` and `Levels` on revit commits without any geometry, by treating all `model` type collections as not convertible

## Changes:

rhino connector

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
@@ -37,7 +37,7 @@ public partial class ConnectorBindingsRhino : ConnectorBindings
     converter.ReceiveMode = state.ReceiveMode;
 
     // set converter settings
-    bool settingsChanged = CurrentSettings == state.Settings ? false : true;
+    bool settingsChanged = CurrentSettings != state.Settings;
     CurrentSettings = state.Settings;
     var settings = GetSettingsDict(CurrentSettings);
     converter.SetConverterSettings(settings);
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
@@ -335,7 +335,7 @@ public partial class ConnectorBindingsRhino : ConnectorBindings
       ApplicationObject NewAppObj()
       {
         var speckleType = current.speckle_type
-          .Split(new[] { ':' }, StringSplitOptions.RemoveEmptyEntries)
+          .Split(new[] { '.' }, StringSplitOptions.RemoveEmptyEntries)
           .LastOrDefault();
 
         return new ApplicationObject(current.id, speckleType)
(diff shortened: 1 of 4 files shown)


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -335,7 +335,7 @@ ApplicationObject CreateApplicationObject(Base current, string containerId)
       ApplicationObject NewAppObj()
       {
         var speckleType = current.speckle_type
-          .Split(new[] { ':' }, StringSplitOptions.RemoveEmptyEntries)
+          .Split(new[] { '.' }, StringSplitOptions.RemoveEmptyEntries)
```
AlanRynne, line 338: Just want to confirm that this is intentional. Didn't this split the speckle_type on `:` so that we'd get the last class entry with it's entire namespace? i.e. `Objects.BuiltElements.Beam:Objects.BuiltElements.Revit.RevitBeam` became `Objects.BuiltElements.Revit.RevitBeam`, but with this change it becomes just `RevitBeam`.

Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.cs
```diff
@@ -679,7 +678,9 @@ public bool CanConvertToNative(Base @object)
       case Level _:
       case Text _:
       case Dimension _:
+      case Collection c when !c.collectionType.ToLower().Contains("model"):
```
AlanRynne, line 681: Can't we check for the `Model` class here instead?

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -37,14 +37,15 @@ public override async Task<StreamState> ReceiveStream(StreamState state, Progres
     converter.ReceiveMode = state.ReceiveMode;
 
     // set converter settings
+    bool settingsChanged = CurrentSettings == state.Settings ? false : true;
```
AlanRynne, line 40: This bool is redundant. `settingsChanged = CurrentSettings != state.settings` would do the trick

AlanRynne (approved):
Code looks fine, just a couple comments/questions but nothing to change if you don't feel they're important or relevant.

Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.cs
clairekuang: Unfortuantely no, `Model` isn't a class / subclass in use currently

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
clairekuang: omg haha good catch

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
clairekuang: yah, some users have given feedback that the namespace is too much info

PR #2899, opened 2023-09-07
chore(revit): update some settings description and dialog message

teocomi:
Clarifies some dialog messages and settings descriptions

image

### DIFF
--- a/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Settings.cs
+++ b/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Settings.cs
@@ -83,18 +83,18 @@ namespace Speckle.ConnectorRevit.UI
         new ListBoxSetting
         {
           Slug = "direct-shape-strategy",
-          Name = "Convert received objects to DirectShape",
+          Name = "Fallback to DirectShape on receive",
           Icon = "Link",
           Values = new List<string> { "Always", "On Error", "Never" },
           Selection = "On Error",
-          Description = "Determines when to attempt conversion of an element into a DirectShape"
+          Description = "Determines when to fallback to DirectShape on receive.\n\nAways: all objects will be received as DirectShapes\nOn Error: only objects that fail or whose types are missing\nNever: disables the fallback behavior"
         },
         new MultiSelectBoxSetting
         {
           Slug = "disallow-join",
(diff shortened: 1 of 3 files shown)


ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Settings.cs
```diff
@@ -109,10 +109,10 @@ public override List<ISetting> GetSettings()
         new MappingSetting
         {
           Slug = "receive-mappings",
-          Name = "Custom Type Mapping",
+          Name = "Missing Type Mapping",
           Icon = "LocationSearching",
           Values = mappingOptions,
-          Description = "Determine how incoming object types are mapped to object types in the host application"
+          Description = "Determines when the missing types dialog is shown\n\nNever: the dialog is never shown\nAlways: the dialog is always shown when types are missing\nFor New Types: the dialog is only shown if there are new unmapped types"
```
connorivy, line 115: Always will ALWAYS show the dialog. Even if there aren't any missing types. 

Use case for doing this is that there aren't any missing types incoming, but someone want to tryout a different floor so they map an existing type to a different existing type. Another reason for always showing it on always is because this is a new feature and there may be edge cases in which the missing incoming types counter will not work 100% correctly.

We could change this to align with what you've described if you don't find the use case compelling.

DesktopUI2/DesktopUI2/Views/Windows/Dialogs/MissingIncomingTypesDialog.xaml
```diff
@@ -14,13 +14,20 @@
 
     <TextBlock
       Grid.Row="1"
-      Margin="15"
+      Margin="15,15,15,5"
+      Foreground="Gray"
+      Text="Some incoming types are missing, would you like to import or map them?"
```
connorivy, line 19: This doesn't necessarily require action, but I wonder if any of our users would not know what it means to "map" types

connorivy (approved):

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Settings.cs
teocomi: Will fix this

PR #2903, opened 2023-09-11
fix(revit): Category not assigned to Blocks>NewFamily

### LINKED ISSUE #2905: Rvt/Rh: Block to Family is missing assigned category when received in Revit
Reported while testing and confirmed to be the case.

Revit API will throw if the category ID we're setting is not allowed. Our own class `RevitCategory` contains all the categories in Revit we support in general, but in this case the allowed selection is even narrower (See last part of issue for screenshots of allowed categories in Revit Families)

image

https://seq.speckle.systems/#/events?signal=signal-m33302&filter=%22Document%20category%20could%20not%20be%20modified%22

The way we do things right now, we're falling back to "Generic Model" if the category assignment fails, but this is not notifying the user that this was the case, as it happens deep within the `Block Instance` conversion (which contains an appObj we could attach reports to)

## Possible solutions:

We have several options, the ideal solution would be _Option 3_, but i'm not sure if its even doable.

A more realistic, achievable approach right now would be _Option 2_, but since we'll still be using `RevitCategory`, I'd suggest we do _Option 1_ also as a preventive measure (notifying the user is always nice)

### Option 1:

Leave the current behaviour, but modify the logic slightly so that we can notify the user (could be done with a small refactor of the logic, making it fail instead of falling back to a different category, and handling the failure upwards in the call stack)

### Option 2:

Restrict the options the mapper provides in Rhino to not allow users to map blocks to non-supported categories

### Option 3:

Find out if there's a way to do this with a different template other than the "Block" one we pack (which derives from the standard Generic Model template)

### Supported Revit Family categories

These are all the architecture related categories
Screenshot 2023-09-12 at 08 11 20
Screenshot 2023-09-12 at 08 10 59

And these are the infrastructure ones (filtered them out in a separate screenshot bc they were a lot)
Screenshot 2023-09-12 at 08 11 11

teocomi:
Fixes #2905  

It turns out the list of catagories that can be used for new family creation is a subset of the BuiltIn ones.
See dialog below. I created a list that matches the one in the Revit UI (manually, as I could not find anything on this in the API) and used that for the Block > New Family Mapper objects:

image

### DIFF
--- a/DesktopUI2/DesktopUI2/ViewModels/MappingTool/Schemas.cs
+++ b/DesktopUI2/DesktopUI2/ViewModels/MappingTool/Schemas.cs
@@ -4,8 +4,8 @@ using System.Linq;
 using System.Runtime.Serialization;
 using Objects.BuiltElements;
 using Objects.BuiltElements.Revit;
-using Objects.Other;
 using Objects.BuiltElements.Revit.RevitRoof;
+using Objects.Other;
 using ReactiveUI;
 using Speckle.Core.Api;
 using Speckle.Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


AlanRynne (approved):
Made some additions on the Revit side of things to ensure the RevitFamilyCategory is used consistently.

Other than that I think we're good to go!

PR #2906, opened 2023-09-12
fix(rvt): Modified revit traversal to prevent traversing displayable objects

AlanRynne:
Revit Traversal is still traversing all elements inside a `displayValue` list, even when we're already considering the parent object to be "displayable", and hence, convertible too.

This leads to duplicated objects when receiving "purely displayable objects" (ones that we can't directly convert), as the traversal will return the parent and all it's children.

Changes in this PR modify the Revit traversal so that it also stops traversing when an object is displayable (Navis objects are always purely displayable)

This could be related to some of the other instance duplication issues we've been encountering.

Kudos to @JR-Morgan for the quick session on traversal details.


teocomi (approved):

clairekuang (approved):
Did some receive testing in revit, looks ok 🤞

PR #2910, opened 2023-09-12
fix(revit): better handling of built-in-categories 

clairekuang:
## Description & motivation

Fixes issues of failed railing creation due to improper sending and retrieval of built in categories for revit 2023 and 2024.

This pr mirror ths logic used by schema builder to remove "OST_" from bic internal names to find the equivalent RevitCategory, when before we were just using the category name. This was causing issues in differentiating between "OST_Railings" and "OST_StairsRailing" bics because they both had the category name of "Railing". Also seems to have fixed some issues in native railing conversion by retrieving the correct type.

## Screenshots:

image

Railings now received on fallback conversion:
image

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Categories.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Categories.cs
@@ -41,6 +41,13 @@ namespace Objects.Converter.Revit
       return Enum.TryParse(builtInName, out bic);
     }
 
+    public static bool GetBuiltInCategoryFromRevitCategory(RevitFamilyCategory c, out BuiltInCategory bic)
+    {
+      var name = Enum.GetName(typeof(RevitFamilyCategory), c);
+      var builtInName = $"OST_{name}";
+      return Enum.TryParse(builtInName, out bic);
+    }
+
     /// <summary>
     /// Returns the corresponding RevitCategory enum from a specific BuiltInCategory
     /// </summary>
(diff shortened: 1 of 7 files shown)


martinromby:
Hi @clairekuang 
Any chance that you could provide a list of category names prior and after the implementation of this PR? 
We have a small whitlisting of category names, so that would be good to update accordingly when this is released. 

Thanks.

AlanRynne (approved):
We discussed this yesterday but forgot to ✅ after our call.

clairekuang:
@martinromby No worries, there are no changes to the actual list of `RevitCategory` enums we support ([link](https://github.com/specklesystems/speckle-sharp/blob/main/Objects/Objects/BuiltElements/Revit/Enums.cs)).

The only difference is that before, some elements were not being sent with the correct enum as their category. Before this PR, railings of built-in type `OST_StairsRailing` were being sent with `category = Railings`, and now they are being sent with `category = StairsRailing`

martinromby:
@clairekuang Thanks for clarifying.
I'm on the latest pre-release, Revit 2024 and same Sample Project as you, but the Railing still seem to be translate to `category = Railings`. 
Not sure if we should update the whitelist with `StairsRailing` or not. 

Here's a commit on the xyz server:
https://speckle.xyz/streams/c1b3bb896b/commits/282d31594b

clairekuang:
@martinromby Are you able to receive this railing in your file, if you set the `Fallback to DirectShape on Receive` setting to `Always`? Just want to check that it's not a bic type difference between your 2024 file and my 2023 file.
image

clairekuang:
@martinromby You should see the updated categories with the next pre-release, after this pr has been merged!

PR #2911, opened 2023-09-12
fix(revit): do not show the missing types dialog if the DS fallback option is set to Always

teocomi:
What the title says


connorivy:
Did you make this PR from the correct branch? This code looks the same (and is the same branch) as this PR https://github.com/specklesystems/speckle-sharp/pull/2899/

PR #2912, opened 2023-09-12
fix(revit): allows converter to skip nested elements on receive

clairekuang:
> ⚠️ **WARNING**
> This PR and #2915 fix the same issue in different ways. Only one of them should be merged

## Description & motivation
In the case of receiving curtain walls, where the parent wall (1) is convertible and (2) has panel and mullion objects inside the `elements` property that **should not** be created, we needed a way to skip the conversion of these elements in the connector.

This implementation includes:
- A new `SkipNestedElementsOnReceive()` method in the converter that creates `ApplicationObject` with status = `skipped` in the converter report
- An extra line in the connector checking for `ApplicationObject` status before sending an object to the converter
- Removes two methods for Hosted elements in the converter utils that didn't have any references

This seems to work without receiving any duplicates when receiving on regular update mode, as well as on the always use direct shape setting.

### DIFF
--- a/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Receive.cs
+++ b/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Receive.cs
@@ -222,8 +222,9 @@ namespace Speckle.ConnectorRevit.UI
         // traverse each element member and convert
         foreach (var obj in GraphTraversal.TraverseMember(nestedElements))
         {
-          // create the application object and log to reports
-          var nestedAppObj = Preview.Where(o => o.OriginalId == obj.id)?.FirstOrDefault();
+          // first try to retrieve this object from the preview and the converter report
+          // create if none exist
+          var nestedAppObj = Preview.Where(o => o.OriginalId == obj.id)?.FirstOrDefault()??converter.Report.ReportObjects[obj.id];
           if (nestedAppObj == null)
           {
             nestedAppObj = new ApplicationObject(obj.id, ConnectorRevitUtils.SimplifySpeckleType(obj.speckle_type))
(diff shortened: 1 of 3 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -28,51 +28,24 @@ public partial class ConverterRevit
   {
     #region hosted elements
 
-    private bool ShouldConvertHostedElement(DB.Element element, DB.Element host)
+    // Creates an ApplicationObject with status = Skipped for each object in the `elements` property
+    // This will tell the connector to skip these elements when traversing 
+    private void SkipNestedElementsOnReceive(Base host)
```
teocomi, line 33: We should allow to pass some rules so that only elements of a specific type would be skipped.
In the case of CurtainWalls only mullions and panels, but not doors and windows.

AlanRynne:
Added a comment on the top to not forget this is option 1 of 2 😉

> ⚠️ **WARNING**
> This PR and #2915 fix the same issue in different ways. Only one of them should be merged

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Receive.cs
```diff
@@ -235,6 +238,10 @@ void ConvertNestedElements(Base @base, ApplicationObject appObj, bool receiveDir
             converter.Report.Log(nestedAppObj);
           }
 
+          //check status, skip if necessary
+          if (nestedAppObj.Status == ApplicationObject.State.Skipped)
```
AlanRynne, line 242: Wouldn't the logic here be "any report object state that is not-unknown"? We wouldn't want this to run if we "Created" or "Updated" children (granted I don't think that happens but still, would be nice to prevent more duplication :)

AlanRynne:
Closing in favor of #2915 as per our conversation yesterday.

PR #2915, opened 2023-09-12
Fix(Revit) : add curtain panel class with ToNative implementation to avoid converting as displayable object

connorivy:
> ⚠️ **WARNING**
> This PR and #2912 fix the same issue in different ways. Only one of them should be merged

In the case of receiving curtain walls, where the parent wall (1) is convertible and (2) has panel and mullion objects inside the elements property that should not be created, we needed a way to skip the conversion of these elements in the connector.

This implementation includes:

- a new RevitCurtainWallPanel object that is just a RevitElement
- implements PanelToNative which just skips converting to native

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
@@ -634,6 +634,9 @@ namespace Objects.Converter.Revit
         case BE.Topography o:
           return TopographyToNative(o);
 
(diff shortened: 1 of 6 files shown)


AlanRynne:
Updated the PR description to clarify this is option 2 of 2

> ⚠️ **WARNING**
> This PR and #2912 fix the same issue in different ways. Only one of them should be merged

AlanRynne:
Set as "Ready for Review" as per our conversation yesterday. @teocomi feel free to check it out and leave any extra comments if need be.

teocomi (approved):

Issue #2918, opened 2023-09-12
Revit: Illegal attempt to modify document. Reason: HandleEvent was not called from UI thread


teocomi:
This exception is thrown when opening the Revit 2024 sample model and as a first thing sending by view to Speckle.
A manual workaround is to open any other view before sending.

![image](https://github.com/specklesystems/speckle-sharp/assets/2679513/34e9be04-c449-4ae0-8f93-d8628e6daf4f)

bimgeek:
Closing this ticket as we are moving to Jira to reduce noise. You can access it there.

PR #2919, opened 2023-09-13
fix(revit): throw better error message

teocomi:
Temporary closing: https://github.com/specklesystems/speckle-sharp/issues/2918

### DIFF
--- a/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
+++ b/ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Autodesk.Revit.DB;
+using Autodesk.Revit.Exceptions;
 using ConnectorRevit;
 using DesktopUI2.Models.Filters;
 using DesktopUI2.Models.Settings;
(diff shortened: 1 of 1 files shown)


ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
```diff
@@ -285,6 +286,10 @@ private List<Element> GetSelectionFilterObjects(ISpeckleConverter converter, ISe
       }
       catch (Exception ex)
       {
+        if (ex is ModificationForbiddenException)
+          throw new SpeckleException($"Modification Forbidden: please open another View and try again.", ex);
+
+
```
AlanRynne, line 292: Only comment I would have here is to catch this in a separate catch clause above this one.

But not a real blocker

AlanRynne (approved):
Approving! Just left 1 comment for your consideration 😉

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Selection.cs
teocomi: Makes sense, txs! 🤦‍♂️

teocomi:
Closed in favour of https://github.com/specklesystems/speckle-sharp/pull/2921

PR #2920, opened 2023-09-13
fix(revit): displayable conversion now replaces existing objects on update receive

clairekuang:
## Description & motivation

Fixes issue where we were getting duplicate objects when using the fallback to directshape setting when receiving.
This now searches for existing elements within the DS conversion not limited to DS objects, and adds a missing applicationID to created DS objects in the ConvertAsDisplayable method.


AlanRynne (approved):
Ohh.... very good catch! Took me a minute to understand the changes i was seeing but it makes total sense.

🙌🏼

Approving to retest on the next alpha (hopefully the last)

PR #2922, opened 2023-09-13
Fix(Revit) : analytical stick and surface fixes

connorivy:
## Description & motivation

Due to changes in our beam, column, floor, and wall conversion, Revit-Speckle elements (RevitWall, RevitFloor, etc.) are now treated differently than the default objects (Wall, Floor, etc.). The revit objects are expected to have level properties and offsets from those levels (which they would if they were actually from Revit).

Therefore instead of creating Revit-Speckle objects to create the physical elements associated with analytical sticks and surfaces, we need to make Generic Speckle objects.

## Screenshots:

Currently receiving on dev
image

Receiving on this branch
image

## Checklist:


AlanRynne (approved):
This makes 100% sense 🙌🏼

Code looks good too so I'm approving this.

PR #2923, opened 2023-09-14
fix(revit): skips category assignment when category already exists on send

clairekuang:
## Description & motivation

Revit tests for `DirectShapes` were failing as reported by @connorivy : 
image

Turns out this was due `DirectShape` having a `category` prop of type `RevitCategory` and not `string` (which is the case for all other revit built elements). This case check was removed in [https://github.com/specklesystems/speckle-sharp/commit/5dc407faaa0cc2a8be30afb65ab59f92e51de307](https://github.com/specklesystems/speckle-sharp/commit/5dc407faaa0cc2a8be30afb65ab59f92e51de307) with the intent of skipping category assignment if a category value already exists, assuming all missing categories would be of type `string`. Pr #2910 subsequently missed the null check and caused DS tests to fail.

Pr adds a type case check back in to make explicit the different types of `category` prop we can expect in our object model, which should also catch the case DirectShapes are being created with no `category` prop assigned

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -251,30 +251,21 @@ namespace Objects.Converter.Revit
 
       speckleElement["worksetId"] = revitElement.WorksetId.ToString();
 
-      // assign the category *if* it hasn't already been assigned (eg, DirectShape conversion should already have assigned the category).
+      // assign the category if it is null
+      // WARN: DirectShapes have a `category` prop of type `RevitCategory` (enum), NOT `string`. This is the only exception as of 2.16.
+      // If the null check is removed, the DirectShape case needs to be handled.
       var category = revitElement.Category;
       var speckleCategory = speckleElement["category"];
       if (speckleCategory is null && category is not null)
       {
         var categoryName = category.Name;
-        switch (speckleElement["category"])
-        {
-          case RevitCategory: // DirectShape as of 2.16 is the only class with `category` prop of type `RevitCategory`
-            speckleElement["category"] = Categories.GetSchemaBuilderCategoryFromBuiltIn(categoryName);
-            break;
-
-          case string:
-            // for the category, we should be mirroring how we handle parsing of built-in-categories for revit 2023/2024
-            // this is because different built-in-categories may have the same name, eg "OST_Railings" and "OST_StairsRailing" both have a Category name of "Railing"
+        // we should use RevitCategory values for BuiltInCategory strings where possible (revit 2023+)
+        // different BuiltInCategory may have the same name, eg "OST_Railings" and "OST_StairsRailing" both have a Category name of "Railing"
 #if !(REVIT2020 || REVIT2021 || REVIT2022)
-            if (Categories.GetRevitCategoryFromBuiltInCategory(category.BuiltInCategory, out RevitCategory c))
-            {
-              categoryName = c.ToString();
-            }
+        if (Categories.GetRevitCategoryFromBuiltInCategory(category.BuiltInCategory, out RevitCategory c))
+          categoryName = c.ToString();
 #endif
-            speckleElement["category"] = categoryName;
-            break;
-        }
+        speckleElement["category"] = categoryName;
       }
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -251,10 +251,11 @@ public void GetAllRevitParamsAndIds(Base speckleElement, DB.Element revitElement
 
       speckleElement["worksetId"] = revitElement.WorksetId.ToString();
 
+      // assign the category *if* it hasn't already been assigned (eg, DirectShape conversion should already have assigned the category).
       // for the category, we should be mirroring how we handle parsing of built-in-categories for revit 2023/2024
       // this is because different built-in-categories may have the same name, eg "OST_Railings" and "OST_StairsRailing" both have a Category name of "Railing"
       var category = revitElement.Category;
-      if (category is not null)
+      if (category is null)
```
clairekuang, line 258: This line is the only change to code

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -251,10 +251,11 @@ public void GetAllRevitParamsAndIds(Base speckleElement, DB.Element revitElement
...
       {
         var categoryName = revitElement.Category.Name;
```
AlanRynne: Is this right? first line in this statement wants `revitElement.Category.Name`, and `category = revitElement.Category` on the line above this one.

This means that getting the category name will throw an error as `revitElement.Category` is null in your changes

AlanRynne: Resolved 1 second after my comment

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -251,19 +251,30 @@ public void GetAllRevitParamsAndIds(Base speckleElement, DB.Element revitElement
...
+        var categoryName = category.Name;
+        switch (speckleElement["category"])
         {
-          categoryName = c.ToString();
-        }
+          case RevitCategory: // DirectShape as of 2.16 is the only class with `category` prop of type `RevitCategory`
+            speckleElement["category"] = Categories.GetSchemaBuilderCategoryFromBuiltIn(categoryName);
+            break;
+
+          case string:
```
connorivy, line 266: We actually don't need this switch. The RevitCategory type is an enum which means it's a non-nullable reference type. So the above statement, `if (speckleCategory is null)` means that this switch case will never hit.

connorivy (changes requested):

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
clairekuang: Yeah I considered this, but included the switch because it makes explicit that we have different types - if we change the logic in the future to handle cases of incorrect default category assignment (we don't currently have a `none` enum, so the 0 value is `abutment` currently), the null check would be removed and this would prevent us from going through this again

connorivy: Hmm, okay. Not sure how I feel about including code that is unreachable. Especially given that it sounds like @AlanRynne and @JR-Morgan have a plan to change the direct shape category from an enum to a string, I don't think we should include this unreachable code just to be explicit that the category can be an enum (when it shouldn't be moving forward). Happy to hear your thoughts on this

clairekuang: If anything I would prefer changing `category` to be it's own `RevitCategory:Category` class, so we can store display names as well as built-in names for revit 😭

clairekuang: Removing the switch in favor of an inline comment for now, in case we do need to make category re-assignments prior to the direct shape category type change

connorivy (approved):

PR #2925, opened 2023-09-14
chore(core): Make client functions consistent with cancellation token position

JR-Morgan:
This PR updates the `Client` functions to accept an optional `CancellationToken` parameter in the last position in the method signature.

This follows the C# convention defined by [CA1068: CancellationToken parameters must come last](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068)
Following this convention improves the DX by 
- reducing the complexity of `Client` (by reducing overloads)
- using a cancellation structure developers are more familiar with
- using a cancellation structure consistent with other functions in Core
- Enables proper use of [`CA2016: Forward the CancellationToken parameter to methods that take one`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2016)
- And for us, we have fewer warnings in core, and in the long run fewer lines of code to manage.

This PR contains no breaking changes.

The current `Obsolete` functions generate warnings for user code.
A few releases down the line these can be made `error: true` before being removed entirely another few releases down the line.

### DIFF
--- a/Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
+++ b/Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
@@ -9,11 +9,11 @@ namespace Speckle.Core.Api;
 
 public partial class Client
 {
-
   /// <summary>
-  /// Cheks if a stream exists by id.
+  /// Checks if a stream exists by id.
   /// </summary>
   /// <param name="id">Id of the stream to get</param>
+  /// <param name="cancellationToken"></param>
   /// <returns></returns>
   public async Task<bool> IsStreamAccessible(string id, CancellationToken cancellationToken = default)
   {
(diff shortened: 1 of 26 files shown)


Core/Core/Api/GraphQL/Client.GraphqlCleintOperations/Client.StreamOperations.cs
```diff
@@ -276,26 +244,17 @@ public async Task<List<Stream>> StreamSearch(CancellationToken cancellationToken
       Variables = new { query, limit }
     };
 
-    var res = await GQLClient.SendMutationAsync<StreamsData>(request, cancellationToken).ConfigureAwait(false);
+    var res = await GQLClient.SendMutationAsync<StreamsData>(request, cancellationToken).ConfigureAwait(false); //WARN: Why do we do this?
     return (await ExecuteGraphQLRequest<StreamsData>(request, cancellationToken).ConfigureAwait(false)).streams.items;
```
AlanRynne, line 248: This is a very good question...

AlanRynne (approved):
This is fantastic. Went through all changes and nothing stood out, other than the `//WARN` comment you added.

gjedlicska (approved):
nice

PR #2926, opened 2023-09-15
fix(revit): type mapping null reference

teocomi:
As reported here: https://www.notion.so/speckle/ImportTypes-Failed-ba359f3d6c8a43ccb0027f5ecce5591e?pvs=4

The `revitDocumentAggregateCache` was never initialized or passed to the `FamilyImporter` class, this caused a null reference exception. Not sure how this was working before or how it broke?


AlanRynne:
I thought I had caught all of these. 🙇🏼‍♂️

In general, we shouldn't assume interfaces that are not part of the constructor will be initialised by the time anybody uses them, and conversions should be able to run without caches of any kind.

I'll get on a fix for this ASAP

AlanRynne (approved):
Approved!

PR #2928, opened 2023-09-18
fix(rvt): Exclude Schedules from Everything filter, only send when category is selected

AlanRynne:
Schedules caused slowness in the sending process, and are usually redundant when sending the entire model.


AlanRynne:
Tested by me and Bilal. Works as expected.

PR #2947, opened 2023-09-19
chore(CSI): Run CSharpier on all CSI related projects

### LINKED ISSUE #2942: Enable CSharpier on ETABS project

AlanRynne:
Fixes #2942

Changes resulting in performing a csharpier run on the `ConnectorCSI` and `Objects/Converters/ConverterCSI/`

Reverted changes made to Core.csproj as it was leading down a CI rabbit hole that had nothing to do with this issue.


JR-Morgan (approved):

Objects/Converters/ConverterCSI/ConverterCSIShared/ConverterCSIUtils.cs
```diff
@@ -19,9 +19,10 @@ public partial class ConverterCSI
     readonly string Send2DResults = "send2DResults";
 
     private string _modelUnits;
+
     public string ModelUnits()
     {
-      if (_modelUnits != null) 
+      if (_modelUnits != null)
         return _modelUnits;
```
BovineOx, line 26: I'd even go with brackets here tbh, for single line - do we at least have line space after?

AlanRynne: CSharpier doesn't remove brackets or any other potentially dangerous changes, just enforces consistent spacing, line-breaks, etc...  

We can also enforce higher level style's such as disallowing single line if statements without brackets with the current setup, but we'd use dotnet format or resharper for that (resharper is waaay better, but not fast on CLI)

Devs already get info messages on their code based on the current preference (which i think is no brackets for single-line statements), but this is not really enforced.

Objects/Converters/ConverterCSI/ConverterCSIShared/ConverterCSIUtils.cs
```diff
@@ -47,7 +50,10 @@ public static List<string> GetAllFrameNames(cSapModel model)
         model.FrameObj.GetNameList(ref num, ref names);
         return names.ToList();
       }
-      catch { return null; }
+      catch
```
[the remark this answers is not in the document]
AlanRynne: Yep.... these null returns are a PITA. But for now I'm happy with consistent line breaks on catch statements 😄

BovineOx (approved):
Have checked to a reasonable level and appears to be all style changes as discussed :)

PR #2949, opened 2023-09-19
chore(core): Removed override ErrorsAsWarnings category

JR-Morgan:
I noticed that the `WarningsAsErrors` group in core.csproj overrides what is defined in the editor build config.
This was not intended so this PR re-enables the proper analysers in core.
This meant handling a few security analysers.


Core/Core/Helpers/Http.cs
```diff
@@ -181,7 +181,9 @@ public static HttpClient GetHttpProxyClient(SpeckleHttpClientHandler? handler =
     IWebProxy proxy = WebRequest.GetSystemWebProxy();
     proxy.Credentials = CredentialCache.DefaultCredentials;
 
-    var client = new HttpClient(handler ?? new SpeckleHttpClientHandler());
+    handler ??= new SpeckleHttpClientHandler();
+    handler.CheckCertificateRevocationList = true;
```
AlanRynne, line 185: What does this do exactly?

AlanRynne (approved):
This looks fine. I'm approving but I do have a question 😄

PR #2954, opened 2023-09-20
hotfix(revit): sets displayValue to empty for curtain walls

clairekuang:
## Description & motivation

Curtain walls were sending with a `null` displayValue property, resulting in their baseline being traversed in the viewer and other connectors. This sets the displayValue to empty to render the wall `displayable` during traversal.

## Changes:

- Revit converter


AlanRynne (approved):
🙌🏼 Nice catch. Kudos to @bimgeek for the discovery.

PR #2957, opened 2023-09-20
chore(core): Adding nullability syntax to API object models

JR-Morgan:
Our graphql api documents explicit nullability on the api models.
I wanted to investigate the potential of using nullability syntax on the sharp side to match the API.

Right now, all I have done here is updated the sharp models with the nullability documented in the API docs.

However, I have left comments on a few that might need some discussion.
- Some values (or list types) that are marked as nullable, but likely the server never responds with a null value.
- Changes of nullability of value types is a breaking change for sharp, so we may want to avoid this unless it's strictly necessary.
    - Ideally I just wanted to decorate props for DX rather than making any changes to functionality/runtime behaviour.
    
This is a low priority item.

Awaiting discussion with Gergo

### DIFF
--- a/Core/Core/Api/GraphQL/Models.cs
+++ b/Core/Core/Api/GraphQL/Models.cs
@@ -1,27 +1,31 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


BovineOx:
@JR-Morgan can someone else review?

PR #2961, opened 2023-09-21
Implemented Traversal Refactor for CSI connectors

### LINKED ISSUE #2938: Investigation: ETABS GraphTraversal should be upgraded
**User Story**
Investigation to support...

  **As a** Speckle engineer
  **I would like** ETABS to use traversal consistent with other connectors
  **In order to** avoid non-standard methods that can lead to bugs, inefficient development etc...

This estimate is the gap analysis between current implementation and what needs to be done. If the gap is found to be small at the end, then we can discuss, document the changes on the ticket and deliver as part of this ticket, otherwise the output should be a new ticket with the technical implementation

JR-Morgan:
Implemented traversal refactor in csi.
Testing Etabs->Etabs and Revit->Etabs shows zero regressions.

See testing report
https://www.notion.so/speckle/CSI-traversal-refactor-f48ad2818696428cb7f559aad1ce4366

closes #2938


BovineOx:
Some *minor comments but I have to confess what is really happening is not going to be clear to me without more context

JR-Morgan:
You're thoughts are very welcome @BovineOx

What I've added in this PR, is effectively a copy paste of the `FlattenCommitObjects` function in Rhino, with the layer behaviour removed.

https://github.com/specklesystems/speckle-sharp/blob/5b0ad24b0f65e3d4a2c959d971bba27977b4fa4a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs#L316-L430

For further context on `GraphTraversal` and `DefaultTraversal` see https://speckle.guide/dev/traversal.html or grab me sometime for a demo.

ConnectorCSI/ConnectorCSIShared/UI/ConnectorBindingsCSI.Recieve.cs
```diff
@@ -144,68 +145,68 @@ private List<ApplicationObject> ConvertReceivedObjects(ISpeckleConverter convert
...
-          Status = ApplicationObject.State.Unknown
-        };
+        if (!StoredObjects.ContainsKey(b.id))
+          StoredObjects.Add(b.id, b);
+      }
 
-        if (converter.CanConvertToNative(@base))
+      ApplicationObject CreateApplicationObject(Base current)
+      {
+        ApplicationObject NewAppObj()
```
AlanRynne, line 163: Am I seeing this right? Is this a local function inside a local function inside a class method? 🥲 😅

AlanRynne (commented):
Just one question before approving

ConnectorCSI/ConnectorCSIShared/UI/ConnectorBindingsCSI.Recieve.cs
JR-Morgan: Not quite,
This is two local functions, inside a class method, side by side, no deeper nesting

JR-Morgan: Wait!! I'm blind,
yes there does appear to be a local function within a local function 😭
More evidence that we need to extract this to a helper class.

AlanRynne: I trust your judgement regarding if this should be "fixed" now, or rather open a different ticket for it and merge this in. Let me know and I'll approve+merge if necessary.

JR-Morgan: This is the same as in  rhino connector (and maybe others too). for now I think it's fine.
It sounds like we might be rewriting some of this for dui3 anyhow.

AlanRynne (approved):
Resolved conversation with Jedd about nested functions. We're good to go!

PR #2963, opened 2023-09-21
feat(autocad/rhino): adds gridline to native conversion

clairekuang:
## Description & motivation

Adds conversions for `Gridline`s as `Curve`s. 
In Rhino, the gridline `label` property is added as the curve `name` as well as two `TextDot`s at the curve start and end. The curve uses a default dashed linetype if no `DisplayStyle` is found.
In AutoCAD, the curve is added with a default dashed linetype if no `DisplayStyle`. The label isn't created in Autocad.

Note: in order for display styles to be accurately represented, gridlines should be sent with a `DisplayStyle` property from Revit and other applications.

## Changes:

- Autocad connector and converter

## Screenshots:

Rhino
image

Autocad
image

### DIFF
--- a/Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.BuiltElements.cs
+++ b/Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.BuiltElements.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Objects.BuiltElements;
 using Objects.Geometry;
+using Objects.Other;
 using Rhino;
 using Rhino.Display;
 using Rhino.DocObjects;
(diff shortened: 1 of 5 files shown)


Objects/Converters/ConverterRhinoGh/ConverterRhinoGhShared/ConverterRhinoGh.BuiltElements.cs
```diff
@@ -206,6 +207,55 @@ public ApplicationObject LevelToNative(Level level)
...
+    // create the curve
+    var curve = CurveToNative(gridline.baseLine);
+    if (curve == null)
+    {
+      appObj.Update(status: ApplicationObject.State.Failed, logItem: "Could not convert curve");
+      return appObj;
+    }
+    // get linetype
+    ObjectAttributes atts = null;
+    if (gridline[@"displayStyle"] as DisplayStyle == null)
```
AlanRynne, line 224: Was this supposed to be `"@displayStyle"`?

AlanRynne (approved):
Perfect! This looks neat @clairekuang! I just have a question regarding a string in the gridline conversion. I think it's ok, but just wanted to be sure this was not a typo.

AlanRynne (approved):

paloknapo:
Tested, works as expected.

PR #2969, opened 2023-09-27
Feat(ETABS) : receive gridlines

connorivy:
## Changes:

- Added DatabaseTableWrapper class
  - many limitations in the CSi api can be overcome by using the Interactive Database Tables. This abstract class provides basic functionality to reading, writing, and baking data into the document using these tables
- ETABSGridLineDefinitionTable
  - There is no api functionality that allows you to create gridLines, so this class is responsible for adding gridLines via the interactive database tables
  - Only works for cartesian gridlines at the moment, but can be adjusted to include polar (which ETABS calls cylindrical) coordinates if we want to.
- ToNativeScalingService
  - This is something that we need in all of our converters, IMO. The scaleToNative function in the converter contains cached scaleFactors and cached document units. This state makes it impossible to call scaleToNative from a class other than the converterCSi class. Therefore I introduced this ToNativeScalingService which is a nice service that only does one thing, scale incoming values to native values.
  - Eventually we would want to replace all "ScaleToNative" methods in the converter with toNativeScalingService.Scale()

## Screenshots:

Revit
image

ETABS
image

## Checklist:


AlanRynne (approved):
Approving after chats with Connor and Pavol to ensure we're all on the same page 🙌🏼

PR #2973, opened 2023-09-29
fix(rvt): Fixes curve extraction method for Ceilings and Floors

AlanRynne:
New curve extraction method for ceilings and floors based on extracting the drawn curve loops from the sketch. Each curve loop will become a polycurve.

This is the same behaviour that Toposolid uses.

Moved `GetSketchCurves` onto the floor file since TopoSolid file is only applied in 2024 and above.

Support for this new extraction logic only applies to >2022, so 2020 and 2021 continue to have the same issue.


connorivy (approved):
tested and looks good

PR #2974, opened 2023-10-02
fix(rhino): default to using speckle id when creating render materials with no name

clairekuang:
## Description & motivation

Uses the speckle id of a `renderMaterial` when no name exists. This was causing many duplicate render materials on receive in rhino.

Note: this needs https://github.com/specklesystems/speckle-server/issues/1806 to be fixed for render materials in ifc uploaded commits to receive well in rhino and other connectors.

## Changes:

- Rhino converter


AlanRynne (approved):
Nice! Just adding here that the other half of this fix is server-side:

https://github.com/specklesystems/speckle-server/pull/1807

PR #2977, opened 2023-10-04
Room height

jozseflkiss:
## Description & motivation

Send correct room height (from Archicad to Revit).

## Changes:

Height property added to Room base class. Archicad Connector sets it.

## Checklist:


AlanRynne (commented):
After a quick chat with @jozseflkiss, we may need to check:

- Should `height` be nullable? Would this result in older walls (which had no `height`) having a null value or a value of `0`
- On Revit, verify that a room with height `0` is something that makes sense and if not, check for that before setting the param. Similar with null, but `TrySetParam` is already failsafe

jozseflkiss:
@AlanRynne In Revit if you set the limit offset (height) of the `Room` to `0`, it doesn't change the height at all. (The property of the class is zero in the debugger, but after closing the transaction it remains unchanged on the UI).
Despite of these, for safety reasons the zero check has been added.

AlanRynne:
Ok then we're good to go!

PR #2987, opened 2023-10-09
feat(revit): send builtInCategories

teocomi:
Reverts: https://github.com/specklesystems/speckle-sharp/pull/2910
Changes the logic to use the BuiltInCategory instead of the `RevitCategory` enum, because more reliable:
- adds a `builtInCategory` prop to every element coming out of Revit
- when falling back to DirectShape first checks if there is a `builtInCategory` prop and uses that, and if not uses the old `RevitCategory` enum

Internal notes: https://www.notion.so/speckle/2-16-Category-Mis-Alignment-8d5a51e672fc4314883eab3af2f02b2d#8ca51ecef5c148ae8cdb48f1b58e4c53

image

### Tests
- send from the sample revit model
- receive in a new doc with direct shape fallback logic set to Always
- railings are received correctly

image

### TODO
in the future we should:
- get rid of the `RevitCategory` enum entirely
- better align the collections names and element category names

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
@@ -3,14 +3,14 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Autodesk.Revit.DB;
+using RevitSharedResources.Extensions.SpeckleExtensions;
+using RevitSharedResources.Helpers;
 using RevitSharedResources.Interfaces;
 using Speckle.Core.Models;
-using OSG = Objects.Structural.Geometry;
 using BE = Objects.BuiltElements;
 using BER = Objects.BuiltElements.Revit;
-using RevitSharedResources.Helpers;
+using OSG = Objects.Structural.Geometry;
 using SHC = RevitSharedResources.Helpers.Categories;
-using RevitSharedResources.Extensions.SpeckleExtensions;
 
 namespace Objects.Converter.Revit
 {
@@ -90,8 +90,14 @@ namespace Objects.Converter.Revit
         return categoryInfo;
       }
 
-      var instanceCategory = @base["category"] as string;
-      if (string.IsNullOrEmpty(instanceCategory)) return categoryInfo;
+      //2.16 onwards we check for "builtInCategory"
+      var instanceCategory = @base["builtInCategory"] as string;
+      //pre 2.16 we used the inconsistent, display value "category"
+      if (string.IsNullOrEmpty(instanceCategory))
+        instanceCategory = @base["category"] as string;
+
+      if (string.IsNullOrEmpty(instanceCategory))
+        return categoryInfo;
 
       var newCategoryInfo = GetRevitCategoryInfo(instanceCategory);
 
@@ -132,19 +138,29 @@ namespace Objects.Converter.Revit
     }
     #endregion
 
+
+
     private IRevitCategoryInfo? GetCategoryInfoForObjectWithExactName(string unformattedCatName)
     {
-      var revitCat = revitDocumentAggregateCache
-        .GetOrInitializeWithDefaultFactory<Category>()
-        .TryGet(unformattedCatName);
+      var bic = BuiltInCategory.INVALID;
+      // 2.16 onwards we're passing the "builtInCategory" string
+      if (unformattedCatName.StartsWith("OST"))
+      {
+        if (!Enum.TryParse(unformattedCatName, out bic))
+          return null;
+      }
+      // pre 2.16 we're passing the "category" string
+      else
+      {
+        var revitCat = revitDocumentAggregateCache
+          .GetOrInitializeWithDefaultFactory<Category>()
+          .TryGet(unformattedCatName);
 
-      if (revitCat == null) return null;
+        if (revitCat == null) return null;
 
-#if REVIT2020 || REVIT2021 || REVIT2022
-      var bic = (BuiltInCategory)revitCat.Id.IntegerValue;
-#else
-      var bic = revitCat.BuiltInCategory;
-#endif
+        bic = Categories.GetBuiltInCategory(revitCat);
+      }
+
 
       var formattedName = CategoryNameFormatted(unformattedCatName);
 
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
@@ -148,7 +148,9 @@ namespace Objects.Converter.Revit
       if (unformattedCatName.StartsWith("OST"))
       {
         if (!Enum.TryParse(unformattedCatName, out bic))
+        {
           return null;
+        }
         formattedName = unformattedCatName.Replace("OST_", "");
       }
       // pre 2.16 we're passing the "category" string
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
@@ -143,11 +143,13 @@ namespace Objects.Converter.Revit
     private IRevitCategoryInfo? GetCategoryInfoForObjectWithExactName(string unformattedCatName)
     {
       var bic = BuiltInCategory.INVALID;
+      string formattedName = "";
       // 2.16 onwards we're passing the "builtInCategory" string
       if (unformattedCatName.StartsWith("OST"))
       {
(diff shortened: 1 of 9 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
```diff
@@ -132,21 +138,31 @@ public IRevitCategoryInfo GetRevitCategoryInfo(string categoryName)
     }
     #endregion
 
+
+
     private IRevitCategoryInfo? GetCategoryInfoForObjectWithExactName(string unformattedCatName)
```
teocomi, line 143: These changes a re a bit less relevant as they touch the type mapper, but I thought to alight its behavior too.

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -67,7 +67,7 @@ private bool ShouldConvertHostedElement(DB.Element element, DB.Element host, Bas
         var elementId = element.Id;
         if (!hostedElementIds.Contains(elementId))
         {
-          extraProps["speckleHost"] = new Base() { applicationId = host.UniqueId, ["category"] = host.Category.Name, };
+          extraProps["speckleHost"] = new Base() { applicationId = host.UniqueId, ["category"] = host.Category.Name };
```
teocomi, line 70: maybe should pass the `builtInCategory` prop here too?

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
```diff
@@ -90,8 +90,14 @@ public IRevitCategoryInfo GetRevitCategoryInfo(Base @base)
         return categoryInfo;
       }
 
-      var instanceCategory = @base["category"] as string;
-      if (string.IsNullOrEmpty(instanceCategory)) return categoryInfo;
```
BovineOx, line 94: Appreciate this one didn't change but recent coding standards request braces for all single line IF statements for readability. also clear line space between conditional blocks, even if the next line contains a comment

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
```diff
@@ -132,21 +138,31 @@ public IRevitCategoryInfo GetRevitCategoryInfo(string categoryName)
...
-        .TryGet(unformattedCatName);
-
-      if (revitCat == null) return null;
+      var bic = BuiltInCategory.INVALID;
+      string formattedName = "";
+      // 2.16 onwards we're passing the "builtInCategory" string
+      if (unformattedCatName.StartsWith("OST"))
+      {
+        if (!Enum.TryParse(unformattedCatName, out bic))
+          return null;
```
BovineOx, line 151: Coding standards, could get repetitive but we will all get used to this.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertDisplayableObject.cs
```diff
@@ -55,45 +55,50 @@ public ApplicationObject DisplayableObjectToNative(Base obj)
...
-          return RevitCategory.GenericModel;
-      }
+      return Categories.GetBuiltInFromSchemaBuilderCategory(category);
+    }
+
+    switch (@object)
+    {
+      case BE.Beam _:
+      case BE.Brace _:
+      case BE.TeklaStructures.TeklaContourPlate _:
```
BovineOx, line 73: I have questions, we are matching the object and within the Revit connector we map Beam, Brace and TeklaStructures to the same built in category? It feels a bit brittle, though I accept we've been doing it so far.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFreeformElement.cs
```diff
@@ -189,18 +189,18 @@ private string CreateFreeformElementFamily(List<DB.Solid> solids, string name, o
...
-              cat = famDoc.Settings.Categories.get_Item(bic);
-              if (cat.SubCategories.Contains(freeformElement.subcategory))
-                cat = cat.SubCategories.get_Item(freeformElement.subcategory);
-              else
-                cat = famDoc.Settings.Categories.NewSubcategory(cat, freeformElement.subcategory);
-            }
+            var bicName = Categories.GetBuiltInFromSchemaBuilderCategory(BuiltElements.Revit.RevitCategory.GenericModel);
+            BuiltInCategory.TryParse(bicName, out bic);
+            cat = famDoc.Settings.Categories.get_Item(bic);
+            if (cat.SubCategories.Contains(freeformElement.subcategory))
```
BovineOx, line 200: {
  // brackets :)
}

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRailing.cs
```diff
@@ -67,6 +67,7 @@ public ApplicationObject RailingToNative(BuiltElements.Revit.RevitRailing speckl
 
         if (topRailType != null && isTopRailExactMatch)
           railingType.TopRailType = topRailType.Id;
```
BovineOx, line 69: you know what I am going to {say} :)

Objects/Objects/BuiltElements/Revit/DirectShape.cs
```diff
@@ -31,6 +31,16 @@ public DirectShape(string name, RevitCategory category, List<Base> baseGeometrie
...
   }
 
+  // moving away from using the RevitCategory Enum
+  public DirectShape(string name, string builtInCategory, List<Base> baseGeometries, List<Parameter> parameters = null)
+  {
+    this.name = name;
+    this.baseGeometries = baseGeometries.FindAll(IsValidObject);
+    this.parameters = parameters.ToBase();
+    //TODO: move to typed property alongside all other revit elements
+    this["builtInCategory"] = builtInCategory;
```
BovineOx, line 41: It would be good if we have hard coded strings to move to some constants, especially if referenced in multiple places

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
teocomi: Agreed, this is someone else's code I removed ;)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRailing.cs
teocomi: Agree, this was pre-existing code tho ;)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertFreeformElement.cs
```diff
@@ -189,17 +189,21 @@ private string CreateFreeformElementFamily(List<DB.Solid> solids, string name, o
...
         {
           //subcategory
+          BuiltInCategory bic;
           if (!string.IsNullOrEmpty(freeformElement.subcategory))
           {
             //by default free form elements are always generic models
             //otherwise we'd need to supply base files for each category..?
-            if (Categories.GetBuiltInCategoryFromRevitCategory(BuiltElements.Revit.RevitCategory.GenericModel, out BuiltInCategory bic))
+            var bicName = Categories.GetBuiltInFromSchemaBuilderCategory(BuiltElements.Revit.RevitCategory.GenericModel);
+            BuiltInCategory.TryParse(bicName, out bic);
```
clairekuang, line 198: handle case where `TryParse` returns false here

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
clairekuang: yes I think so, pass both the `category` and `builtInCategory` props

clairekuang (changes requested):
Looks good, only general comment is that it may be better to store the `builtInCategory` property string in a variable

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRailing.cs
BovineOx: All of this looked changed apart from the first one!

clairekuang (approved):
Tested on 2022 and 2024, looks good - @teocomi 's decision regarding using a variable to contain the `builtInCategory` property string was to unify all hardcoded strings into variables at some later point

PR #2988, opened 2023-10-11
fix(CSI): Fixes missing materials on receive

AlanRynne:
Structural materials where only created when being used within a `Property1D` element.

This PR enhances material support so that:
- Converter has top-level support for `StructuralMaterial` on receive. Meaning `CanConvertToNative` now returns true, and `ConvertToNative` will call `MaterialToNative`. Previously we weren't supporting this use-case
- Property2D was using the material name, but not ensuring the material existed in the model. MaterialToNative is now called and the resulting material name is used when needed.

### DIFF
--- a/Objects/Converters/ConverterCSI/ConverterCSIShared/ConverterCSI.cs
+++ b/Objects/Converters/ConverterCSI/ConverterCSIShared/ConverterCSI.cs
@@ -13,6 +13,8 @@ using Speckle.Core.Models;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Objects.Other;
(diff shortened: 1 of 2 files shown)


AlanRynne:
@teocomi I'm adding you as a reviewer. Connor let me know that currenty, StructuralMaterial has no unit scaling support, which is out of scope for this particular ticket, but should be taken into account for further tickets, as this could lead to inconsistencies at best and, at worst, completely incorrect material values.

Connor would know more about this than me on the subject, just wanted to raise it here, as we're increasing the exposure to creating materials.

connorivy (approved):
LGTM

PR #2989, opened 2023-10-12
feat: check if server is using fe2

teocomi:
Adds info needed to deliver a better UX when using DU2 & FE2:
- adds a (temporary?) `frontend2` prop to the `serverInfo` class
- checks against the response headers to see if a server/account is using frontend 2


AlanRynne (commented):
Looks good! Do we need to unit test this too?

teocomi:
If the test server is using fe2 this could be easily done. Does it serve up the full frontend?

AlanRynne:
The current server setup does not include fe2, so I guess we can't really test this fully yet.

AlanRynne (approved):

PR #2991, opened 2023-10-12
feat(dui): better fe2 compatibility

teocomi:
Depends on: #2989 

This PR improves how DUI2 integrates with FE2:

- moves the FE2 toggle to the main toolbar
- any stream related action (eg view online) now automatically detects if a stream is on fe2 and uses the new routes
- added more string replacements here and there

NOTE: this PR depends on some changes to Manager which are already being rolled out (current beta)

fe2-toggle

### DIFF
--- a/Core/Core/Api/GraphQL/Models.cs
+++ b/Core/Core/Api/GraphQL/Models.cs
@@ -383,6 +383,10 @@ public class ServerInfo
   public string version { get; set; }
   public string adminContact { get; set; }
   public string description { get; set; }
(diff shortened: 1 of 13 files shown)


AlanRynne (changes requested):
I only found one issue, but it can be quite annoying. When adding a stream using the URL, the `model`(branch) that gets selected is always `main` even if the URL pointed to another model (see gif)

AlanRynne (approved):
Nice! Thanks for the fix!

I'll add the StreamWrapper FE2 alignment to our backlog.

PR #2994, opened 2023-10-16
feat(core): Adds SpeckleLogConfig input to Setup.Initialize

AlanRynne:
Opening this PR as a followup to my discord message about this.

This change allows devs to pass in the SpeckleLogConfig of choice when calling `Setup.Initialize` to allow for granular control over what gets logged.

### DIFF
--- a/Core/Core/Logging/Setup.cs
+++ b/Core/Core/Logging/Setup.cs
@@ -40,7 +40,11 @@ public static class Setup
   /// </summary>
   internal static string VersionedHostApplication { get; private set; } = HostApplications.Other.Slug;
 
-  public static void Init(string versionedHostApplication, string hostApplication)
+  public static void Init(
+    string versionedHostApplication,
+    string hostApplication,
+    SpeckleLogConfiguration logConfiguration = null
+  )
   {
     if (_initialized)
     {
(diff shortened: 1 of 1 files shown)


JR-Morgan (approved):

Core/Core/Logging/Setup.cs
```diff
@@ -40,7 +40,11 @@ static Setup()
   /// </summary>
   internal static string VersionedHostApplication { get; private set; } = HostApplications.Other.Slug;
 
-  public static void Init(string versionedHostApplication, string hostApplication)
+  public static void Init(
+    string versionedHostApplication,
+    string hostApplication,
+    SpeckleLogConfiguration logConfiguration = null
```
AlanRynne, line 46: @JR-Morgan file did not have nullability activated, so I decided to leave this input "non-nullable" but defaulting to null....

`SpeckleLog.Init` does take a nullable `SpeckleLogConfiguration?`.

gjedlicska (approved):

Core/Core/Logging/Setup.cs
JR-Morgan: That's fine. In a non-nullable context, this is a Null oblivious type.

PR #2998, opened 2023-10-18
Performance(csi) : commit database tables in single transaction

connorivy:
## Description & motivation

CSi provides the ability to edit the document via the interactive database tables. However, it is quite expensive to modify the document using these tables. Therefore we want to batch the items instead of baking each one individually.

## Changes:

- Add CSiSharedResources project (similar to RevitSharedResources)
- Add csi specific implementation of ISpeckleConverter (ICSiSpeckleConverter) which implements method 'CommitAllDatabaseTableChanges'
- Call this method from the converter after the receive operation is finished

This change is not "kit friendly" as it requires anyone making a new converter to implement an interface that is defined in CSiSharedResources instead of Core which could be confusing, but it still can be done. IMO implementing DI in the connectors will mean that users can implement specific conversions instead of having to replace the entire converter, so I am not so sad about this PR being not kit friendly.

## Validation of changes:

Receiving large amounts of grids still works an is much faster

## Checklist:


AlanRynne (changes requested):
Lets have a chat later! I really don't think we should be merging this as-is.

It seems the only reason `ICsiSpeckleConverter.CommitAllDatabaseTableChanges` exists is to call `ETABSGridLineDefinitionTable.CommitPendingChanges`

It seems to me that `ETABSGridLineDefinitionTable` should be "object agnostic", which would allow us to move it to the SharedResources project. Giving access to this CSI API wrapper to both sides.

This still leaves us with the same issue RevitConverter has in `SetContextDocument`: we'd still need a way to pass this down from Connector->Converter and the current solution followed by Revit is not the best DX.

At least this would be consistent across connectors (Revit, ETABS) which would make our lives easier when we finally get to a better pattern for our converters.

connorivy:
@AlanRynne 
Changes applied per our discussion today

AlanRynne (approved):
Looking good 🙌🏼 I'll approve but I'll let @BovineOx to give it a look before merging too since he suggested this pattern 😉

BovineOx:
I've added myself to the review and will take a look this AM. @connorivy you may want to merge dev back into this branch as GH is complaining of conflicts now

BovineOx:
@AlanRynne I think we gave the all clear on this one didn't we?

BovineOx (approved):
As discussed all fine

PR #3000, opened 2023-10-23
fix(core): Fixed regression with stream collaborators query parameters

JR-Morgan:
In https://github.com/specklesystems/speckle-sharp/pull/2925, I accidently introduced a regression by renaming  `id` -> `streamId` in StreamGetPendingCollaborators

(side note, this demonstrates we are missing an integration test for this query)


AlanRynne (approved):
Good catch!

PR #3009, opened 2023-10-27
chore(objects): corrected nullability to objects in structural namespace

JR-Morgan:
Corrected nullability syntax in objects in structural namespace


connorivy (approved):
LGTM

PR #3019, opened 2023-11-02
fix(rvt): Modify signature of `RevitConverter.UpdateParameter` to return AppObj

AlanRynne:
> ⚠️ Reported by Pavol, pending linking with Notion backlog issue

> ❗️Targeting `main` as I'm assuming we'd want this as a hotfix, but we can target `dev` if this is not the case.

Fixes issue with ParameterUpdater objects from Grasshopper not succeeding conversion when received in Revit.

Nothing was wrong with the conversion itself, but the addition of `DisplayableObject` conversions introduced the assumption that if the conversion returned `null` it had failed.

This assumption did not hold for `ParameterUpdater` conversions, where **it was always returning null** as the function had a `void` return.

### Solution

This PR changes the signature of the UpdateParameter method so that it returns an app object, in line with every other conversion. This results in the connector assuming conversion has succeeded (and correctly reporting it to the user as such)

Screenshot 2023-11-02 at 11 34 50

### EDGE CASE NOT HANDLED

If a user were to set the `receive as display mesh` option, the parameter updater conversion **will fail** as this object is not displayable. I consider this the correct behaviour, but I wanted to leave it here to verify.

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
@@ -43,9 +43,6 @@ namespace Objects.Converter.Revit
       }
       else
       {
-        Report.LogConversionError(
-          new System.Exception($"Could not find element to update: Element Id = {paramUpdater.elementId}")
-        );
         appObj.Update(
           status: ApplicationObject.State.Failed,
           logItem: $"Could not find element to update: Element Id = {paramUpdater.elementId}"
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using Speckle.Core.Helpers;
 using Speckle.Core.Models;
 
-
 namespace Objects.Converter.Revit
 {
   public partial class ConverterRevit
@@ -17,8 +16,11 @@ namespace Objects.Converter.Revit
       //try to get element using ApplicationId
       //this will only work if the element has been successfully been received in Revit before
       //var element = GetExistingElementByApplicationId(paramUpdater.revitId);
-      
-      var appObj = new ApplicationObject(paramUpdater.id, paramUpdater.speckle_type) { applicationId = paramUpdater.applicationId };
+
+      var appObj = new ApplicationObject(paramUpdater.id, paramUpdater.speckle_type)
+      {
+        applicationId = paramUpdater.applicationId
+      };
       Element element = null;
 
       //try to get element using ElementId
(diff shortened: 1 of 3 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
```diff
@@ -3,44 +3,50 @@
...
 
       //try to get element using UniqueId
-      if (element == null)
+      element ??= Doc.GetElement(paramUpdater.elementId);
+
+      if (element != null)
       {
-        element = Doc.GetElement(paramUpdater.elementId);
+        SetInstanceParameters(element, paramUpdater);
+        appObj.Update(status: ApplicationObject.State.Updated, logItem: $"Successfully updated instance parameters for element {paramUpdater.elementId}");
```
BovineOx, line 37: where is this logItem going? I'm a bit meh about "I did this thing and it worked" messages as it bloats logging, happy to discuss

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
```diff
@@ -3,44 +3,50 @@
...
       }
-
-      if (element == null)
+      else
       {
-        Report.LogConversionError(new System.Exception($"Could not find element to update: Element Id = {paramUpdater.elementId}"));
-        return;
+        Report.LogConversionError(
+          new System.Exception($"Could not find element to update: Element Id = {paramUpdater.elementId}"));
```
BovineOx, line 42: I mean, this is kind of interesting....  we call this method passing it an exception. Also, do we have a better exception, if not, can we please? :)

BovineOx (changes requested):
Some minor things to discuss, would prefer more specific exception at the least :)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
AlanRynne: Agreed! Any particular suggestion?

If we need to create a brand new exception for this, where should it live?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
AlanRynne: this gets added to the list of logs for that "application object", and displayed to the user in the report tab.

Screenshot 2023-11-02 at 11 34 50

Our converter/conversion logging is not great, but at least this aligns it with everything else. Not sure how good/bad that is 😅 Happy to discuss this a bit further in a call if needed!

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
BovineOx: I don't know what we currently have, also, where does what we currently have live? Somewhere common is the answer there though

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
BovineOx: If this is current practise then let's continue as is, but it feels like logging policy is needed

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/UpdateParameter.cs
AlanRynne: I've removed this line altogether as it turned out to be unnecessary, flagging this conversation as resolved 😄

BovineOx (approved):
LGTM 🤣

PR #3023, opened 2023-11-02
feat(objects/revit): new `RebarGroup` class and `RebarToSpeckle` conversion in Revit

### LINKED ISSUE #3022: feat(revit/objects): new rebar class and sending rebar from revit
see details in https://www.notion.so/speckle/Rebar-Support-in-Revit-Revit-workflow-c1922ee7c1d0450184b16221c6b939b6?pvs=4

clairekuang:
## Description & motivation

Adds a new `RebarGroup`, `RebarHook`, and `RebarShape` class to `Objects`, designed to support rebar from Revit and Tekla.
- Includes `RevitRebarGroup`, `RevitRebarHook`, and `RevitRebarShape` subclasses for Revit
- Marks existing `Rebar` and subclasses as obsolete

All rebar from Revit are now sent as `RevitRebarGroup`.

Closes #3022

## Changes:

- Objects
- Revit Converter

## Checklist:

- [x] I have updated or added relevant documentation: [https://github.com/specklesystems/speckle-docs/pull/188](https://github.com/specklesystems/speckle-docs/pull/188)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      try
       {
-        var result = RebarFreeFormValidationResult.Success;
-        rebar ??= DB.Structure.Rebar.CreateFreeForm(Doc, barType, host, curveLoops, out result);
-        if (result != RebarFreeFormValidationResult.Success)
-        {
-          appObj.Update(status: ApplicationObject.State.Failed, logItem: "Freeform Rebar could not be created from closed curves.");
-          return appObj;
-        }
+        rebar = DB.Structure.Rebar.CreateFromCurves(
```
BovineOx, line 94: Ha, sorry, I spent 5 minutes considering this as being quite a long signature and then realised this is the Revit AP! Nice going Revit! :)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+          normal,
+          curves,
+          startHookOrientation,
+          endHookOrientation,
+          true,
+          true
+        );
       }
-      else if (openCurves.Count > 0)
+      catch (Autodesk.Revit.Exceptions.ArgumentNullException nullEx)
```
BovineOx, line 109: I wonder if an alias might help readability here, no strong feeling but RevitException.ArgumentNullExecption might be a bit easier to read, and so on...

it's great to catch all the relevant exceptions, but it looks like we treat most the same...

This is probably a good case for catching a generic (preferably Revit) exception and then pattenring matching on the type..  or...

There's also this handy when syntax, which might be quite new and maybe not usable yet (I only just saw it)

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when
https://code-maze.com/csharp-catch-multiple-exceptions/

This should improve the catch readability quite a bit.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
-      if (speckleRevitRebar != null)
-        SetInstanceParameters(rebar, speckleRevitRebar);
+          // set additional params
+          if (speckleRebar.barPositions > 0)
+          {
+            rebar.NumberOfBarPositions = speckleRebar.barPositions;
+          }
 
-      appObj.Update(status: ApplicationObject.State.Created, createdId: rebar.UniqueId, convertedItem: rebar);
+          // deleting instead of updating for now!
```
BovineOx, line 146: do we only offer replace atm? How does that sit with UX for the user, will they expect this?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
 
-    // need to test to see possible types of curves for rebar
-    private void ProcessRebarCurves(Rebar rebar, out List<DB.Curve> open, out List<DB.CurveArray> closed)
+    private RevitRebarGroup RebarToSpeckle(DB.Structure.Rebar revitRebar)
     {
-      var _open = new List<DB.Curve>();
-      var _closed = new List<DB.CurveArray>();
-
-      foreach (var curve in rebar.curves)
+      // skip freeform rebar for now: not supported by RevitRebarGroup class
```
BovineOx, line 158: I don't see this skipping would make it into any report?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      // get type
+      var type = revitRebar.Document.GetElement(revitRebar.GetTypeId()) as ElementType;
 
-    private Base RebarToSpeckle(DB.Structure.Rebar revitRebar)
-    {
-      // get rebar centerline curves using transform
-      var bars = revitRebar.GetCenterlineCurves(true, true, true, MultiplanarOption.IncludeOnlyPlanarCurves, revitRebar.NumberOfBarPositions - 1);
-      var curves = new List<ICurve>();
+      // get the rebar shape
+      var revitShape = revitRebar.Document.GetElement(revitRebar.GetShapeId()) as DB.Structure.RebarShape;
```
BovineOx, line 169: Document is always valid? Can revitRebar be null? Does RebarShapeToSpeckle() accept null if it did not cast for some reason?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      if (revitStartHookId != ElementId.InvalidElementId)
+      {
+        var revitStartHook = revitRebar.Document.GetElement(revitStartHookId) as RebarHookType;
+        speckleStartHook = RebarHookToSpeckle(
+          revitStartHook,
+          revitRebar.GetHookOrientation(0).ToString(),
+          hookBendRadius
+        );
+      }
+      DB.ElementId revitEndHookId = revitRebar.GetHookTypeId(1);
```
BovineOx, line 190: Whitespace :)

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      RevitRebarHook speckleEndHook = null;
+      if (revitEndHookId != ElementId.InvalidElementId)
+      {
+        var revitEndHook = revitRebar.Document.GetElement(revitEndHookId) as RebarHookType;
+        speckleEndHook = RebarHookToSpeckle(revitEndHook, revitRebar.GetHookOrientation(1).ToString(), hookBendRadius);
```
BovineOx, line 195: Not sure I follow why we're turning this into a string

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+        0
+      );
+
+      var centerlines = new List<ICurve>();
+      for (int i = 0; i < revitRebar.NumberOfBarPositions; i++)
+      {
+        // skip end bars that are excluded
+        if (!isSingleLayout)
+        {
+          if (
```
BovineOx, line 220: I'd prefer the () braces start and don't behave like {} in this context as I find it makes it harder to read the if statement, i.e. with the ) at the same indent and immediately before the next {

Do we need both If statements? presumably we can collapse this into one?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+            !revitRebar.IncludeFirstBar && i == 0
+            || !revitRebar.IncludeLastBar && i == revitRebar.NumberOfBarPositions - 1
+          )
+          {
+            continue;
```
BovineOx, line 225: Confess I am not a massive fan of continue as it can make the code harder to read - could the be refactored to check whether the following code SHOULD execute?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+        speckleEndHook = RebarHookToSpeckle(revitEndHook, revitRebar.GetHookOrientation(1).ToString(), hookBendRadius);
+      }
+
+      // get the layout rule - this determines exceptions that may be thrown by accessing invalid props
+      bool isSingleLayout = revitRebar.LayoutRule == RebarLayoutRule.Single;
+
+      // get centerline curves for display value
       RebarShapeDrivenAccessor accessor = null;
       if (revitRebar.IsRebarShapeDriven())
         accessor = revitRebar.GetShapeDrivenAccessor();
```
BovineOx, line 204: lets hoover this up in {} and ensure there is a linespcae after the }

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+        normal = VectorToSpeckle(accessor.Normal, revitRebar.Document);
       }
 
-      var speckleRebar = new RevitRebar();
-      speckleRebar.host = revitRebar.GetHostId().ToString();
-      speckleRebar.type = revitRebar.Document.GetElement(revitRebar.GetTypeId()).Name;
-      speckleRebar.curves = curves;
-      speckleRebar.shapes = revitRebar.GetAllRebarShapeIds().Select(o => o.ToString()).ToList(); // freeform rebar with bent workshop has multiple shapes
+      // create speckle rebar
+      var speckleRebar = new RevitRebarGroup();
```
BovineOx, line 263: could an object initialise work here?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+          break;
+      }
+
+      // get the curves representing the default values of the shape
+      List<ICurve> curves = revitRebarShape
+        .GetCurvesForBrowser()
+        .Select(o => CurveToSpeckle(o, revitRebarShape.Document))
+        .ToList();
+
+      var speckleRebarShape = new RevitRebarShape();
```
BovineOx, line 304: object initializer?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      speckleRebarShape.name = revitRebarShape.Name;
+      speckleRebarShape.curves = curves;
+      speckleRebarShape.rebarType = rebarType;
+
+      GetAllRevitParamsAndIds(speckleRebarShape, revitRebarShape);
+      return speckleRebarShape;
```
BovineOx, line 310: I might be include to add a line of WS before the return so it can be a little easier to see

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+    }
+
+    // rebar hook
+    private RebarHookType RebarHookToNative(RevitRebarHook speckleRebarHook)
+    {
+      var revitRebarHook = RebarHookType.Create(Doc, speckleRebarHook.angle, 10); // using default multiplier, will be set by parameters later
```
BovineOx, line 316: when you say later? You mean this 10 is always overridden? I might still be inclined to create a const here, so we can say DefaultMultiplier or maybe more

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      SetInstanceParameters(revitRebarHook, speckleRebarHook);
+
+      return revitRebarHook;
+    }
+
+    private RevitRebarHook RebarHookToSpeckle(RebarHookType revitRebarHook, string orientation, double radius)
+    {
+      var speckleRebarHook = new RevitRebarHook();
```
BovineOx, line 324: object initialiser?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      speckleRebarHook.angle = revitRebarHook.HookAngle;
+      speckleRebarHook.orientation = orientation;
+      speckleRebarHook.radius = radius;
+      GetAllRevitParamsAndIds(speckleRebarHook, revitRebarHook);
+      return speckleRebarHook;
```
BovineOx, line 329: WS line

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
```diff
@@ -92,13 +86,13 @@ public override void IncludeObject(Base conversionResult, Element nativeElement)
     switch (_commitCollectionStrategy)
```
BovineOx, line 86: Is there only one of these? I cannot see what the switch is adding here?

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
```diff
@@ -146,7 +149,8 @@ private void AddNestingInstructionsForMEPElement(Element nativeElement, List<Nes
   {
     var mepSystemName = GetMEPSystemName(nativeElement);
 
-    if (string.IsNullOrEmpty(mepSystemName)) return;
+    if (string.IsNullOrEmpty(mepSystemName))
+      return;
```
BovineOx, line 153: {} pls :)

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
```diff
@@ -170,9 +174,10 @@ private void AddNestingInstructionsForMEPElement(Element nativeElement, List<Nes
     return element switch
```
BovineOx, line 174: Not finding this terribly readable, might be worth discussing...

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
```diff
@@ -211,7 +216,7 @@ private static string GetCategoryId(Base conversionResult, Element revitElement)
 #endif
```
BovineOx, line 216: Could not discover where the #if began :O

Objects/Objects/BuiltElements/Rebar.cs
```diff
@@ -20,6 +159,17 @@ public class Rebar : Base, IHasVolume, IDisplayValue<List<Mesh>>
 
 namespace Objects.BuiltElements.TeklaStructures
```
BovineOx, line 160: I think it's worth moving these classes to their own files, but.. I appreciate that begs the question, where should they live...  let's chat, maybe we get Alan and Jedd involved in that chat also...

BovineOx (changes requested):
This is great work! It may be a lot of queries, but don't let that worry you - I am still learning a lot, have a lot of Qs and it's not uncommon for there to be some back and forth questions...

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+      };
 
       // skip if element already exists in doc & receive mode is set to ignore
       if (IsIgnore(docObj, appObj))
         return appObj;
 
-      if (speckleRebar.curves.Count == 0)
+      // return failed if rebar shape is null or has no curves
+      var barShape = speckleRebar.shape as RevitRebarShape;
+      if (barShape == null || !barShape.curves.Any())
```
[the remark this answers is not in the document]
clairekuang: yes, but `curves` may be an empty list - without this shape curve there is no way to create the rebar

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,337 @@
...
         appObj.Update(status: ApplicationObject.State.Failed);
         return appObj;
       }
-      var rebarStyle = speckleRevitRebar?.barStyle == "StirrupTie" ? RebarStyle.StirrupTie : RebarStyle.Standard;
-
-      // get construction curves (only works for revit rebar due to need for host)
-      var closedCurves = new List<DB.CurveArray>();
-      var openCurves = new List<DB.Curve>();
-      ProcessRebarCurves(speckleRebar, out openCurves, out closedCurves);
+      var barStyle =
```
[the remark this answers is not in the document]
clairekuang: ah `RebarBarType` is captured in the `RevitRebarGroup.type` property already, and `RebarStyle` in revit is captured by the `RebarType` enum in the Speckle class. I noticed another issue however, in not properly handling unknown types!

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,332 @@
...
+        speckleRebar.shape.rebarType != RebarType.Unknown && speckleRebar.shape.rebarType == RebarType.Standard
+          ? RebarStyle.Standard
+          : RebarStyle.StirrupTie;
 
-      // find some way to test if curves should be converted to curve loops for freeform rebar or not for rebar from curves
-      var curveLoops = closedCurves.Select(o => CurveArrayToCurveLoop(o)).ToList();
-
-      DB.Structure.Rebar rebar = null;
+      // get start and end hooks and orientations
+      var speckleStartHook = speckleRebar.startHook as RevitRebarHook;
```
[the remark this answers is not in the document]
clairekuang: `RevitRebarHook` is a subclass of `RebarHook`, which is the inherited type of the `startHook` and `endHook` properties of `RevitRebarGroup`. So we still need to check that we are working with the subtype specifically in Revit - if you have suggestions on type polymorphism in the `RebarGroup` / `RevitRebarGroup:RebarGroup`, and `RebarHook` / `RevitRebarHook:RebarHook` class design then we can discuss further!

>"Given that we are casting it RevitRebarHook it seems a bit confusing we're calling it 'speckle'StartHook"
This is confusing yes, but `RevitRebarHook` is a Speckle class and its equivalent in revit is `RebarHookType`

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: Ah, I was handling the specific exceptions when debugging to see why sometimes the `Create` method was throwing, but they ended up not being helpful. Will revert to catching a generic exception since we are handling them the same for now.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: We should generally try to update and existing element instead of replace when possible, but this is rather tricky for most revit elements. As a first pass, we've been allowing replace instead of a true update (you'll see this in some other to native conversions). But this is a good point, we haven't clarified `update` behavior across and within connectors.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: Ah it just returns null which would be marked as a failed conversion in the connector

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: Hmm good point, I'll add more robust conditionals here, even though those params should always be valid

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: >"I'd prefer the () braces start and don't behave like {} in this context"

This is an autoformat decision, to be changed later

And yes, could collapse into one but I chose not to because accessing the `.IncludeFirstBar`and `.IncludeLastBar` props will throw if this is a single layout rebar, could get lost with the `&&` operator

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: In this case, it's skipping the retrieval of the curves for the first and/or last bar based on the `HasFirstBar` and `HasLastBar` props, but definitely retrieving the curves for all bars in between

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: It should be overridden inside the `SetInstanceParameters()` method, but I'll add it as a property on `RevitRebarHook` instead for now

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
clairekuang: resolving as autoformatting change

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
clairekuang: resolving as autoformatting change

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitCommitObjectBuilder.cs
clairekuang: line 212

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: The class property is a string, we could send the enum but this is not really human-readable after serialization

Objects/Objects/BuiltElements/Rebar.cs
clairekuang: Sure! This is currently how we've been organizing our applciation-specific namespaces for application-specific subclasses, but open for reconsideration!

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -843,12 +845,12 @@ public bool CanConvertToNative(Base @object)
         BE.Opening _ => true,
         BERC.RoomBoundaryLine _ => true,
         BERC.SpaceSeparationLine _ => true,
+        //BER.RevitRebarGroup _ => true,
```
BovineOx, line 848: commented out?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
BovineOx: I think this is consistent then with our not being consistent with update behaviour and probably a much bigger problem. Maybe, as a team, we should consider whether update is part of the story in future and size appropriately.

BovineOx (approved):

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
```diff
@@ -1,175 +1,348 @@
...
       {
-        appObj.Update(status: ApplicationObject.State.Failed, logItem: "No convertible rebar curves where found.");
-        return appObj;
+        appObj.Update(status: ApplicationObject.State.Failed, logItem: e.Message);
       }
+      finally
+      {
+        if (rebar != null)
+        {
+          SetInstanceParameters(rebar, speckleRebar);
```
connorivy, line 126: This seems pretty strange. If creating the rebar throws an exception, then we're still checking if the rebar was created and then continuing to set parameters. Did you actually experience cases where this would happen?

Objects/Objects/BuiltElements/Rebar.cs
```diff
@@ -57,6 +210,88 @@ public enum shape
...
+      set
+      {
+        if (value is not RevitRebarShape s)
+        {
+          throw new ArgumentException($"Expected object of type {nameof(RevitRebarShape)}");
+        }
+
+        revitShape = s;
+      }
+    }
```
connorivy, line 229: This is a bit tricky. C# 9 supports covarience on return types, but I think we're stuck on 8 in objects. Did you consider generic properties instead?
```
public class RebarGroup<T> where T : RebarShape
{
  public T shape { get; set; }
}
```
You'd get more type safety that way, but I'm actually not sure how well our serializer handles generics.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertRebar.cs
clairekuang: Ah this was left over from when I was handling specific exceptions - will update to return in the catch block

Objects/Objects/BuiltElements/Rebar.cs
clairekuang: This seems like a good suggestion - will implement this and remove the new `TeklaRebarGroup` class, since I haven't scoped a `TeklaRebarShape` class yet

BovineOx: I don't think we are on 8, Alan took a look and seems to be 11...https://github.com/specklesystems/speckle-sharp/blob/348fe44cd379a56cbdaf6da8454300271e88826b/Directory.Build.props#L17C3-L18C26

BovineOx: But I like Connor's suggestion, but...  so long as we are serialising a RevitRebarGroup Alan believes this should be fine with the serialiser

PR #3024, opened 2023-11-02
log(structral): Unified logging between structural connectors

JR-Morgan:
Unified connector logging between CSI, Tekla, and Revit

- Fixed issue with CSI connectors reporting as CoreUnknown
    - Added `Setup.Init` call just after we create bindings (same way Revit & Autocad)

- Added new connector helper functions for logging converter errors
   - I was finding that I was duplicating the big catch cases between connectors. So I extracted them out to two functions. One for logging to seq, the other for grabbing the correct AppObject status to report.
   
- Added converter settings to logging
   - Previously, we were just reporting the converter name + author + direction. I added the converter settings dictionary to the mix so we can create some dashboards on which settings users are using. This might help us inform defaults + where to focus our effort in future.


AlanRynne (approved):
This looks great @JR-Morgan. I love the addition of the settings ♥️

Tested in Revit and in seems to be working as expected!

PR #3008, opened 2023-10-27
chore(CSI): code cleanup and maintainability 

JR-Morgan:
Since #2970 focuses mainly on ToNative
There were a few tweaks I wanted to make to ToSpeckle to improve code readability

This PR is mostly auto-refactorings and code style changes to improve the readability of the CSI converter codebase, and to ensure we are conforming to our codeing standards.

Before Merge:
Re-target to `dev`

PR #3029, opened 2023-11-08
chore(CSI): added exception for failing areas and frames

JR-Morgan:
Extracted one important change from #3008 to ensure that we never set objects to AppObject.Status.Failed and instead throw an exception.
This guarantees that failed objects will always have an error message in the report.

### DIFF
--- a/Objects/Converters/ConverterCSI/ConverterCSIShared/PartialClasses/Geometry/ConvertArea.cs
+++ b/Objects/Converters/ConverterCSI/ConverterCSIShared/PartialClasses/Geometry/ConvertArea.cs
@@ -1,9 +1,6 @@
 using System;
 using System.Collections.Generic;
-using System.Text;
-using CSiAPIv1;
 using Objects.Structural.Geometry;
-using Objects.Structural.Analysis;
 using Objects.Structural.CSI.Properties;
 using Speckle.Core.Models;
 using StructuralUtilities.PolygonMesher;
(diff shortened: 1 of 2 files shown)


Objects/Converters/ConverterCSI/ConverterCSIShared/PartialClasses/Geometry/ConvertArea.cs
```diff
@@ -133,46 +139,42 @@ ref tableData
...
-            {
-              $"Area may not have updated successfully. Number of error messages for operation is {numErrorMsgs}",
-              importLog
-            }
-          );
-      }
-      else
+
+      string guid = null;
+      Model.AreaObj.GetGUID(name, ref guid);
```
AlanRynne, line 167: I'm assuming by this point, GetGuid is not going to fail as we've created the area successfully.
Just want to double check this is the case.

If GetGuid could still fail, we should also deal with that successCode

AlanRynne (approved):
Approved! I just got a question but not really a blocker (unless you deem it so @JR-Morgan)

Objects/Converters/ConverterCSI/ConverterCSIShared/PartialClasses/Geometry/ConvertArea.cs
JR-Morgan: It can fail (in that, there is a int return). but I'm unsure the best way to handle it here (and in the may other places we make api calls without checking the result.)

In this case, we've already created/updated the object. The guid is only used to track it for the next update.

connorivy (approved):
Did some light testing and seems like everything works the same as before except with louder errors. Good job on this 👍

PR #3030, opened 2023-11-08
Fix(Revit) : add material quantites by mesh

connorivy:
## Description & motivation

Solves issue https://github.com/specklesystems/speckle-sharp/issues/2117 and improves the calculation of material quantities.

There are some problems with obtaining quantities in Revit:

RevitAPI GetMaterialIds does not return all materials
Computation of area not clear
Some categories need further evaluation

## Changes:

- Defines a few routes to retrieve material quantities
  - If the API GetMaterialArea call will return the area of a single face (instead of all faces), then it will use that method
  - If the element is a duct or pipe, then the material values can be found on the MEP system
  - Otherwise try to get the material and the area/volume from the element's mesh.

## Checklist:

## References

Child of this PR https://github.com/specklesystems/speckle-sharp/pull/2715


Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -271,9 +271,18 @@ public void GetAllRevitParamsAndIds(Base speckleElement, DB.Element revitElement
...
+        materialQuantities = MaterialQuantitiesToSpeckle(revitElement, speckleElement["units"] as string)?
+          .ToList();
+      }
+      catch (Exception e)
+      {
+        SpeckleLog.Logger.Error(e, "Error retrieving material quantities from element of type {elementType} and category {elementCategory}", revitElement.GetType(), revitElement.Category);
+      }
+      if (materialQuantities != null)
+        speckleElement["materialQuantities"] = materialQuantities;
     }
```
AlanRynne, line 286: We can simplify this by moving the final `if` statement into the try/catch.
Would allow us to remove the null list initialisation at the top and just do `var materialQuantities = ...` in line 277.

AlanRynne, line 286: We can simplify this by moving the final if statement into the try/catch.
Would allow us to remove the null list initialisation at the top and just do var materialQuantities = ... in line 277.

Actually, if `MaterialQuantitiesToSpeckle` will either succeed or throw, there is no reason for the null check altogether.

```csharp
try
{
    speckleElement["materialQuantities"] = MaterialQuantitiesToSpeckle(...);
} catch (Exception e)
{
    SpeckleLog.Logger.Error(...)
}
```

AlanRynne (changes requested):
I just got 1 change request and a bunch of questions/considerations that are non-critical.

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -271,9 +271,18 @@ public void GetAllRevitParamsAndIds(Base speckleElement, DB.Element revitElement
...
-      var qs = MaterialQuantitiesToSpeckle(revitElement, speckleElement["units"] as string);
-      if (qs != null)
-        speckleElement["materialQuantities"] = qs;
+      List<MaterialQuantity>? materialQuantities = null;
+      try
+      {
+        materialQuantities = MaterialQuantitiesToSpeckle(revitElement, speckleElement["units"] as string)?
+          .ToList();
+      }
+      catch (Exception e)
```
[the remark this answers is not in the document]
BovineOx: Let's not catch a GE plz

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -11,69 +10,148 @@ namespace Objects.Converter.Revit
...
   {
-    #region MaterialQuantity
     /// <summary>
-    /// Gets the quantitiy of a material in one element
+    /// Material Quantities in Revit are stored in different ways and therefore need to be retrieved 
+    /// using different methods. According to this forum post https://forums.autodesk.com/t5/revit-api-forum/method-getmaterialarea-appears-to-use-different-formulas-for/td-p/11988215
+    /// "Hosts" (whatever that means) will return the area of a single side of the object while other
+    /// objects will return the combined area of every side of the element. Certain MEP element materials 
+    /// are attached to the MEP system that the element belongs to.
     /// </summary>
```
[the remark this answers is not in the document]
BovineOx: +1

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -11,69 +10,148 @@ namespace Objects.Converter.Revit
...
-      double volume = element.GetMaterialVolume(material.Id);
-      double area = element.GetMaterialArea(material.Id, false);
+      DB.Options options = new() { DetailLevel = ViewDetailLevel.Fine };
+      var (solids, _) = GetSolidsAndMeshesFromElement(element, options);
 
-      // Convert revit interal units to speckle commit units
-      double factor = ScaleToSpeckle(1);
-      volume *= factor * factor * factor;
-      area *= factor * factor;
+      var (area, volume) = GetAreaAndVolumeFromSolids(solids);
```
BovineOx, line 61: Given the types for area volume are not explicit, we should be.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -11,69 +10,148 @@ namespace Objects.Converter.Revit
...
+      foreach (ElementId matId in GetMaterialsFromSolids(solids))
+      {
+        var (area, volume) = GetAreaAndVolumeFromSolids(solids, matId);
+        yield return Create(element, matId, area, volume, units);
+      }
+    }
 
-      var speckleMaterial = ConvertAndCacheMaterial(material.Id, material.Document);
-      var materialQuantity = new Objects.Other.MaterialQuantity(speckleMaterial, volume, area, units);
+    private MaterialQuantity Create(
```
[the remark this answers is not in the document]
BovineOx: Yes this should be more explicit, agree CreateMaterialQuantity() seems obvious candidate?

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -11,69 +10,148 @@ namespace Objects.Converter.Revit
...
+      else if (element is DB.Architecture.Railing)
+      {
+        materialQuantity["length"] = (element as DB.Architecture.Railing).GetPath().Sum(e => e.Length) * factor;
+      }
+      else if (element is DB.Architecture.ContinuousRail)
+      {
+        materialQuantity["length"] = (element as DB.Architecture.ContinuousRail).GetPath().Sum(e => e.Length) * factor;
+      }
+
       return materialQuantity;
```
BovineOx, line 103: what if there's no match? I wonder if a switch on element would be better with default containg the if (LocationToSpeckle(element) is ICurve curve) and if it doesn't match....  should it just return the default MaterialQuantity or should it explode?

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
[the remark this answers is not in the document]
connorivy: I am not aware of any exceptions that will be thrown by this method. But I'm always a little nervous there may be some exceptions occur that I didn't run into during development or testing. If any exceptions do occur for some reason, then we don't want the entire object conversion to fail. 

Is there a better way that we could acheive something like this?

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
connorivy: Nice 👍

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -11,69 +10,148 @@ namespace Objects.Converter.Revit
...
+      double area = areaOfLargestFaceInEachSolid.Sum();
+      return (area, volume);
+    }
 
-      var materials = matIDs.Select(material => element.Document.GetElement(material) as DB.Material);
-      return MaterialQuantitiesToSpeckle(element, materials, units);
+    private IEnumerable<DB.ElementId> GetMaterialsFromSolids(List<Solid> solids)
+    {
+      return solids
+        .Where(solid => solid.Volume > 0 && !solid.Faces.IsEmpty)
```
BovineOx, line 129: I guess if !solid.Faces.IsEmpty is true, the m.faces.get_item(0) should be valid? Can there be other faces that should be considered?

BovineOx (commented):
Some minor things to consider

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
```diff
@@ -11,69 +10,148 @@ namespace Objects.Converter.Revit
...
+      }
+      else if (MaterialIsAttachedToMEPSystem(element))
+      {
+        MaterialQuantity quantity = GetMaterialQuantityForMEPElement(element, units);
+        return quantity == null ? new List<MaterialQuantity>() : new List<MaterialQuantity>() { quantity };
+      }
+      else
+      {
+        return GetMaterialQuantitiesFromSolids(element, units);
+      }
```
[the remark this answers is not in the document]
connorivy: The method doesn't expect the element to be a solid. It's a bit of a catch-all method that will take an element and try to get any material data that lives on the element's Revit solid. The method will be fine if there isn't any solid data on the element that is passed in

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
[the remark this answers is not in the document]
connorivy: Yeah this makes sense

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
connorivy: 👍

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
connorivy: This is an assumption that comes from the Gauip dev that a single solid will only have one material. I'm actually not sure if this is true.

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertMaterialQuantities.cs
connorivy: Changed to switch. Definitely no need for the conversion to explode if one of these criteria is not met

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -1089,21 +1098,16 @@ public static DB.Material GetMEPSystemRevitMaterial(Element e)
           idType = system.GetTypeId();
         }
       }
-      else if (IsSupportedMEPCategory(e))
+      else if (e.GetConnectorManager() is ConnectorManager connectorManager)
       {
-        MEPModel m = ((DB.FamilyInstance)e).MEPModel;
-
-        if (m != null && m.ConnectorManager != null)
+        //retrieve the first material from first connector. Could go wrong, but better than nothing ;-)
```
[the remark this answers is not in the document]
BovineOx: How do we capture this @AlanRynne?

BovineOx (approved):

AlanRynne (approved):
All comments have been resolved. We're good to go.

AlanRynne:
Added Kilian as co-author ;)

PR #3032, opened 2023-11-09
fix(archicad): Archicad nullable levels fix

jozseflkiss:
## Description & motivation

In case of null objects, `public override Level? level` can accept nulls, however an exception is thrown.

## Changes:

Enable nulls to be set as `level` properties.

## Checklist:

### DIFF
--- a/Objects/Objects/BuiltElements/Beam.cs
+++ b/Objects/Objects/BuiltElements/Beam.cs
@@ -141,10 +141,10 @@ namespace Objects.BuiltElements.Archicad
       get => archicadLevel;
       internal set
       {
-        if (value is not ArchicadLevel l)
+        if (value is ArchicadLevel or null)
+          archicadLevel = value as ArchicadLevel;
+        else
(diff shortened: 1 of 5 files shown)


Objects/Objects/BuiltElements/Beam.cs
```diff
@@ -141,10 +141,10 @@ public class ArchicadBeam : Beam
       get => archicadLevel;
       internal set
       {
-        if (value is not ArchicadLevel l)
+        if (value is ArchicadLevel or null)
```
BovineOx, line 144: Can we fix the braces aroind the if, the throw doens't need to be in an else as it will throw if not entering the if

AlanRynne (approved):
No red flags on my end 😉 We're good to go

Objects/Objects/BuiltElements/Beam.cs
JR-Morgan: @BovineOx @jozseflkiss
The throw Does need to be in an else here? otherwise it will always throw. We aren't returning here.

PR #3036, opened 2023-11-10
fix(core): Fixes Branch get Limit up to 500

AlanRynne:
Fully aligns codebase with changes from #3021.

- Created a new `ServerLimits` static class to hold this magic number (and others that will come)
- Applied the change to all calls of `StreamGetBranches`
- Added tests to verify branch limits are respected in SDK and resulting data aligns with expectations.
- Added ENV var in docker-compose.yml to upgrade test server limit to 500.

Tests are quite simple, but I didn't want to start re-engineering our integration test logic to fit this particular edge-case (tests depend on each other, can get messy). Ideally, API.cs in the integration tests should be split into pieces.

### DIFF
--- a/Core/IntegrationTests/Api.cs
+++ b/Core/IntegrationTests/Api.cs
@@ -108,7 +108,6 @@ public class Api
     Assert.True(res);
   }
 
-
   [Test, Order(13)]
   public async Task StreamSearch()
   {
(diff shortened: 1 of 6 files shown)


Core/IntegrationTests/Api.cs
```diff
@@ -275,6 +274,59 @@ public async Task StreamGetBranches()
...
+    // now let's send HTTP requests to each of these URLs in parallel
+    var options = new ParallelOptions { MaxDegreeOfParallelism = 20 };
+
+    // now let's send HTTP requests to each of these URLs in parallel
+    await Parallel.ForEachAsync(
+      Enumerable.Range(0, branchCount),
+      options,
+      async (i, cancellationToken) =>
+      {
+        await client.BranchCreate(
```
BovineOx, line 318: Are the branches literally created on the server or are they created as a mock of some sort. I don't think we want to create 500 fresh branches each time do we?

BovineOx (commented):
A query about the 500 branches

Core/IntegrationTests/Api.cs
BovineOx: If this DOES create branches we should tidy them up, maybe us a particular naming convention also, so another test could check that we don't have a bazillon branches and we fail the test, forcing us to do clean up or something, just thinking out loud

AlanRynne: Yes it does create 500 branches, but the server we're using is a brand new one spun up for every CI run. So cleanup happens once the job is over. No deployed servers are used for integration testing.

That said, I do agree with the sentiment, but our current way of doing these tests doesn't allow for much flexibility.

BovineOx: No that's perfect

BovineOx (approved):
LGTM

PR #3043, opened 2023-11-13
fix(revit): receiving generic mep elements

clairekuang:
## Description & motivation

Fixes bugs when receiving generic mep elements, which was a regression due to the recent revit `Network` refactor. This was causing failed conversions when receiving built elements from Grasshopper.

Includes some minor changes to `RoofToNative` and `FreeformElementToNative` conversions, but does not successfully fix receiving for those elements sent from Grasshopper BuiltElements test file.

## Changes:

- Revit converter

## Screenshots:

Before:
Untitled

After:
image

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertDuct.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertDuct.cs
@@ -16,14 +16,16 @@ namespace Objects.Converter.Revit
     public ApplicationObject DuctToNative(BuiltElements.Duct speckleDuct)
     {
       var speckleRevitDuct = speckleDuct as RevitDuct;
-      var docObj = GetExistingElementByApplicationId(speckleDuct.applicationId);
+      Element docObj = GetExistingElementByApplicationId(speckleDuct.applicationId);
       var appObj = new ApplicationObject(speckleDuct.id, speckleDuct.speckle_type) { applicationId = speckleDuct.applicationId };
 
       // skip if element already exists in doc & receive mode is set to ignore
       if (IsIgnore(docObj, appObj))
+      {
         return appObj;
+      }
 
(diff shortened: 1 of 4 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertDuct.cs
```diff
@@ -87,9 +91,9 @@ public ApplicationObject DuctToNative(BuiltElements.Duct speckleDuct)
         TrySetParam(duct, BuiltInParameter.RBS_VELOCITY, speckleRevitDuct.velocity, speckleRevitDuct.units);
 
         SetInstanceParameters(duct, speckleRevitDuct);
-      }
 
-      CreateSystemConnections(speckleRevitDuct.Connectors, duct, receivedObjectsCache);
+        CreateSystemConnections(speckleRevitDuct.Connectors, duct, receivedObjectsCache);
```
clairekuang, line 95: This is the actual bug fix: for generic mep elements, `speckleRevitDuct` is null

Objects/Converters/ConverterRevit/ConverterRevitShared/Partial Classes/ConvertPipe.cs
```diff
@@ -114,12 +127,13 @@ public ApplicationObject PipeToNative(BuiltElements.Pipe specklePipe)
       }
 
       if (speckleRevitPipe != null)
+      {
         SetInstanceParameters(pipe, speckleRevitPipe);
+        CreateSystemConnections(speckleRevitPipe.Connectors, pipe, receivedObjectsCache);
```
clairekuang, line 132: This is the actual bug fix: for generic mep elements, `speckleRevitPipe` is null

BovineOx (approved):
Looks fine, I see only the one concrete change to solve the null reference and the rest are style changes

PR #3045, opened 2023-11-14
fix(revit): extrusion roof and freeform elements conversion

clairekuang:
## Description & motivation
Extrusion roofs were not being properly converted to native due to incorrect use of `roof.outline` in place of `extrusionRoof.referenceLine`.

Freeform elements received in revit from GH with no subcategory assigned are now successfully created. Previously, setting null or generic model default subcategory was throwing errors on solid creation.

## Changes:
- Revit converter


AlanRynne (approved):
Looks legit (including braces 😉)  let's go for it

PR #3047, opened 2023-11-14
Clearer error message for failed spaces

JR-Morgan:
For Revit to native, we are still updating the app object directly.
Not nice, but safer imo than throwing right now until we properly refactor this aspect of the connector (like we've already done for ToSpeckle)


clairekuang (changes requested):
Let's use "Basepoint was null" in the appObj log message

clairekuang (approved):
Just saw the change! Looks good to me

PR #3048, opened 2023-11-15
fix(Archicad): Timeout after 100s when sending large models #2

jozseflkiss:
## Description & motivation

Additional timeout increase for https://github.com/specklesystems/speckle-sharp/pull/3038.

## Changes:

Timeout of HTTP request calls to Archicad API has been increased to 300s.
(Investigation of possible performance slow down is to come later.)

## Checklist:

- [x] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x] My commits are related to the pull request and do not amend unrelated code or documentation.
- [x] My code follows a similar style to existing code.
- [ ] I have added appropriate tests.
- [ ] I have updated or added relevant documentation.

PR #3052, opened 2023-11-20
fix(Archicad): Timeout after 100s when sending large models #3

jozseflkiss:
## Description & motivation

Solution for https://github.com/specklesystems/speckle-sharp/pull/3048.
Because Grid elements have 3D lines in their 3D representation in Archicad, as a general solution, all kind of edge data is transferred from Archicad to the C# connector. This extra data slowed down the send operation. Since currently we don't story edge information for meshes on the Speckle server, it is not required to transfer all edge data, only the lines without adjacent polygons (which is proper for grids).

## Changes:

C++ connector sends only 3D lines.
Also, time measure points have been introduced for the send operation as well.

## Checklist:


ConnectorArchicad/ConnectorArchicad/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "ConnectorArchicad": {
+      "commandName": "Project",
+      "commandLineArgs": "19723 888"
```
AlanRynne, line 5: What are these obscure arguments 😅 Could we get some comments to detail what they mean?

ConnectorArchicad/AddOn/Sources/AddOn/Commands/GetModelForElements.cpp
```diff
@@ -79,7 +79,9 @@ static void GetModelInfoForElement (const Modeler::Elem& elem,
 		// edges
 		for (ULong edgeIdx = 0; edgeIdx < body.GetEdgeCount (); ++edgeIdx) {
 			const EDGE& edge = body.GetConstEdge (edgeIdx);
-			modelInfo.AddEdge (ModelInfo::EdgeId (edge.vert1 + vetrexOffset, edge.vert2 + vetrexOffset), ModelInfo::EdgeData (ModelInfo::VisibleEdge, edge.pgon1, edge.pgon2));
+			// send only edges which have no polygons (only 3D lines)
+			if (edge.pgon1 == Brep::MeshBrep::Edge::InvalidPgonIdx && edge.pgon2 == Brep::MeshBrep::Edge::InvalidPgonIdx)
+				modelInfo.AddEdge (ModelInfo::EdgeId (edge.vert1 + vetrexOffset, edge.vert2 + vetrexOffset), ModelInfo::EdgeData (ModelInfo::VisibleEdge, edge.pgon1, edge.pgon2));
```
AlanRynne, line 84: Not sure I fully understand the reasoning behind this, or what it is actually doing.

Could you elaborate more on what other "non 3D lines" were getting sent? And why are they not necessary?

AlanRynne (changes requested):
I've just got 2 questions 👇🏼

ConnectorArchicad/ConnectorArchicad/Properties/launchSettings.json
jozseflkiss: The first argument is the HTTP port of the Archicad JSON API, and the second is the Archicad version number (for testing purposes I use very distinct numbers). I'll remove this file from the PR.

ConnectorArchicad/AddOn/Sources/AddOn/Commands/GetModelForElements.cpp
jozseflkiss: The are vertices, edges and polygons for meshes in Archicad. The edge.pgon1 and edge.pgon2 are the indices of the neighbour polygons of the edge (there are max two of them). If the edge.pgon1 and edge.pgon2 are InvalidPgonIdx, it means that edge is just a line in the 3D model with not polygons attached. This is the case for grid elements. So only these kind of edge/lines are sent to speckle. (Edge status for Mesh elements are not supported yet.)

AlanRynne: Oohhh, ok I get it now. This sounds legit 🙌🏼 flagging this as resolved

ConnectorArchicad/ConnectorArchicad/Properties/launchSettings.json
AlanRynne: Beautiful, i'll wait for this removal and aprove+merge

AlanRynne (approved):
Approving after first review + chat with Jozsef. We should release an `rc` to test this in the release version.

PR #3053, opened 2023-11-20
chore: symbol naming enforcement across all csprojects

JR-Morgan:
Open Questions
- Do we want to keep the Resharper stuff?
- If so, is it fully aligned with our other analysers?
- Should we make `EnforceCodeStyleInBuild` only true for CI


AlanRynne:
As for your questions, we chatted about this but might as well leave this here.

> * Do we want to keep the Resharper stuff?
Decision was to kill any resharper options
> * If so, is it fully aligned with our other analysers?
Non applicable ☝🏼
> * Should we make `EnforceCodeStyleInBuild` only true for CI
YES

PR #3054, opened 2023-11-20
Fix(Revit) : correctly parse units coming from other applications

connorivy:
## Description & motivation

units set manually in other applications such as GH were not being correctly picked up and scaled

## Changes:

Don't use the Speckle.Core.Kits.Units.IsUnitSupported(units) method, which expects a very specific unit string format (i.e. returns true for "mm" but not for "millimeter"). Instead, we use our "UnitsToNative" method which will attempt to parse many variations of the unit string.

## Validation of changes:

receiving a model, https://speckle.xyz/streams/9ff64f51e6/commits/06c8bd7bec, now correctly sets the floor offsets which it wasn't correctly doing before

## Checklist:


AlanRynne (approved):
Nice! Thx for the last minute change ;)

AlanRynne:
Switched target to `main` instead.

PR #3056, opened 2023-11-21
fix (navisworks): Handle calling Clash API and receiving nothing back

jsdbroughton:
The ClashAPI, under certain circumstances, returns nothing, in place of the API object with callable methods.

The documented behaviour is to have the methods return nulls or empty lists, not that the ClashAPI object be itself null.

This PR adds defensive NullReference checking.


ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
```diff
@@ -30,7 +30,7 @@ public override List<ISelectionFilter> GetSelectionFilters()
 
     var selectSetsRootItem = _doc.SelectionSets.RootItem;
 
-    var savedSelectionSets = selectSetsRootItem.Children.Select(GetSets).ToList();
+    var savedSelectionSets = selectSetsRootItem?.Children.Select(GetSets).ToList() ?? new List<TreeNode>();
```
BovineOx, line 33: Will Children always be non-null in these two linq statements also?

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
```diff
@@ -68,20 +66,31 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
+        {
+          //  var clashReportFilter = new TreeSelectionFilter
+          //  {
+          //    Slug = "clashes", Name = "Clash Detective Results", Icon = "MessageAlert",
+          //    Description = "Select group clash test results.",
+          //    Values = groupedClashResults
+          //  };
+          //  filters.Add(clashReportFilter);
+        }
+        Console.WriteLine("Sending Clashes is not supported yet.");
```
BovineOx, line 88: I guess this should come out or go via structured logger?

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
```diff
@@ -68,20 +66,31 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
+      var clashTests = clashPlugin?.TestsData;
 
-    if (groupedClashResults?.Count >= 0)
+      if (clashTests != null)
+      {
+        var groupedClashResults = clashTests.Tests.Select(GetClashTestResults).Where(x => x != null).ToList();
+
+        if (groupedClashResults.Count >= 0)
+        {
+          //  var clashReportFilter = new TreeSelectionFilter
```
BovineOx, line 80: commented out code :pained_face

Are you wanting to retain this code for when we DO want to support this? If so, probably better to create or add to an issue and put the code there... I do wonder what the code is doing if in the inner most if block we don't actually do anything :D

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
```diff
@@ -68,20 +66,31 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
+    catch (NullReferenceException)
     {
-      //  var clashReportFilter = new TreeSelectionFilter
-      //  {
-      //    Slug = "clashes", Name = "Clash Detective Results", Icon = "MessageAlert",
-      //    Description = "Select group clash test results.",
-      //    Values = groupedClashResults
-      //  };
-      //  filters.Add(clashReportFilter);
+      // ignore
```
BovineOx, line 93: 😨 So.... you are putting this here as belt and braces but it's going to swallow any NREs, if it is to catch unexpected NREs then we should at least be logging them with some consideration as to how verbose that might be

BovineOx (commented):
A few queries

ConnectorNavisworks/ConnectorNavisworks/Entry/SpeckleNavisworksCommand.cs
```diff
@@ -25,6 +25,7 @@ namespace Speckle.ConnectorNavisworks.Entry;
     ExtendedToolTip = "Speckle Connector for Navisworks"
   )
 ]
+// ReSharper disable once ClassNeverInstantiated.Global - This is because of the mechanism Navisworks uses to load plugins.
```
AlanRynne, line 28: FYI, we're thinking of downgrading all the resharper warnings to hints, so that this will no longer be necessary, as we'll be guiding ourselves by the Code Analyzers rules instead (which are enforceable on CI).

No need for removing this, just FYI it may not be necessary for much longer ;)

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
jsdbroughton: API documentation says it will be an empty list

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
jsdbroughton: i only added this to clear the IDE complaining of an empty conditional branch

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
jsdbroughton: This is a not-yet-supported feature. It was in earlier releases but with suboptimal results.

jsdbroughton:
A feature release is available on Manager for now - if this IS indeed the cause of the manifested problem, I will tidy this up with the comments on the board. This is quicker than installing a copy of Simulate just now.

ConnectorNavisworks/ConnectorNavisworks/Bindings/ConnectorNavisworksBindings.Filters.cs
jsdbroughton: removed

AlanRynne (approved):
All comments have been resolved. Approving 🚀

PR #3061, opened 2023-11-22
fix main build warnings batchuploader

JR-Morgan:


AlanRynne (approved):
All these are safe changes.

PR #3081, opened 2023-11-30
Chore(props): Reduced some more subjective warnings down to suggestions/none

JR-Morgan:
See build props changes
I have reduced several subjective warnings down to a `suggestion` or `none`

I have also enabled some optional flags to make CA2000 more lenient for disposal object ownership transfers that don't quite match Microsoft ideal patterns.
see https://github.com/dotnet/roslyn-analyzers/blob/main/docs/Analyzer%20Configuration.md#dispose-analysis-kind-for-idisposable-rules

This should mean for us, the remaining CA2000 violations should all be genuine disposal issues 🤞.


AlanRynne (approved):
🙌🏼 Beautiful! Thnx for all the investigation going behind this.

PR #3082, opened 2023-11-30
chore(core): CNX-7435 Core warning reduction

JR-Morgan:
Resolves many warnings in Core

Merge first:
- https://github.com/specklesystems/speckle-sharp/pull/3065 
- https://github.com/specklesystems/speckle-sharp/pull/3081 first


JR-Morgan:
I want to give this some quick testing in some connectors and automate. All the changes I've made should be pretty innocent, but there's so many, it's worth being cautious.

Core/Core/Api/Exceptions.cs
```diff
@@ -13,8 +13,8 @@ namespace Speckle.Core.Api;
 /// </summary>
 public class SpeckleGraphQLException<T> : SpeckleException
 {
-  private GraphQLRequest _request;
-  public readonly GraphQLResponse<T>? Response;
+  private readonly GraphQLRequest _request;
+  public GraphQLResponse<T>? Response { get; }
```
AlanRynne, line 17: Shouldn't the Response also be `readonly`?

Core/Core/Api/Operations/Operations.Serialize.cs
AlanRynne, line 1: If I understand this correctly, this is a binary change but not a breaking change to a 3rd party dev right?

Core/Core/Credentials/StreamWrapper.cs
```diff
@@ -207,13 +207,13 @@ private void StreamWrapperFromUrl(string streamUrl)
...
+          if (uri.Segments[1].ToLowerInvariant() != "streams/")
           {
-            StreamId = uri.Segments[2].Replace("/", "");
+            throw new SpeckleException($"Cannot parse {uri} into a stream wrapper class.");
           }
           else
           {
-            throw new SpeckleException($"Cannot parse {uri} into a stream wrapper class.");
+            StreamId = uri.Segments[2].Replace("/", "");
           }
```
AlanRynne, line 217: I'm amazed this was not flagged as redundant.

Core/Core/Logging/SpeckleLog.cs
```diff
@@ -247,17 +247,17 @@ public static void OpenCurrentLogFolder()
     {
       Process.Start(_logFolderPath);
     }
-    catch (Exception ex)
+    catch (FileNotFoundException ex)
```
AlanRynne, line 250: Could this potentially be another exception, such as unauthorised access?

Core/Core/Models/Base.cs
```diff
@@ -24,18 +23,22 @@ namespace Speckle.Core.Models;
...
 {
-  private static readonly Regex ChunkSyntax = Constants.ChunkPropertyNameRegex;
+  private static readonly Regex s_chunkSyntax = Constants.ChunkPropertyNameRegex;
 
-  private string __type;
+  private string _type;
 
   /// <summary>
-  /// A speckle object's id is an unique hash based on its properties. <b>NOTE: this field will be null unless the object was deserialised from a source. Use the <see cref="GetId(bool)"/> function to get it.</b>
+  /// A speckle object's id is an uniq`ue hash based on its properties. <b>NOTE: this field will be null unless the object was deserialised from a source. Use the <see cref="GetId(bool)"/> function to get it.</b>
```
AlanRynne, line 35: Typo!

Core/Core/Api/Exceptions.cs
JR-Morgan: Response was previously a public readonly field.
I've made it a get only property, since public fields should be avoided.
So in that sense, by being a `{ get; }` its already read-only, you can't use the `readonly` keyword with class properties.

Core/Core/Credentials/StreamWrapper.cs
JR-Morgan: Which? the `else`?

Not going to lie, I don't remember making this change. Perhaps it was a simple rider refactor preferring exceptional paths first.
Happy to revert if you'd prefer

Core/Core/Logging/SpeckleLog.cs
JR-Morgan: Good point, perhaps we can make this catch all `IOException` and `UnauthorisedAccessException`
Would be a safer change.

I did check seq for this error template, zero occurrences in the last 3 months.

Core/Core/Transports/ServerUtils/ServerAPI.cs
```diff
@@ -18,16 +18,21 @@ namespace Speckle.Core.Transports.ServerUtils;
...
+  private const int MAX_MULTIPART_COUNT = 5;
+  private const int MAX_MULTIPART_SIZE = 25_000_000;
+  private const int MAX_OBJECT_SIZE = 25_000_000;
+
+  private const int MAX_REQUEST_SIZE = 100_000_000;
+
+  private const int RETRY_COUNT = 3;
+  private static readonly HashSet<int> s_retryCodes = new() { 408, 502, 503, 504 };
+  private static readonly char[] s_separator = { '\t' };
+  private static readonly string[] s_filenameSeparator = { "filename=" };
```
AlanRynne, line 33: We should consider moving these to a "magic numbers" class at some point, as ideally all these values should be defined by the server.

Core/Core/Transports/ServerUtils/ServerAPI.cs
```diff
@@ -72,13 +77,13 @@ public async Task<string> DownloadSingleObject(string streamId, string objectId)
...
-    HttpResponseMessage rootHttpResponse = null;
-    while (ShouldRetry(rootHttpResponse))
+    HttpResponseMessage rootHttpResponse;
+    do
     {
       rootHttpResponse = await _client
         .SendAsync(rootHttpMessage, HttpCompletionOption.ResponseContentRead, CancellationToken)
         .ConfigureAwait(false);
-    }
+    } while (ShouldRetry(rootHttpResponse));
```
AlanRynne, line 86: ♥️

AlanRynne (approved):
Added some single comments before this review by accident, and a couple more within this 👇🏼

Core/Core/Api/Operations/Operations.Serialize.cs
AlanRynne: Agreed!

Core/Core/Api/Exceptions.cs
AlanRynne: oh yes... obviously 🤦🏼‍♂️

Core/Core/Logging/SpeckleLog.cs
AlanRynne: 🙌🏼 Let's move it to be `IOException` just for extra safety.

PR #3083, opened 2023-11-30
Warnings/fix/navisworks

jsdbroughton:
This global PR tackles many warnings to pre-empt the more granular initiative.

It is understood to address:

- CA1032: Implement standard exception constructors
- CA1051: Do not declare visible instance fields
- CA1508: Avoid dead conditional code
- CA1724: Type names should not match namespaces
- CA1854: Prefer 'IsEmpty' over 'Count' to determine whether the collection contains any items
- CA1859: Avoid unnecessary lambda captures
- CA1860: Use Enumerable.Any() to determine whether any item exists in a collection
- CA1861: Do not use 'Enumerable.Last()' to determine whether any item exists in a collection
- IDE0004: Cast is redundant
- IDE0010: Add missing cases to switch statements
- IDE0033: Use explicitly provided tuple name
- IDE0036: Order modifiers
- IDE0044: Make field read-only
- IDE0047: Remove unnecessary parentheses
- IDE0059: Unnecessary assignment of a value
- IDE0066: Convert switch statement to expression

3 suppressions have been introduced with Justifications:

1. CA1812: Avoid uninstantiated internal classes: for the two classes that Navis will instantiate at runtime - this is the plugin pattern
2. CA1508: Avoid dead conditional code: for one class that adds condition handlers not yet migrated to UI options. Yes, this is a partial implementation floating around.
3. CA1031: Do not catch general exception types: For a single instance where Avalonia is initiated, I have no clue what exceptions that may throw.


jsdbroughton:
@alan i targetted this at dev as requested, but having branched from IDE0005 it includes some noise from the merged fixes contained within. It passes CI, but if you need me to finesse this a bit, let me know. Your GitFu maybe needed.

ConnectorNavisworks/ConnectorNavisworks/Other/SelectionBuilder.cs
```diff
@@ -48,24 +48,26 @@ public void GetFromFilter()
...
-      case FilterTypes.Sets:
+      case FilterTypes.SETS:
         _uniqueModelItems.AddRange(GetObjectsFromSavedSets());
         break;
 
-      case FilterTypes.Views:
+      case FilterTypes.VIEWS:
         _uniqueModelItems.AddRange(GetObjectsFromSavedViewpoint());
         break;
+      default:
```
BovineOx, line 62: Being picky I'd prefer a line space between the break of the previous case and the default case and generally between break and cases but maybe that's coming soon.

BovineOx (approved):
Happy for this to go as-is but worth reviewing my comments as it throws up a few Qs on some of the existing code.

PR #3087, opened 2023-12-01
Fix(dui3): squash and rebase all dui3 changes onto dev history

oguzhankoral:
Since `dui3000` history is dead when `dev` squashed and merged into it, we needed to reset history to make sure everything aligned perfectly with `dev`.


AlanRynne (approved):
Nice! 🙌🏼 I'm very subjectively aproving this 😉

PR #3092, opened 2023-12-02
fix(csi): CNX-8342 Fix CA1806 for CSI projects

AlanRynne:
Removes all CA1806 warnings from CSI projects:

This fixes a bug in `GetAllNamesOfObjectType` where the `Model` name was not being included in the list.

Also removes some unnecessary `ToList` calls whose output was not being used.


connorivy (approved):
👍

PR #3094, opened 2023-12-04
feat(Core): CNX-8352 operations.send

JR-Morgan:
As part of our effort to reduce analyser warnings across our code base. It's important that our Core functions are clear about what exceptions types they throw.

This motivation for this PR was to focus on `Operations.Send` and attempt to enforce and provide clear guidance on the exception types that can be thrown.

Our original `Operations.Send` overloads were doing some smelly things:

In the signature of the function alone:
1. We were allowing users to pass in a `SerializerVerion` enum to switch to using the old v1 serializer. This serializer hasn't been used (or tested) in almost two years, its past time we drop exposing it as an option for `Operations.Send`
2. A long time ago, we were using `OnErrorAction` to signal errors in transports and sending. We have since deprecated this in favour of throwing proper Exceptions instead. This property is effectively unused with our current `Operations.Send`
3. We were exposing a `disposeTransports` boolean (defaulting to false) that, when true, we would loop through and dispose any `IDisposable` transports that the user passes through. This doesn't follow any normal C# guidance on how `IDisposables` should be handled, and on top of that, we weren't disposing inside a finally block, so if an exception did occur, we weren't disposing the transports anyway. Thus this "feature" was not really usable, and much cleaner solved by expecting caller to handle the disposal using e.g. using a `using` keyword.

And in the implementation:
1. We were performing input validation, logging, timers, error handling, SQLite setup, transport manipulation + extra nonsense for serailizer v1, as well as performing the actual send!

In my new implementation, we have far fewer parameters, removing the obsolete and questionable designed ones.
We have also separated logging and error handing into one function `SendToTransports` and the actual send logic in an internal `SendInternal` function.

### DIFF
--- a/Core/Core/Api/Operations/Operations.Send.cs
+++ b/Core/Core/Api/Operations/Operations.Send.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -19,76 +20,67 @@ namespace Speckle.Core.Api;
 public static partial class Operations
 {
   /// <summary>
-  /// Sends a Speckle Object to the provided <paramref name="transport"/> and the default local <see cref="SQLiteTransport"/>
+  /// Sends a Speckle Object to the provided <paramref name="transport"/> and (optionally) the default local cache
   /// </summary>
-  /// <param name="value">The object you want to send</param>
-  /// <param name="transport">Where you want to send them</param>
-  /// <param name="onProgressAction">Action that gets triggered on every progress tick (keeps track of all transports)</param>
-  /// <param name="cancellationToken"></param>
-  /// <returns>The id (hash) of the object sent</returns>
-  /// <exception cref="ArgumentNullException"><paramref name="value"/> or <paramref name="transport"/> was null</exception>
-  /// <exception cref="SpeckleException">Serialization or Send operation was unsuccessful</exception>
-  /// <exception cref="TransportException"><paramref name="transport"/> failed to send</exception>
-  /// <exception cref="OperationCanceledException"><paramref name="cancellationToken"/> requested cancellation</exception>
-  /// <example>
-  /// <code>
+  /// <remarks/>
+  /// <inheritdoc cref="Send(Base, IReadOnlyCollection{ITransport}, Action{ConcurrentDictionary{string, int}}?, CancellationToken)"/>
+  /// <param name="useDefaultCache">When <see langword="true"/>, an additional <see cref="SQLiteTransport"/> will be included</param>
+  /// <example><code>
   /// using ServerTransport destination = new(account, streamId);
-  /// string id = await Operations.Send(mySpeckleData, destination);
-  /// </code>
-  /// </example>
+  /// string objectId = await Send(mySpeckleObject, destination, true);
+  /// </code></example>
   public static async Task<string> Send(
     Base value,
-    ITransport transport,
+    ITransport? transport,
+    bool useDefaultCache,
     Action<ConcurrentDictionary<string, int>>? onProgressAction = null,
     CancellationToken cancellationToken = default
   )
   {
-    if (transport is null)
+    List<ITransport> transports = new();
+    if (transport != null)
     {
-      throw new ArgumentNullException(nameof(transport));
+      transports.Add(transport);
     }
 
-    using var sqLiteTransport = new SQLiteTransport { TransportName = "LC" };
-    ITransport[] transports = { sqLiteTransport, transport };
-    return await SendToTransports(value, transports, onProgressAction, cancellationToken).ConfigureAwait(false);
+    using SQLiteTransport? localCache = useDefaultCache ? new SQLiteTransport { TransportName = "LC" } : null;
+    if (localCache != null)
+    {
+      transports.Add(localCache);
+    }
+
+    return await Send(value, transports, onProgressAction, cancellationToken).ConfigureAwait(false);
   }
 
   /// <summary>
   /// Sends a Speckle Object to the provided <paramref name="transports"/>
   /// </summary>
   /// <remarks>Only sends to the specified transports, the default local cache won't be used unless you also pass it in</remarks>
+  /// <returns>The id (hash) of the object sent</returns>
   /// <param name="value">The object you want to send</param>
   /// <param name="transports">Where you want to send them</param>
   /// <param name="onProgressAction">Action that gets triggered on every progress tick (keeps track of all transports)</param>
   /// <param name="cancellationToken"></param>
-  /// <returns>The id (hash) of the object sent</returns>
   /// <exception cref="ArgumentException">No transports were specified</exception>
-  /// <exception cref="ArgumentNullException"><paramref name="value"/> was null</exception>
+  /// <exception cref="ArgumentNullException">The <paramref name="value"/> was <see langword="null"/></exception>
   /// <exception cref="SpeckleException">Serialization or Send operation was unsuccessful</exception>
   /// <exception cref="TransportException">One or more <paramref name="transports"/> failed to send</exception>
-  /// <exception cref="OperationCanceledException"><paramref name="cancellationToken"/> requested cancellation</exception>
-  /// <example>
-  /// <code>
-  /// using ServerTransport t1 = new(account, streamId);
-  /// MemoryTransport t2 = new();
-  /// string id = await Operations.Send(mySpeckleData, new ITransport[] {t1, t2});
-  /// </code>
-  /// </example>
-  public static async Task<string> SendToTransports(
+  /// <exception cref="OperationCanceledException">The <paramref name="cancellationToken"/> requested cancellation</exception>
+  public static async Task<string> Send(
     Base value,
     IReadOnlyCollection<ITransport> transports,
     Action<ConcurrentDictionary<string, int>>? onProgressAction = null,
     CancellationToken cancellationToken = default
   )
   {
-    if (transports == null || transports.Count == 0)
-    {
-      throw new ArgumentException("Expected at least on transport to be specified", nameof(transports));
-    }
-
     if (value is null)
     {
       throw new ArgumentNullException(nameof(value));
+    }
+
+    if (transports.Count == 0)
+    {
+      throw new ArgumentException("Expected at least on transport to be specified", nameof(transports));
     }
 
     var transportContext = transports.ToDictionary(t => t.TransportName, t => t.TransportContext);
@@ -115,7 +107,7 @@ public static partial class Operations
       string hash;
       try
       {
-        hash = await SendInternal(value, serializerV2, cancellationToken).ConfigureAwait(false);
+        hash = await SerializerSend(value, serializerV2, cancellationToken).ConfigureAwait(false);
       }
       catch (Exception ex)
       {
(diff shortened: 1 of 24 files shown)


gjedlicska (changes requested):

Core/Core/Api/Operations/Operations.Send.cs
```diff
@@ -18,17 +18,183 @@ namespace Speckle.Core.Api;
...
+  /// <exception cref="OperationCanceledException"><paramref name="cancellationToken"/> requested cancellation</exception>
+  /// <example>
+  /// <code>
+  /// using ServerTransport destination = new(account, streamId);
+  /// string id = await Operations.Send(mySpeckleData, destination);
+  /// </code>
+  /// </example>
+  public static async Task<string> Send(
+    Base value,
+    ITransport transport,
```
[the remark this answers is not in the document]
JR-Morgan: Gergo and I had a quick discussion about this. There's a couple things we can do to make it clearer which overloads can be used.
I'm going to make some changes and we can discuss the improvement

Core/Core/Models/Base.cs
```diff
@@ -32,7 +32,7 @@ public class Base : DynamicBase
   private string _type;
 
   /// <summary>
-  /// A speckle object's id is an unique hash based on its properties. <b>NOTE: this field will be null unless the object was deserialised from a source. Use the <see cref="GetId(bool)"/> function to get it.</b>
+  /// A speckle object's id is an uniq`ue hash based on its properties. <b>NOTE: this field will be null unless the object was deserialised from a source. Use the <see cref="GetId(bool)"/> function to get it.</b>
```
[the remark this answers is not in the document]
BovineOx: I think it's a just a classy, slightly European way of saying unique

Core/Core/Api/Operations/Operations.Send.cs
```diff
@@ -18,17 +18,183 @@ namespace Speckle.Core.Api;
...
+    Action<ConcurrentDictionary<string, int>>? onProgressAction = null,
+    CancellationToken cancellationToken = default
+  )
+  {
+    if (transport is null)
```
BovineOx, line 46: Is this necessary anymore? transport isn't nullable, calling this and ignoring the contract would be a bi daft

Core/Core/Api/Operations/Operations.Serialize.cs
```diff
@@ -18,11 +18,11 @@ public static partial class Operations
...
   /// <see cref="Send(Speckle.Core.Models.Base,System.Collections.Generic.List{Speckle.Core.Transports.ITransport}?,bool,System.Action{System.Collections.Concurrent.ConcurrentDictionary{string,int}}?,System.Action{string,System.Exception}?,bool,Speckle.Core.Api.SerializerVersion)"/>
   /// </remarks>
-  /// <param name="object"></param>
-  /// <param name="cancellationToken">Propagates notification that operations should be canceled.</param>
+  /// <param name="value">The object to serialise</param>
+  /// <param name="cancellationToken"></param>
   /// <returns>A json string representation of the object.</returns>
   public static string Serialize(
-    Base @object,
+    Base value,
```
BovineOx, line 25: thank you :)

Core/Core/Api/Operations/Operations.Serialize.cs
```diff
@@ -47,7 +46,7 @@ public static partial class Operations
   /// </summary>
   /// <param name="objects"></param>
   /// <returns></returns>
-  [Obsolete("Please use the Serialize(Base @object) function. This function will be removed in later versions.")]
+  [Obsolete("Please use the Serialize(Base @object) function. This function will be removed in later versions.", true)]
```
BovineOx, line 49: Does this need to now be (Base value)  ?

Core/Core/Kits/Applications.cs
```diff
@@ -53,45 +56,46 @@ public string GetVersion(HostAppVersion version)
...
+    Other = new("Other", "other");
 
   /// <summary>
   /// Gets a HostApplication form a string. It could be the versioned name or a string coming from a process running.
   /// </summary>
   /// <param name="appname">String with the name of the app</param>
   /// <returns></returns>
-  public static HostApplication GetHostAppFromString(string appname)
+  [Pure]
+  public static HostApplication GetHostAppFromString(string? appname)
```
BovineOx, line 98: could we not create a static dinctionary here of the HostApplication instances and turn this into a bit of linq? I think that is cleaner and a bit more extensible?

Core/Core/Serialisation/BaseObjectSerializerV2.cs
```diff
@@ -217,7 +216,7 @@ public string Serialize(Base baseObj)
     IEnumerable<string> dynamicProperties = baseObj.GetDynamicMembers();
 
     // propertyName -> (originalValue, isDetachable, isChunkable, chunkSize)
```
BovineOx, line 218: So object has to be nullable because GetValue() can return null presuambly and we add the tuple anyways, even if baseValue is null? If GetValue() can return null are we sure that Name is always valid? Again, appreciate this is slighly beyond the scope of the nullability change

BovineOx: Similar occurs below for dynamicProps also

Core/Core/Transports/ITransport.cs
```diff
@@ -90,9 +96,9 @@ public interface ITransport
   /// <param name="targetTransport">The transport you want to copy the object to.</param>
   /// <param name="onTotalChildrenCountKnown">(Optional) an <see cref="Action{T}"/> that will be invoked once, when the number of object children to be copied over is known.</param>
   /// <returns>The string representation of the root object.</returns>
-  /// <exception cref="InvalidOperationException">The transport was in an invalid state</exception>
   /// <exception cref="ArgumentException">The provided arguments are not valid</exception>
-  /// <exception cref="OperationCanceledException"></exception>
+  /// <exception cref="InvalidOperationException">The transport was in an invalid state</exception>
+  /// <exception cref="OperationCanceledException"><see cref="CancellationToken"/> requested cancel</exception>
```
BovineOx, line 101: Again, beyond scope but the commented out GetBlock() call below? Worth binning it while we are here?

Core/Core/Transports/Memory.cs
```diff
@@ -57,10 +57,7 @@ public void BeginWrite()
   public void SaveObject(string id, string serializedObject)
   {
     var stopwatch = Stopwatch.StartNew();
-    if (CancellationToken.IsCancellationRequested)
-    {
-      return; // Check for cancellation
-    }
+    CancellationToken.ThrowIfCancellationRequested();
```
BovineOx, line 60: Kind of nervous about the frequency of these checks - typically cancellation checks are performmed within a loop, I don't think it has to be caught absolutely immediately... but again, it's a little outside the remit

Core/Core/Transports/Memory.cs
```diff
@@ -107,18 +104,18 @@ public void SaveObject(string id, ITransport sourceTransport)
 
   public Task WriteComplete()
   {
-    return Utilities.WaitUntil(() => true);
+    return Task.CompletedTask;
```
BovineOx, line 107: Is this a bit naughty? Or is it just always complete because the act of saving alwaygs completes is the contract for this such that you write and wait for complete and then write the next blob and so on?

Core/Core/Kits/Applications.cs
JR-Morgan: This is all pre-existing. I don't like how it is, a dictionary would be so much nicer. But for now, I didn't want to refactor any more than necessary.

Core/Core/Serialisation/BaseObjectSerializerV2.cs
JR-Morgan: We allow for `null` dynamic values.
I wanted to make sure the nullability syntax was correct here

Core/Core/Api/Operations/Operations.Serialize.cs
JR-Morgan: It should be, I'll fix in another PR

Core/Core/Api/Operations/Operations.Send.cs
JR-Morgan: I've changed this with my latest changes, we are no longer doing this check for transports.
However, I've still kept it in for `value`, since I'm not confident that it would throw lower down (we CAN serialize a null) but in the context of a send, we don't want users doing this.

If we were in .NET Standard2.1, we would want to mark the parameter as nullable, but use the `[NotNull]` to say that this function will always throw when a null is passed in. see https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis 
We can use Polysharp to bring this attribute back to us, but we haven't got that setup (tho, I believe we're no longer blocked on the CI side)

So for now, I think this is fine.

Core/Core/Transports/ITransport.cs
JR-Morgan: Depends whether you consider this a comment to stop developers adding a `GetBlob` function back later down the road.
I'll leave this one up to Gergo

Core/IntegrationTests/ServerTransportTests.cs
```diff
@@ -69,7 +69,7 @@ public async Task SendAndReceiveObjectWithBlobs()
     var myObject = Fixtures.GenerateSimpleObject();
     myObject["blobs"] = Fixtures.GenerateThreeBlobs();
 
-    var sentObjectId = await Operations.Send(myObject, new List<ITransport> { transport }).ConfigureAwait(false);
+    var sentObjectId = await Operations.Send(myObject, transport, false).ConfigureAwait(false);
```
JR-Morgan, line 72: Bit concerned here that we were previously using the default SQLite transport... If so, that wouldn't actually test our server transport.

Am I ok to assume this wasn't the intention? @gjedlicska , and I'm right to change this test?

JR-Morgan:
I would have liked to make this change 100% non-breaking, allowing devs currently using Send to have their code still compile and work 100%.

This will be true for most cases, there are a couple permutation of positional and keyword args where devs will now see a compiler error saying "parameter mismatch" or "Could not find suitable method, candidates are...".

image

Unless we are prepared to change the name of our new `Send` function, this is unavoidable.

I would suggest that this shouldn't be a blocker. Any changes to overloads are technically breaking when you look at various positional keyword args. We have made far more breaking changes only last release with the inclusion of some extra functions in `ISpeckleConverter`,

---

To get a flavour for how likely this is to affect users. All existing uses of `send` within our codebase (with the exception of a single one in our integration tests) were no breaking.

Core/Core/Api/Operations/Operations.Send.cs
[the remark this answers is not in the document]
JR-Morgan: @gjedlicska I've implemented the changes we discussed, + a couple of other minor tweaks following Alans suggestion.

The single transport overload now accepts a `UseDefaultCache`. 
We've made it non-optional for a couple reasons.
    1. We want caller to be explicit about the transports that are being used when sending.
    2. It helps make the difference between the two overloads more explicit, and prevent misunderstandings
    3. It also means our new overloads conflict less with existing ones, aiding backwards compat.
    
The overload that takes multiple transports has been renamed to `Send` to make the two overloads super clear.

Core/Core/Api/Operations/Operations.Send.cs
JR-Morgan: I've changed it again.
The single transport version is truly nullable, with the `useDefaultCache` bool exposed. Now the null check is a feature, not a bug. And with it the DX should be improved.

gjedlicska (approved):

AlanRynne (commented):
Reviewed this live with Jedd on Friday. Changes look good and I'd accept the slight breaking change due to parameter mismatch or ambiguous call

BovineOx (approved):

Core/IntegrationTests/ServerTransportTests.cs
JR-Morgan: Gergo confirmed that the new behaviour is correct (potentially before we were relying on cache missess to properly cover)

PR #3095, opened 2023-12-04
enhancement(Navisworks): Add global unhandled exception handler

jsdbroughton:
Each plugin is in an isolated Worker Thread (until it isn't) but despite this, there are `currentdomain` errors that can trigger the Navisworks Application error handler kicking in, even in debug.

In this specific case, I'm looking to head off errors arising from the Main Thread invocation that can fail to capture with the InvokerHelper wrappers I have employed; these are almost exclusively Interop errors and get caught by the hostApp global error handler. As such, these errors are on the main thread.

I intend to see what this can learn. This logs the fatal error and then rethrows to ensure no change in operation and then learns from the logging, which should be more contextual than relying on crash dumps, which, more often than not, do not deserialise innerException stack traces.


jsdbroughton:
I am parking this indefinitely - there are still circumstances where the host application event handler kicks in - the number of contexts where it will be useful have been reduced by the work done by others in Core and DUI

PR #3096, opened 2023-12-04
Fix CA1806 for Revit

connorivy:
## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertGeometry.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertGeometry.cs
@@ -1259,7 +1259,12 @@ public partial class ConverterRevit
             // Update trim indices with new item.
             // TODO: Make this better.
             var trimIndices = sEdge.TrimIndices.ToList();
-            trimIndices.Append(sTrimIndex); //TODO Append is a pure function and the return is unused
+
+            // suppressing the warning even though it is completely valid.
+            // reason we are just suppressing is because this entire class is broken anyways
+#pragma warning disable CA1806 // Do not ignore method results
+            trimIndices.Append(sTrimIndex);
+#pragma warning restore CA1806 // Do not ignore method results
             sEdge.TrimIndices = trimIndices.ToArray();
           }
         }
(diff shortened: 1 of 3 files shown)


BovineOx (approved):
Seems fine, couple of comments but mostly questions, assuming they yield nothing worrying, approved...

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertGeometry.cs
```diff
@@ -1259,7 +1259,12 @@ public Brep BrepToSpeckle(Solid solid, Document d, string units = null)
             // Update trim indices with new item.
             // TODO: Make this better.
             var trimIndices = sEdge.TrimIndices.ToList();
-            trimIndices.Append(sTrimIndex); //TODO Append is a pure function and the return is unused
+
+            // suppressing the warning even though it is completely valid.
+            // reason we are just suppressing is because this entire class is broken anyways
+#pragma warning disable CA1806 // Do not ignore method results
+            trimIndices.Append(sTrimIndex);
```
BovineOx, line 1266: This might be true, but as this line isn't doing anything surely we need either a Add or AddRange() here?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertRebar.cs
```diff
@@ -61,27 +61,31 @@ public ApplicationObject RebarToNative(RevitRebarGroup speckleRebar)
       speckleRebar.shape.rebarType == RebarType.Standard ? RebarStyle.Standard : RebarStyle.StirrupTie;
 
     // get start and end hooks and orientations
-    var speckleStartHook = speckleRebar.startHook as RevitRebarHook;
     RebarHookType startHook = null;
     RebarHookOrientation startHookOrientation = RebarHookOrientation.Right;
-    if (speckleStartHook != null)
+    if (speckleRebar.startHook is RevitRebarHook speckleStartHook)
     {
       startHook = RebarHookToNative(speckleStartHook);
```
BovineOx, line 68: We expect RebarHookToNative to return something but the if the TryParse() fails the startHook is left as .Right - is this sensible? Would we perhaps eschew ding the conversion if we cannot correctly classify the hook? I think maybe this is best as new Speckle data might have hooks older plugins cannot parse?

BovineOx: It also begs the question as to how we see such micro failures - we _expect_ speckleStartHook.orientation, to be a type we understand but if it is not we don't do anything about it...

BovineOx (commented):
A few things to consider, particularly whether to do the conversion if we cannot parse the hook type out..

PR #3102, opened 2023-12-11
fix(core): CNX-8400 grasshopper unable to send to fe2 addresses

JR-Morgan:
wip - See comments on Jira ticket

### DIFF
--- a/DesktopUI2/DesktopUI2/ViewModels/StreamViewModel.cs
+++ b/DesktopUI2/DesktopUI2/ViewModels/StreamViewModel.cs
@@ -271,17 +271,8 @@ public class StreamViewModel : ReactiveObject, IRoutableViewModel, IDisposable
 
       Branches = await Client.StreamGetBranchesWithLimitRetry(Stream.id, 0).ConfigureAwait(true);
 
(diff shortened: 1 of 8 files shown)


DesktopUI2/DesktopUI2/ViewModels/StreamViewModel.cs
```diff
@@ -271,17 +271,8 @@ internal async void GetBranchesAndRestoreState()
 
       Branches = await Client.StreamGetBranchesWithLimitRetry(Stream.id, 0).ConfigureAwait(true);
 
-      //TODO: Core's API calls and the StreamWrapper class need to be updated to properly support FE2 links
-      //this is a temporary workaround
-      var index = -1;
```
AlanRynne, line 276: Is this no longer a temporary workaround?

AlanRynne (changes requested):
Some considerations to be had. I set it as "request changes" but we may end up not making any changes to this.

I'll try to do some Grasshopper tests of this today and confirm.

AlanRynne:
Closing after discussions with @gjedlicska and @JR-Morgan. Server-side fix is on the way.

PR #3103, opened 2023-12-11
feat(Archicad): Support for Archicad quantity properties 

jozseflkiss:
## Description & motivation

https://spockle.atlassian.net/browse/CNX-8153

## Changes:

Two send options have been introduced:
* Send Properties: sends Properties created in the Property Manager
* Send Listing Parameters: sends general and tool-specific parameters available for the fields of Element Schedules
  image

Properties are exported on:
* element-level
  image
* component-level
  image

On element level:
* Properties: user defined properties attached to element driven by classifications
* Listing Parameters: almost all parameters of the General Parameters group
are exported.

On component level:
* Properties: user defined properties attached to Building Material driven by classifications
* all properties of the Components and Building Material groups
are exported.

## Code

Objects:
* new fields introduced for BIM classes
```
    public Base? elementProperties { get; set; }
    public Base? componentProperties { get; set; }
```
* new `Property` classes: transform JSON values of "name" and "value" keys, to key-value pairs

Connector:
* `GetData... ` command classes have a new base class which manages export options

Archicad add-on:
* PropertyExportManager: singleton class which manages property definitions
** It caches filtered property definitions based on classification and element type
* Can filter by property groups, property guids etc.

## Checklist:

### DIFF
--- a/Objects/Objects/BuiltElements/Archicad/Property.cs
+++ b/Objects/Objects/BuiltElements/Archicad/Property.cs
@@ -4,21 +4,6 @@ using Speckle.Core.Models;
 
 namespace Objects.BuiltElements.Archicad;
 
-public static class PropertyHelper
-{
-  private static readonly HashSet<char> DisallowedPropNameChars = new() { '.', '/' };
-
-  public static string RemoveDisallowedPropNameChars(string propertyName)
-  {
-    foreach (char c in DisallowedPropNameChars)
-    {
-      propertyName = propertyName.Replace(c, ' ');
-    }
-
-    return propertyName;
-  }
-}
-
 public class Property : Base
 {
   public Property() { }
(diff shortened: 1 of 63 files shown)


Objects/Objects/BuiltElements/Beam.cs
```diff
@@ -136,6 +136,12 @@ public class ArchicadBeam : Beam
     public string? elementType { get; set; } /*APINullabe*/
     public List<Classification>? classifications { get; set; } /*APINullabe*/
 
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public Base? elementProperties { get; set; }
+
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public Base? componentProperties { get; set; }
```
AlanRynne, line 143: Do we respect `NullValueHandling` in our serialiser?

If I understand this correctly, this causes the property to not be added to the `json` when null, but that may be counter productive for data analysis scenarios like Excel and PowerBI where we don't have an object model available.

Is this necessary for some reason?

Objects/Objects/BuiltElements/Archicad/Property.cs
```diff
@@ -0,0 +1,162 @@
...
+
+  public static string RemoveDisallowedPropNameChars(string propertyName)
+  {
+    foreach (char c in DisallowedPropNameChars)
+    {
+      propertyName = propertyName.Replace(c, ' ');
+    }
+
+    return propertyName;
+  }
```
AlanRynne, line 19: It would be interesting if we can coordinate with @clairekuang on how disallowed characters are dealt with in Rhino/Acad. And possibly align this?

This doesn't seem to be strictly "property related" but rather part of `Base` right? Would it make sense to discuss aligning this at the SDK level?

Objects/Objects/BuiltElements/Archicad/Property.cs
AlanRynne, line 1: It'll make me much happier if we could move each class into it's own file 🙏🏼

I know this is not what we've been doing so far (at all 🙇🏼‍♂️) but we've gotta start somewhere... 😄

AlanRynne (changes requested):
I'm not going to pretend I know anything about the `cpp` side of things... but I left some comments on the Objects side 🙌🏼

Objects/Objects/BuiltElements/Beam.cs
jozseflkiss: The reason is not to export empty null values. OK, since this info is relevant for PowerBI, I removed that NullValueHandling.

Objects/Objects/BuiltElements/Archicad/Property.cs
jozseflkiss: As a first step I moved this function to Core.

Objects/Objects/BuiltElements/Archicad/Property.cs
jozseflkiss: Done.

Objects/Objects/BuiltElements/Beam.cs
AlanRynne: Perfect thx!

Objects/Objects/BuiltElements/Archicad/Property.cs
AlanRynne: Ok i'll open a ticket for this and ping you guys there

AlanRynne (approved):
Approving after Jozsefs last changes :)

PR #3105, opened 2023-12-12
chore(navisworks): Simplify methods to consistently using lambda expressions

jsdbroughton:


AlanRynne (approved):
I accept these very neat changes ♥️

PR #3106, opened 2023-12-12
Fix(Revit) : CNX-8387 ca1031 don't catch general exceptions

connorivy:
## Changes:

- when it was obvious which exception would be thrown, catch statement was altered to that exception
- when it was not obvious, seq logs were checked
  - if exception was being logged and had not occured in the last few months, the try-catch was removed
  - if exception was not being logged, seq logging was added

## Checklist:


Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -1236,16 +1244,20 @@ public bool TryAppendLineSafely(CurveArray curveArray, Line line, ApplicationObj
...
     }
+#pragma warning disable CA1031 // Do not catch general exception types
     try
     {
       curveArray.Append(LineToNative(line));
       return true;
     }
-    catch (Exception e)
+    catch (Exception ex)
     {
```
AlanRynne, line 1254: Another one I'm not sure we should silence.

If the issue is that we don't have a known exception that can be thrown from `LineToNative`, then we should pull on that thread and find out.

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.DxfImport.cs
```diff
@@ -239,11 +240,17 @@ public DB.Element CreateDxfImport(List<Base> objects, string fileName, double sc
     var symbol = Doc.GetElement(fam.GetFamilySymbolIds().First()) as DB.FamilySymbol;
     symbol.Activate();
 
+#pragma warning disable CA1031 // Do not catch general exception types
     try
     {
       File.Delete(tempFamilyPath);
```
AlanRynne, line 246: Same comment on `File.Delete` as above: it throws known exceptions so we should catch the ones we can recover from, if any.

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.Previews.cs
```diff
@@ -24,15 +26,19 @@ private ApplicationObject PreviewGeometry(Base @object)
...
     try
     {
       server = new DirectContext3DServer(meshes, Doc);
       appObj.Update(status: ApplicationObject.State.Created);
     }
     catch (Exception ex)
     {
+      // TODO : check if catch statement is necessary
+      SpeckleLog.Logger.LogDefaultError(ex);
       appObj.Update(status: ApplicationObject.State.Failed, logItem: ex.Message);
```
AlanRynne, line 39: I fail to see how this would ever throw unless `appObj` is null

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertAdaptiveComponent.cs
```diff
@@ -66,7 +66,7 @@ public ApplicationObject AdaptiveComponentToNative(AdaptiveComponent speckleAc)
 
         isUpdate = true;
       }
-      catch
+      catch (Autodesk.Revit.Exceptions.ApplicationException)
       {
         //something went wrong, re-create it
       }
```
AlanRynne, line 72: If we fail here, should we notify the user (i.e update the appObj log) or just fail.

If this fails, I'm assuming the user would end. up with 2 adaptive components in the document instead of 1?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertCeiling.cs
```diff
@@ -126,7 +125,7 @@ public ApplicationObject CeilingToNative(Ceiling speckleCeiling)
     {
       CreateVoids(revitCeiling, speckleCeiling);
     }
-    catch (Exception ex)
+    catch (Autodesk.Revit.Exceptions.ApplicationException ex)
     {
       appObj.Update(logItem: $"Could not create openings: {ex.Message}");
```
AlanRynne, line 130: If we fail to create something, shouldn't we also update the status of the appObj to be failed?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertColumn.cs
```diff
@@ -111,7 +111,10 @@ public ApplicationObject ColumnToNative(Column speckleColumn)
         }
         isUpdate = true;
       }
-      catch { }
+      catch (Autodesk.Revit.Exceptions.ApplicationException)
+      {
+        //something went wrong, re-create it
+      }
```
AlanRynne, line 117: same update behaviour, we should double check what happens if this fails unexpectedly

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertConnector.cs
```diff
@@ -22,18 +22,18 @@ public RevitMEPConnector ConnectorToSpeckle(Connector connector)
     {
       speckleMEPConnector.angle = connector.Angle;
     }
-    catch { }
+    catch (Autodesk.Revit.Exceptions.InvalidOperationException) { }
```
AlanRynne, line 25: If these are well known errors that we can safely recover from, could we at least have a comment explaining why is this ok on each case?

Could this be prevented if we re-wrote this logic a bit?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertFaceWall.cs
```diff
@@ -58,11 +60,17 @@ public ApplicationObject FaceWallToNative(RevitFaceWall speckleWall)
...
     try
     {
       File.Delete(tempMassFamilyPath);
     }
-    catch { }
+    catch (Exception ex)
+    {
+      // TODO : check if catch block is necessary
+      SpeckleLog.Logger.LogDefaultError(ex);
+    }
```
AlanRynne, line 72: Same fixable `Delete` generic catch exception

AlanRynne (changes requested):
I guess there's still a lot we need to figure out, and since your PR is the first... we're hitting a bunch of existential questions.

Only half way there but I'm taking a break for the day and didn't want to loose this. I'll do a review of the other half tomorrow

ConnectorRevit/ConnectorRevit/Entry/App.cs
```diff
@@ -35,7 +36,7 @@ public Result OnStartup(UIControlledApplication application)
     {
       application.CreateRibbonTab(tabName);
     }
-    catch { }
+    catch (Autodesk.Revit.Exceptions.ApplicationException) { }
```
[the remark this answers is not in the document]
JR-Morgan: While catching `ApplicationException` is a marked improvement over `Exception`. If we can be more specific, lets do it!

And a perfect example of that, a `ArgumentNullException` would be a surprise to us, so we don't want to handle it here. We take responsibility for ensuring we don't pass null through. If that assumption doesn't hold (maybe someone changes the code and we start passing nulls around), we want that to fail loud! else we may not notice.

BovineOx (commented):
I think we need to provide a bit more guidance @connorivy so I'll Leave it there :)

ConnectorRevit/ConnectorRevit/Storage/StreamStateManager.cs
```diff
@@ -36,10 +39,13 @@ public static List<StreamState> ReadState(Document doc)
...
       return states;
     }
-    catch (Exception e)
+    catch (Exception ex)
     {
+      // TODO : check if catch statement is necessary
+      SpeckleLog.Logger.LogDefaultError(ex);
       return new List<StreamState>();
     }
+#pragma warning restore CA1031 // Do not catch general exception types
```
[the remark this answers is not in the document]
JR-Morgan: I see a lot of areas where we can improve this `StreamStateManager` (and the other 3 `StreamStateManager`s that exist in other connectors 😅)
Perhaps we could take a look at a refactor across them all? or perhaps with DUI3, these will need to be re-written anyway.

ConnectorRevit/ConnectorRevit/Entry/App.cs
```diff
@@ -154,28 +166,15 @@ Autodesk.Revit.DB.Events.ApplicationInitializedEventArgs e
 
       //AppInstance.ViewActivated += new EventHandler<ViewActivatedEventArgs>(Application_ViewActivated);
     }
+    catch (KitException ex)
+    {
+      SpeckleLog.Logger.Warning(ex, "Error loading kit on startup");
+      NotifyUserOfErrorStartingConnector(ex);
+    }
     catch (Exception ex)
```
BovineOx, line 174: just reference CNX-8644 here

ConnectorRevit/ConnectorRevit/Entry/SpeckleRevitCommand.cs
BovineOx, line 1: this is so much cleaner now

ConnectorRevit/ConnectorRevit/Storage/StreamStateManager.cs
```diff
@@ -36,8 +43,10 @@ public static List<StreamState> ReadState(Document doc)
 
       return states;
     }
-    catch (Exception e)
+    catch (Exception ex)
```
BovineOx, line 46: Let's use the !IsFatal(ex) pattern and remove the TODO we will see what exception turn up as we logging it

ConnectorRevit/ConnectorRevit/TypeMapping/ElementTypeMapper.cs
```diff
@@ -219,9 +220,12 @@ await Dispatcher.UIThread
       }
       catch (Exception ex)
```
BovineOx, line 221: use the !IsFatal(ex) pattern and you can have the warning on then

ConnectorRevit/ConnectorRevit/TypeMapping/ElementTypeMapper.cs
```diff
@@ -348,15 +352,18 @@ string category
...
         },
       };
+#pragma warning disable CA1031 // Do not catch general exception types
       try
       {
         previousMappingExists = true;
         return JsonConvert.DeserializeObject<TypeMap>(mappingSetting.MappingJson, settings);
       }
-      catch
+      catch (Exception ex)
```
BovineOx, line 361: use the !IsFatal(ex) pattern and you can have the warning on then

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Receive.cs
```diff
@@ -197,7 +201,7 @@ await elementTypeMapper
           {
             CurrentDoc.Document.Delete(elementToDelete.Id);
           }
-          catch
+          catch (Autodesk.Revit.Exceptions.ApplicationException)
           {
             // unable to delete previously recieved object
```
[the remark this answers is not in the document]
BovineOx: Agree at the least let's log something, we are essentially skipping somehting, it should go in the report

ConnectorRevit/RevitSharedResources/Helpers/Extensions.cs
```diff
@@ -47,14 +47,12 @@ public static bool IsPhysicalElement(this Element e)
   /// <remarks>This function will never throw, returning false instead</remarks>
   public static bool HasCategory(this IEnumerable<BuiltInCategory> categories, Category category)
   {
-    try
-    {
-      return categories.Select(x => (int)x).Contains(category.Id.IntegerValue);
-    }
-    catch (Exception e)
+    if (category?.Id?.IntegerValue is not int categoryInt)
```
BovineOx, line 50: hmm not sure that indemnifies us from the cast below?

ConnectorRevit/RevitSharedResources/Models/ErrorEater.cs
```diff
@@ -40,7 +40,7 @@ public FailureProcessingResult PreprocessFailures(FailuresAccessor failuresAcces
         failuresAccessor.ResolveFailure(failure);
         resolvedFailures++;
       }
-      catch
+      catch (Autodesk.Revit.Exceptions.ApplicationException)
```
BovineOx, line 43: We are logging Fatal, I guess is this Fatal for the plugin? How does the host app know the plugin is not healthy?

ConnectorRevit/RevitSharedResources/Models/TransactionManager.cs
```diff
@@ -186,7 +186,7 @@ public static TResult ExecuteInTemporaryTransaction<TResult>(Func<TResult> funct
         t.Start();
         result = function();
       }
-      catch
+      catch (Autodesk.Revit.Exceptions.ApplicationException)
       {
         // ignore because we're just going to rollback
```
BovineOx, line 191: But what happened, don't we need something to appear in the report?

ConnectorRevit/RevitSharedResources/Models/TransactionManager.cs
```diff
@@ -203,7 +203,7 @@ public static TResult ExecuteInTemporaryTransaction<TResult>(Func<TResult> funct
...
```
BovineOx, line 208: and the same :up

DesktopUI2/DesktopUI2/ViewModels/TypeMappingOnReceiveViewModel.cs
```diff
@@ -28,7 +28,7 @@ public TypeMappingOnReceiveViewModel(ITypeMap typeMap, IHostTypeContainer contai
   {
     Mapping = typeMap;
     hostTypeContainer = container;
-    SelectedCategory = Mapping.Categories.First();
+    SelectedCategory = Mapping.Categories.FirstOrDefault() ?? TypeCatMisc;
```
BovineOx, line 31: This now being null is goin to be okay?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertBeam.cs
```diff
@@ -56,7 +56,7 @@ public ApplicationObject BeamToNative(Beam speckleBeam, StructuralType structura
     {
       if (level != null)
       {
-        level = GetLevelByName(speckleRevitBeam.level.name);
+        level = GetLevelByName(speckleRevitBeam.level?.name);
```
BovineOx, line 59: maybe: level = level ?? GetLevelByName(...)

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertBlock.cs
```diff
@@ -59,7 +59,7 @@ public ApplicationObject BlockInstanceToNative(BlockInstance instance)
         isUpdate = true;
         Doc.Delete(docObj.Id);
       }
-      catch (Exception e)
+      catch (Autodesk.Revit.Exceptions.ApplicationException e)
```
BovineOx, line 62: and continuing will be okay...

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertCurves.cs
```diff
@@ -74,7 +76,7 @@ public ApplicationObject DetailCurveToNative(DetailCurve speckleCurve)
       {
         revitCurve = Doc.Create.NewDetailCurve(Doc.ActiveView, baseCurve);
       }
-      catch (Exception)
+      catch (Autodesk.Revit.Exceptions.ApplicationException)
```
BovineOx, line 79: If this fails at all, is the whole DetailCurveToNative() broken? Could it be creating an incorrect native object?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertCurves.cs
```diff
@@ -140,12 +142,15 @@ public ApplicationObject ModelCurveToNative(ICurve speckleLine)
       return appObj;
     }
 
+#pragma warning disable CA1031 // Do not catch general exception types
     try
     {
       return ModelCurvesFromEnumerator(CurveToNative(speckleLine).GetEnumerator(), speckleLine, appObj);
     }
-    catch (Exception e)
+    catch (Exception ex)
```
BovineOx, line 150: Should be the !IsFatal(ex) pattern

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertDirectShape.cs
```diff
@@ -230,8 +233,11 @@ public ApplicationObject DirectShapeToNative(DirectShape speckleDs, ToNativeMesh
     }
     catch (Exception ex)
     {
+      // TODO : check if catch block is necessary
```
BovineOx, line 236: None of these conversions should be catching Exception, so we should at least have the !IsFatal() pattern and the warning should not be disabled. Reserving judgement on LogDefaultError() but it would be good to know WHY we throw exceptions here as we can then provide a better log message.

I won;t keep logging the same for these conversisons, but it stants for each place - we need to use the !IsFatal() pattern here, log, and not suppress the warning.

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -153,9 +155,11 @@ public void GetHostedElements(Base @base, Element host, out List<string> notes)
         }
         catch (Exception ex)
```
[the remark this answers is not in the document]
BovineOx: A the least it should follow the IsFatal() pattetn

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.Previews.cs
BovineOx: Presumably in DirectContext3DServer() ? if so what can that throw? Is this a top level handler?

Objects/Converters/ConverterRevit/ConverterRevitShared/ConverterRevit.cs
```diff
@@ -407,7 +407,7 @@ public Base ConvertToSpeckle(object @object)
         var material = GetElementRenderMaterial(@object as DB.Element);
         returnObject["renderMaterial"] = material;
       }
-      catch (Exception e)
+      catch (Autodesk.Revit.Exceptions.ApplicationException)
       {
         // passing for stuff without a material (eg converting the current document to get the `Model` and `Info` objects)
```
BovineOx, line 412: Should we at least be logging something here?

Objects/Converters/ConverterRevit/ConverterRevitShared/DirectContext3DServer.cs
```diff
@@ -134,7 +134,7 @@ public void RenderScene(View dBView, DisplayStyle displayStyle)
           );
         }
       }
-      catch (Exception e)
+      catch (Autodesk.Revit.Exceptions.ApplicationException e)
       {
         System.Diagnostics.Debug.WriteLine(e.ToString());
```
BovineOx, line 139: Presume this should come out in favour of actual logging

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitVersionHelper.cs
```diff
@@ -97,7 +97,7 @@ public static bool IsCurveClosed(NurbSpline curve)
     {
       return curve.IsClosed;
     }
-    catch
+    catch (Autodesk.Revit.Exceptions.ApplicationException)
     {
       return true;
```
BovineOx, line 102: Is this a resonable return value? why is this going to be thrown?

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitVersionHelper.cs
```diff
@@ -123,7 +123,7 @@ public static bool IsCurveClosed(Curve curve)
     {
       return curve.IsClosed;
     }
-    catch
+    catch (Autodesk.Revit.Exceptions.ApplicationException)
```
BovineOx, line 126: as above?

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/SpeckleConversionTest.cs
```diff
@@ -194,6 +195,7 @@ await APIContext
           {
             converter.Report.LogConversionError(new Exception(e.Message, e));
           }
+#pragma warning restore CA1031 // Do not catch general exception types
```
BovineOx, line 198: As with other statements, if this is not top level handler, then !IsFatal() and log

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/SpeckleUtils.cs
```diff
@@ -46,6 +47,7 @@ internal static class SpeckleUtils
       {
         tcs.TrySetException(exception);
       }
+#pragma warning restore CA1031 // Do not catch general exception types
```
BovineOx, line 50: Not sure about this one, possible we consider it a TL handler because of what it is? Not convinced though

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertConnector.cs
connorivy: yes, refactored

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Receive.cs
```diff
@@ -132,41 +127,32 @@ await elementTypeMapper
...
-          string message = $"Fatal Error: {ex.Message}";
-          if (ex is OperationCanceledException)
-          {
-            message = "Receive cancelled";
-          }
-
-          progress.Report.LogOperationError(new Exception($"{message} - Changes have been rolled back", ex));
-
           transactionManager.RollbackAll();
-          return (false, ex); //We can't throw exceptions in from RevitTask, but we can return it along with a success status
```
JR-Morgan, line 150: @connorivy we were doing this (returning rather than throwing) for a reason. Is this no longer needed since we switched away from `RevitTask`?

ConnectorRevit/RevitSharedResources/Models/ErrorEater.cs
JR-Morgan: I think we'd prefer to keep `Fatal` for cases where we are unsure if we've fully recovered (i.e. may have left DUI or the host app in an invalid state).

If this handler allows us to safely recover (i.e. we may have failed receiving, but we've properly rolled back), then I think I'd suggest `Error` level.
But we've historically been a little lossy goosy with Fatal vs Error.

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/SpeckleUtils.cs
JR-Morgan: Assuming we ensure the task completion source is always observed. I think its safe to catch all here.
However, to be extra safe. In the few places of Core where we are using TaskCompletionSources, I'm using this pattern 

https://github.com/specklesystems/speckle-sharp/blob/d0952efb1965bbd75e64ddebd828e86c368b2290/Core/Core/Serialisation/DeserializationWorkerThreads.cs#L51-L63C6

 we may consider doing something similar here.

ConnectorRevit/RevitSharedResources/Models/ErrorEater.cs
JR-Morgan: That being said.. I'm a bit puzzled why we're using a Task completion source here anyway. since we await the APIContext call anyway...

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/SpeckleUtils.cs
JR-Morgan: That being said.. I'm a bit puzzled why we're using a Task completion source here anyway. since we await the APIContext call anyway...

ConnectorRevit/ConnectorRevit/Storage/StreamStateManager.cs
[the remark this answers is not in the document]
BovineOx: I think it needs to be a consideration for new architecture. We could do with a Architecture Improvements Parent page where we can capture projects.

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Events.cs
```diff
@@ -24,17 +24,32 @@ public partial class ConnectorBindingsRevit
...
+        })
+        .ConfigureAwait(false);
+    }
+    // note : I don't think this exception catch is necessary, but this is an async VOID instead
+    // of an async Task. Revit WILL CRASH if this method tries to throw an error, so keeping this
+    // catch out of an abundance of caution
+    catch (Exception ex) when (!ex.IsFatal())
+    {
+      SpeckleLog.Logger.Fatal(
+        ex,
```
AlanRynne, line 47: Perfect, this looks good to me. We can start checking SEQ for this and determine if something else is required later on.

ConnectorRevit/RevitSharedResources/Models/TransactionManager.cs
AlanRynne: I think Connor makes a valid point. But ideally we'd want to know if something went wrong in a sub-transaction.

I wouldn't change the existing behaviour right now, but given this is in the generic `TransactionManager`, having a way to know if a temp transaction went wrong (and what went wrong exactly) seems useful.

Maybe the solution here would be to capture the exception, rollback the transaction in a `finally` catch and throw right after (though this also sounds admittedly hacky)

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -241,16 +238,18 @@ public void GetAllRevitParamsAndIds(Base speckleElement, DB.Element revitElement
...
       }
-      catch
+      catch (InvalidPropNameException)
+      {
+        //ignore
+      }
+      catch (SpeckleException)
       {
         //ignore
       }
```
AlanRynne, line 252: Maybe we should add some logs here, specially if we're swallowing the exception.

In an ideal world failing to get "some" parameters should be reported to the user, but this is not as straightforward and somewhat out of scope for this ticket.

A log call would do here just fine.

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/SpeckleConversionTest.cs
```diff
@@ -186,16 +187,14 @@ await APIContext
...
-#pragma warning disable CA1031 // Do not catch general exception types
           try
           {
             res = converter.ConvertToNative(el);
           }
-          catch (Exception e)
+          catch (Exception ex) when (!ex.IsFatal())
           {
-            converter.Report.LogConversionError(new Exception(e.Message, e));
+            converter.Report.LogConversionError(new Exception(ex.Message, ex));
```
AlanRynne, line 196: If we're not adding any new information to the exception, we might as well just log that instead, as this just duplicates the information in the message.

LogConversionError will take care of the rest.

AlanRynne (approved):
Approving with some minor comments!

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
connorivy: Added log call in the "SpeckleException" catch. Do we also want one in the InvalidPropName catch? This is thrown whenever we try to assign a dynamic property with invalid characters (e.g. var x = base["hello-world"]) and this happens all the time for these properties.

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/SpeckleConversionTest.cs
connorivy: Good point

Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
AlanRynne: Ok if this happens all the time, then maybe logging `InvalidPropName` is a bit overkill.

The log in speckle exception is good enough 😄

ConnectorRevit/RevitSharedResources/Models/TransactionManager.cs
AlanRynne: I think we should open a ticket to refactor this in a way that it could throw, but right now just adding it to the logs should be useful enough for us.

AlanRynne (approved):
All the last proposed changes look good. I'm merging this in.

ConnectorRevit/RevitSharedResources/Models/ErrorEater.cs
AlanRynne: I've lowered this log to be an error instead

PR #3108, opened 2023-12-12
fix(sln): Added postProject build dependencies for missing connector

JR-Morgan:
I noticed several connector projects in our All.sln were missing the indirect build dependency on their converter. 
This is needed so when you make changes to the converter, and then build/run the connector, the converter should also be rebuilt.


AlanRynne (approved):
I can only say HELL YES! to this 🙌🏼

Super good catch @JR-Morgan 👀

PR #3109, opened 2023-12-13
Chore(core): CNX-7435 - Resolved even more core warnings

JR-Morgan:
Main changes with this PR:
 - Supressed warnings on obsolete functions (since they are effectively "resolved")
 - Cleared up the last the "easy to fix" warnings in core. The rest need some level of planning (although to varying degrees)
   Just 274 left!
 - I've made nullability syntax default in core since we had more exceptions (classes with `#nullable enabled` than we had implicitly disabled), this should also encourage us to design new classes with clear nullability in mind

### DIFF
--- a/Core/Core/Kits/Attributes.cs
+++ b/Core/Core/Kits/Attributes.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 
 namespace Speckle.Core.Kits;
@@ -8,7 +9,7 @@ public sealed class SchemaInfo : Attribute
   public SchemaInfo(string name, string description)
     : this(name, description, null, null) { }
 
-  public SchemaInfo(string name, string description, string category, string subcategory)
+  public SchemaInfo(string name, string description, string? category, string? subcategory)
   {
     Name = name;
     Description = description;
@@ -16,9 +17,9 @@ public sealed class SchemaInfo : Attribute
     Subcategory = subcategory;
   }
 
-  public string Subcategory { get; }
+  public string? Subcategory { get; }
 
-  public string Category { get; }
+  public string? Category { get; }
 
   public string Description { get; }
 
--- a/Core/Core/Kits/Attributes.cs
+++ b/Core/Core/Kits/Attributes.cs
@@ -1,3 +1,4 @@
+#nullable disable
 using System;
 
 namespace Speckle.Core.Kits;
@@ -8,7 +9,7 @@ public sealed class SchemaInfo : Attribute
   public SchemaInfo(string name, string description)
     : this(name, description, null, null) { }
 
-  public SchemaInfo(string name, string description, string? category, string? subcategory)
+  public SchemaInfo(string name, string description, string category, string subcategory)
   {
     Name = name;
     Description = description;
@@ -16,9 +17,9 @@ public sealed class SchemaInfo : Attribute
     Subcategory = subcategory;
   }
 
-  public string? Subcategory { get; }
+  public string Subcategory { get; }
 
-  public string? Category { get; }
+  public string Category { get; }
 
   public string Description { get; }
 
--- a/Core/Core/Kits/Attributes.cs
+++ b/Core/Core/Kits/Attributes.cs
@@ -1,4 +1,3 @@
-#nullable enable
 using System;
 
 namespace Speckle.Core.Kits;
--- a/Core/Core/Kits/ISpeckleConverter.cs
+++ b/Core/Core/Kits/ISpeckleConverter.cs
@@ -1,3 +1,4 @@
+#nullable disable
 using System.Collections.Generic;
 using Speckle.Core.Models;
 
--- a/Core/Core/Kits/ISpeckleConverter.cs
+++ b/Core/Core/Kits/ISpeckleConverter.cs
@@ -23,37 +23,41 @@ public interface ISpeckleConverter
   /// <summary>
   /// Converts a native object to a Speckle one
   /// </summary>
-  /// <param name="object">Native object to convert</param>
+  /// <param name="value">Native object to convert</param>
   /// <returns></returns>
-  public Base ConvertToSpeckle(object @object);
+  /// <exception cref="System.Exception"></exception>
+  public Base ConvertToSpeckle(object value);
 
   /// <summary>
   /// Converts a list of objects to Speckle.
   /// </summary>
-  /// <param name="objects"></param>
+  /// <param name="values"></param>
   /// <returns></returns>
-  public List<Base> ConvertToSpeckle(List<object> objects);
+  /// <exception cref="System.Exception"></exception>
+  public List<Base> ConvertToSpeckle(List<object> values);
 
   /// <summary>
   /// Checks if it can convert a native object to a Speckle one
   /// </summary>
-  /// <param name="object">Native object to convert</param>
+  /// <param name="value">Native object to convert</param>
   /// <returns></returns>
-  public bool CanConvertToSpeckle(object @object);
+  public bool CanConvertToSpeckle(object value);
 
   /// <summary>
   /// Converts a Speckle object to a native one
   /// </summary>
-  /// <param name="object">Speckle object to convert</param>
+  /// <param name="value">Speckle object to convert</param>
   /// <returns></returns>
-  public object ConvertToNative(Base @object);
+  /// <exception cref="System.Exception"></exception>
+  public object ConvertToNative(Base value);
 
   /// <summary>
   /// Converts a list of Speckle objects to a native ones.
   /// </summary>
-  /// <param name="objects"></param>
+  /// <param name="values"></param>
   /// <returns></returns>
-  public List<object> ConvertToNative(List<Base> objects);
+  /// <exception cref="System.Exception"></exception>
+  public List<object> ConvertToNative(List<Base> values);
 
   /// <summary>
   /// Converts a given speckle objects as a generic native object.
@@ -68,16 +72,16 @@ public interface ISpeckleConverter
   /// In this case, deciding which to use is dependent on each connector developer.
   /// Preferably, <see cref="ConvertToNativeDisplayable"/> should be used as a fallback to the <see cref="ConvertToNative(Speckle.Core.Models.Base)"/> logic.
   /// </remarks>
-  /// <param name="object">Speckle object to convert</param>
-  /// <returns>The native object that resulted after converting the input <paramref name="object"/></returns>
-  public object ConvertToNativeDisplayable(Base @object);
+  /// <param name="value">Speckle object to convert</param>
+  /// <returns>The native object that resulted after converting the input <paramref name="value"/></returns>
+  public object ConvertToNativeDisplayable(Base value);
 
   /// <summary>
   /// Checks if it can convert a Speckle object to a native one
   /// </summary>
-  /// <param name="object">Speckle object to convert</param>
+  /// <param name="value">Speckle object to convert</param>
   /// <returns></returns>
-  public bool CanConvertToNative(Base @object);
+  public bool CanConvertToNative(Base value);
 
   /// <summary>
   /// Checks to verify if a given object is: 1) displayable and  2) can be supported for conversion to the native application.
@@ -89,9 +93,9 @@ public interface ISpeckleConverter
   /// Preferably, <see cref="CanConvertToNativeDisplayable"/> should be used as a fallback to the <see cref="CanConvertToNative"/> logic.
   /// Objects found in the 'displayValue' property are assumed to be universally convertible by all converters and the viewer, but are not guaranteed to be so.
   /// </remarks>
-  /// <param name="object">Speckle object to convert</param>
+  /// <param name="value">Speckle object to convert</param>
   /// <returns>True if the object is "displayable" and the converter supports native conversion of the given speckle object in particular.</returns>
-  public bool CanConvertToNativeDisplayable(Base @object);
+  public bool CanConvertToNativeDisplayable(Base value);
 
   /// <summary>
   /// Returns a list of applications serviced by this converter
--- a/Core/Core/Kits/KitDeclaration.cs
+++ b/Core/Core/Kits/KitDeclaration.cs
@@ -1,3 +1,4 @@
+#nullable disable
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/Core/Core/Kits/KitDeclaration.cs
+++ b/Core/Core/Kits/KitDeclaration.cs
@@ -8,7 +8,7 @@ namespace Speckle.Core.Kits;
 /// <summary>
 /// Needed so we can properly deserialize all the Base-derived objects from Core itself.
 /// </summary>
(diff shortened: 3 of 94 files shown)


JR-Morgan:
Going to resolve the integration tests failing in the morning

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
```diff
@@ -74,7 +76,15 @@ public Base Deserialize(string rootObjectJson)
         string objId = closure.Item1;
         // pausing for getting object from the transport
         stopwatch.Stop();
-        string objJson = ReadTransport.GetObject(objId);
+        string? objJson = ReadTransport.GetObject(objId);
```
JR-Morgan, line 79: See https://spockle.atlassian.net/browse/CNX-8541

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
```diff
@@ -74,7 +76,15 @@ public Base Deserialize(string rootObjectJson)
...
+
+        //TODO: We should fail loudly when a closure can't be found
+        //but adding this throw breaks blobs tests, see CNX-8541
+        // if (objJson is null)
+        // {
+        //   throw new TransportException($"Closure {objId} was not found in transport {ReadTransport}");
```
JR-Morgan, line 85: See https://spockle.atlassian.net/browse/CNX-8541

Core/Core/Kits/Attributes.cs
```diff
@@ -8,17 +8,17 @@ public sealed class SchemaInfo : Attribute
   public SchemaInfo(string name, string description)
     : this(name, description, null, null) { }
 
-  public SchemaInfo(string name, string description, string category, string subcategory)
+  public SchemaInfo(string name, string description, string? category, string? subcategory)
```
AlanRynne, line 11: Not sure about this, all SchemaInfo's should have a category set so the component generator will pick those up.

I'll double check this but if some SchemaInfo does not have category or subcategory i'd consider that a problem :)

Core/Core/Kits/ISpeckleConverter.cs
```diff
@@ -23,37 +24,41 @@ public interface ISpeckleConverter
   /// <summary>
   /// Converts a native object to a Speckle one
   /// </summary>
-  /// <param name="object">Native object to convert</param>
+  /// <param name="value">Native object to convert</param>
   /// <returns></returns>
-  public Base ConvertToSpeckle(object @object);
+  /// <exception cref="System.Exception"></exception>
+  public Base ConvertToSpeckle(object value);
```
AlanRynne, line 30: 🙌🏼🙌🏼🙌🏼

Core/Core/Kits/KitDeclaration.cs
```diff
@@ -26,39 +27,4 @@ public ISpeckleConverter LoadConverter(string app)
...
-
-  public void SetContextDocument(object @object)
-  {
-    throw new NotImplementedException();
-  }
-
-  public bool TryLoadConverter(string app, out ISpeckleConverter speckleConverter)
-  {
-    throw new NotImplementedException();
-  }
```
AlanRynne, line 63: I don't understand how this works at all 🥲

Core/Core/Kits/Units.cs
```diff
@@ -10,35 +11,36 @@ public static class Units
   public const string Meters = "m";
   public const string Kilometers = "km";
   public const string Inches = "in";
-  public const string Feet = "ft"; // smelly ones
```
AlanRynne, line 13: 🥲 I liked this silly comment hahahaha RIP

Core/Core/Kits/Units.cs
```diff
@@ -234,109 +236,67 @@ public static double GetConversionFactor(string from, string to)
...
+      "cm" or "centimetre" or "centimeter" or "centimetres" or "centimeters" => Centimeters,
+      "m" or "meter" or "metre" or "meters" or "metres" => Meters,
+      "inches" or "inch" or "in" => Inches,
+      "feet" or "foot" or "ft" => Feet,
+      "ussurveyfeet" => USFeet, //BUG: why don't we match on "us_ft"?
+      "yard" or "yards" or "yd" => Yards,
+      "miles" or "mile" or "mi" => Miles,
+      "kilometers" or "kilometer" or "km" => Kilometers,
+      "none" => None,
+      _ => throw new ArgumentOutOfRangeException(nameof(unit), $"Unrecognised unit string {unit}"),
```
AlanRynne, line 265: can I love and hate this at the same time? 😅

That BUG comment is legit, good catch @JR-Morgan 👀 @clairekuang maybe you know the reason?

Core/Core/Logging/Setup.cs
```diff
@@ -74,4 +75,12 @@ static Setup()
...
     }
   }
+
+  [Obsolete("Use " + nameof(Mutex))]
+  [SuppressMessage("Style", "IDE1006:Naming Styles")]
+  public static Mutex mutex
+  {
+    get => Mutex;
+    set => Mutex = value;
+  }
```
AlanRynne, line 85: This is new but also obsolete? I'm assuming it's coming from some other file where it was moved, but I must have missed it.

Core/Core/Models/Utilities.cs
```diff
@@ -32,8 +31,8 @@ public static string HashString(string input, HashingFunctions func = HashingFun
   {
     return func switch
     {
-      HashingFunctions.SHA256 => Crypt.Sha256(input, end: HashLength),
-      HashingFunctions.MD5 => Crypt.Md5(input, end: HashLength),
+      HashingFunctions.SHA256 => Crypt.Sha256(input, length: HashLength),
+      HashingFunctions.MD5 => Crypt.Md5(input, length: HashLength),
```
AlanRynne, line 35: were we using the wrong prop all along? or was `end` just deprecated in favour of `length`

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
AlanRynne: Interesting! @BovineOx you should have a look at this if you haven't already

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
```diff
@@ -161,7 +176,15 @@ private List<(string, int)> GetClosures(string rootObjectJson)
...
-    object? converted = ConvertJsonElement(doc1);
+    object? converted;
+    try
+    {
+      converted = ConvertJsonElement(doc1);
+    }
+    catch (Exception ex)
+    {
+      throw new SpeckleException($"Failed to deserialize {doc1} as {doc1.Type}", ex);
+    }
```
AlanRynne, line 187: I see what you're doing here... I'm assuming this was out of scope for the current ticket and it was cleaner to wrap it in a SpeckleException and move the fix to the `Exception Warning` effort. Is this correct?

If so, we can live with it.

Core/Core/Transports/SQLite.cs
```diff
@@ -164,7 +164,7 @@ public void BeginWrite()
       ret[objectId] = rowFound;
     }
 
-    return ret;
+    return Task.FromResult(ret);
```
AlanRynne, line 167: Is there any difference between these 2? Is this just because we're not calling `await` anywhere?

AlanRynne (commented):
I think I've got mostly questions... 🤕 🙌🏼

Core/Core/Kits/Attributes.cs
JR-Morgan: Happy to adjust if necessary, The constructor above was passing nulls through so I simply typed what I saw.
If unsure, I can simply revert this change, and keep this class null oblivious

AlanRynne: Oh you are right... 🤦🏼‍♂️ well i'm not sure we should be allowing this. I'll check the usage of this on our object side and get back to you with this

Core/Core/Kits/Units.cs
JR-Morgan: You'd prefer the switch statement over the switch expression? (I disagree, but not hard enough to object)
Or simply you dislike the fact we matching several raw string values to another raw string value... (this I agree is smelly)

JR-Morgan: As for us feet, I have a feeling the us_ft are not being used by us, so we may want to flag as [Obsolete].

Core/Core/Logging/Setup.cs
JR-Morgan: image

This was the real change, This prop was simply to not break Manager, and give us chance to gracefully transition.

Core/Core/Models/Utilities.cs
JR-Morgan: These our our functions that I recently exposed the `start` and `end` arguments.
However, I didn't realise that these are not start and end, they are startIndex and length.
Hence my rename here.  

This change hasn't shipped yet, so no damage.

Core/Core/Kits/Attributes.cs
AlanRynne: There are only 3 usages of this constructor in the entire object model, so I say we add the missing category/subcategory to those and remove (or flag as obsolete) that constructor.

I'd be happier if it were removed to be honest, even if it's a breaking change.

Core/Core/Kits/Units.cs
AlanRynne: > Or simply you dislike the fact we matching several raw string values to another raw string value

this ☝🏼 hehe switch expressions are lovable things ♥️

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
AlanRynne: OH this makes so much sense... 🤐 I'll accept it, and maybe we should keep this as a good example for the team too!

@BovineOx would love your take on this too just to be sure we're all on the same page.

Core/Core/Transports/SQLite.cs
JR-Morgan: Yea, I'm not 100% on the specific runtime differences between the two. However, there is an analyser for async methods that lack an `await` 
`CS1998: This async method lacks 'await'` (tho it appears to lack documentation)

But there's some sense here,  the `async` keyword simply means "I may be using `await`"

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
BovineOx: We might want to remove the commented out code, even if we plan to get there soon, just in case...

JR-Morgan: @gjedlicska has been pinged (as the one who implemented blobs) but we've agreed it can wait until the new year.

JR-Morgan: Happy to remove the commented out code for now, but if there aren't any objections, I'd like to keep the comment about "don't throw here (here be dragons)"

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
BovineOx: My only concern would be if there are critical failure that we would be masking by wrapping this, probably OOM springs to mind. We could create some WrapIfNotTerminal helper that would re-throw if the type of the exception was not one we wanted to mask, but I otherwise think it's fine.

BovineOx:
Added my 2 cents

Core/Core/Kits/Attributes.cs
JR-Morgan: Shall we address this in a seperate PR? I can keep this class null oblivious for now.
Then in a separate PR, we can `Obsolete(true)` the ctor,  resolve our own usages, and remove the `nullable disable` (without making any props nullable)

Core/Core/Kits/Units.cs
JR-Morgan: Perhaps Rider's AI assistant can satisfy your need for humorous dialogue adjacent to our code base.

JR-Morgan: Although, happy to revert this, don't want to step on *too many toes*

Core/Core/Kits/Attributes.cs
AlanRynne: yes lets do that to prevent polluting this PR with slightly tangential changes.

Resolving this conversation in favour of https://spockle.atlassian.net/browse/CNX-8542

Core/Core/Kits/KitDeclaration.cs
AlanRynne: Yes now that I look at this twice, this seems like at some point we the converter in the `CoreKit`.

Let's not get ahead of ourselves and remove that class, but I'm tempted!

Marking this as resolved.

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
JR-Morgan: Agreed with @BovineOx, the OOM is a great illustration.

It would be preferable to catch specific exception here, but our serializer is quite the beast, with no design towards exceptions (in that we call and pray) so I'm hesitant to improve on this much further (I'd be happier to re-write our serializer with more explicit ownership of exceptional paths)

---
The `WrapIfNotTerminal` is an interesting idea. Although, if I may iterate slightly; something like this?

```cs
if(IsTerminal(ex))
{
   throw;
}
throw new SpeckleException("...", ex);
```

However, this shouldn't be a substitute to good code design, so I'm hesitant to jump on this too.

JR-Morgan: So, what do we think, any action needed for this PR?

BovineOx: We probably pollute the stack less with the inline check but I am bit meh about having that everywhere..

It might be better to have some TryThrowWrapper, but how do we feel about that, it would be perhaps... TryThrowWrapper(<operation>, Action), something like...

Utils.TryThrowWrapper("Create Toposolid", () =>
{
  // my action
});

And internally would handle the catches nicely and throw he new SpeckleException($"Action: {action}", ex);

I think it's worth picking something, implementing it here and then advertising it as a use this going forwards for these types of cases and then we have one place to augment, tweak it.

Thoughts?

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
BovineOx: Let's do that, it's cleaner

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
AlanRynne: Resolving this as the change is incoming https://github.com/specklesystems/speckle-sharp/pull/3114

AlanRynne (approved):
Many, many conversations later, this is ready to go 🚀

PR #3110, opened 2023-12-14
fix(rhino): CNX-8388 handle ca1031 warnings in rhino connector

clairekuang:
## Description & motivation

Resolves `ca1031` exception handling warnings in the rhino connector, where easily done.
Includes minor refactor and improvement for layer handling methods in `Utils` after finding bugs during testing.
Adds comments for unresolved `ca1031` warnings, to be investigated further.

## Changes:

-Rhino connector


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Utils.cs
```diff
@@ -370,7 +450,7 @@ public static string TimeAgo(string timestamp)
     {
       timeAgo = DateTime.Now.Subtract(DateTime.Parse(timestamp));
     }
-    catch (FormatException e)
+    catch (FormatException)
     {
       Debug.WriteLine("Could not parse the string to a DateTime");
       return "";
```
AlanRynne, line 456: Interesting that the format may be wrong on the timestamp. I think we'd want to log that as I can't imagine how that would happen, but since the catch is there, it must have been for a reason

AlanRynne (commented):
Added some comments so I won't forget when we do our live review @clairekuang.

Also pinged @JR-Morgan and @BovineOx in a potentially interesting one.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -314,26 +314,29 @@ public override async Task<StreamState> ReceiveStream(StreamState state, Progres
...
-      try
+      if (Utils.GetGuidFromString(applicationId, out Guid id))
       {
-        RhinoObject obj = Doc.Objects.FindId(new Guid(applicationId));
-        if (obj != null)
+        if (Doc.Objects.FindId(id) is RhinoObject obj)
         {
           match.Add(obj);
         }
       }
```
BovineOx, line 339: Do we know why we might have wanted to catch here? Not seeing anything obvious?

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -607,8 +610,7 @@ void FlattenConvertedObject(object item)
           // handle render material
           if (render != null)
           {
-            var convertedMaterial = converter.ConvertToNative(render) as RenderMaterial; //Maybe wrap in try catch in case no conversion exists?
-            if (convertedMaterial != null)
+            if (converter.ConvertToNative(render) is RenderMaterial convertedMaterial)
```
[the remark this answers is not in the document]
BovineOx: @AlanRynne do you mean that if render is not RenderMaterial this an error as we're expecting it to actually BE one?

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
```diff
@@ -76,36 +76,30 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
-      RhinoObject obj = null;
-      try
+      if (Utils.GetGuidFromString(id, out Guid guid))
       {
-        obj = Doc.Objects.FindId(new Guid(id)); // this is a rhinoobj
-      }
-      catch
-      {
-        continue; // this was a named view!
-      }
```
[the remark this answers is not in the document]
BovineOx: Looks like never - good call taking it out @clairekuang

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
```diff
@@ -76,36 +76,30 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
-
-      if (obj != null)
-      {
-        if (deselect)
+        if (Doc.Objects.FindId(guid) is RhinoObject obj)
         {
-          obj.Select(false, true, false, true, true, true);
+          if (deselect)
+          {
+            obj.Select(false, true, false, true, true, true);
```
BovineOx, line 89: I would remove the if clause and make one call to obj.Select passing !deselect, for clarity I'd probably set a new variable with a useful name if possible to !deselect and pass that in.

BovineOx: But appreciate this is pre-existing

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
```diff
@@ -76,36 +76,30 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
+          PreviewConduit.SelectPreviewObject(id, deselect);
+          PreviewConduit.Enabled = true;
         }
-      }
-      else if (isPreview)
-      {
-        PreviewConduit.Enabled = false;
-        PreviewConduit.SelectPreviewObject(id, deselect);
-        PreviewConduit.Enabled = true;
+        // if null, this was a preivew probably
```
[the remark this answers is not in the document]
BovineOx: I would move the comment because it is too detached from the code it relates to

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -26,41 +26,44 @@ public MappingBindingsRhino()
 
   public override MappingSelectionInfo GetSelectionInfo()
   {
-    try
+    List<RhinoObject> selection = RhinoDoc.ActiveDoc.Objects.GetSelectedObjects(false, false)?.ToList();
```
BovineOx, line 29: legitimate I assume here to, that we expect each of these references to not be null?

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -26,41 +26,44 @@ public MappingBindingsRhino()
...
           result = result.Where(x => schemas.Any(y => y.Name == x.Name)).ToList();
         }
-
-        //incompatible selection
-        if (!result.Any())
-        {
-          return new MappingSelectionInfo(new List<Schema>(), selection.Count);
-        }
+        catch (ArgumentNullException) { }
       }
```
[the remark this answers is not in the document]
BovineOx: Lot's in this, can schemas be null? I am not a fan of the check !Result.Any() but reading it again it seems that if we don't have anything we overwrite the original object with the result of GetObjectSchema() which appears to always return but is itself swallowing all exceptions.

Ultimately my question is whether the moving of the try exposes us to legitimate exceptions and then secondary we are just swallowing ArgumentNullException. That feels like a smell that something can be null. It seems the fix has been to catch the exception where I would argue nothing at this point should be null - it must be x or y because comparing x.Name and y.Name should not explode any.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -26,41 +26,44 @@ public MappingBindingsRhino()
...
+      //incompatible selection
+      if (!result.Any())
+      {
+        return new MappingSelectionInfo(new List<Schema>(), selection.Count);
+      }
     }
+
+    return new MappingSelectionInfo(result, selection.Count);
+
+    // todo: previously entire block was try catched: SpeckleLog.Logger.Error(ex, "Could not get selection info: {exceptionMessage}", ex.Message);
```
BovineOx, line 65: I'm not a fan of adding more todos to the code, I would prefer we raise tickets and discus getting them onto the sprint if we have actions to follow up

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -158,6 +161,7 @@ private List<Schema> GetObjectSchemas(RhinoObject obj)
         }
       }
     }
+    // todo: this exception needs to be investigated further
```
BovineOx, line 164: It feels like we should add a ticket to look into this, then we can add comments referencing the ticket# this I would prefer - CNX-9001

BovineOx (commented):
I'll leave my review there as I think we are learning that we need to pool the exception experience and provide more guidance. We also need a pattern that involves a ticket for following up so we don't feel compelled to write more todos into the code :)

If TODOs are also not a warning...

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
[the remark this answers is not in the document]
clairekuang: It's the `new Guid(id)` that throws - I'll walk you and Ian through my handling of this since it's a pattern worth discussing imo

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
clairekuang: Same as [this comment](https://github.com/specklesystems/speckle-sharp/pull/3110#discussion_r1427382698)

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
```diff
@@ -263,17 +264,30 @@ private void EnsureVersionSettings()
...
           File.Delete(path);
         }
-        catch (Exception ex)
+        catch (IOException ioEx)
+        {
+          SpeckleLog.Logger.Error(
+            ioEx,
+            "Failed to delete Speckle toolbar .rui file with {exceptionMessage}",
+            ioEx.Message
+          );
```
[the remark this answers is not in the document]
clairekuang: In this case the rhino speckle toolbar failing to be deleted and updated doesn't prevent the speckle plugin from loading and also doesn't affect the application state. It just means the user may be left with an out-of-date toolbar, but they can still call the plugin via command, and many users probably don't use the toolbar at all.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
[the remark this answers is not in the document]
clairekuang: > since `ConvertToNative` is quite a critical operation, i feel like this is hiding it a bit. I like the change for this to be an if statement, but maybe we should do:
> 
> ```cs
> var result = converter.ConvertToNative(render);
> if(render is RenderMaterial converterMaterial){...}
> ```
> 
> I don't have very strong feelings about this though.

happy to make this change, I also don't have strong feelings about it

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
```diff
@@ -210,6 +210,7 @@ protected override LoadReturnCode OnLoad(ref string errorMessage)
     {
       Init();
     }
+    // need investigation in seq of specific excpetions thrown (FileNotFound, TypeInitialization)
     catch (Exception ex)
```
BovineOx, line 214: I think you can probably begin using the IsFatal() method Jedd has created. Though if this is at the topmost level we probably mneed a more specific way to handle IsFatal() to inform the user

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/SpeckleMappingsCommandWin.cs
```diff
@@ -28,6 +29,8 @@ protected override Result RunCommand(RhinoDoc doc, RunMode mode)
     }
     catch (Exception e)
     {
+      // needs more investigation. logging to seq for now.
```
BovineOx, line 32: Similarly we need to consider use of IsFatal() here but it might fall under the remit of that toplevel exception handling issue, I will raise one and ping the connectors team and then we can be referncing these places with the ticekt and disabling the warning for now.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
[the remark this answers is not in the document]
BovineOx: Can you book something for after the break? I think this seems better now anyways

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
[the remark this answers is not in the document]
BovineOx: Still not sure about this one - I'd like to understand the empty catch and whether we can avoud the need to for t/c

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -158,6 +161,7 @@ private List<Schema> GetObjectSchemas(RhinoObject obj)
...
     catch (Exception ex)
```
BovineOx, line 165: I think this needs to use !IsFatal(ex) at the least

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -278,8 +282,14 @@ private Schema GetExistingObjectSchema(RhinoObject obj)
 
       return JsonConvert.DeserializeObject<Schema>(viewModel, settings);
     }
-    catch
+    // todo: deserializer exception needs to be investigated later, adding logging for now
+    catch (Exception ex)
```
BovineOx, line 286: Similarly !IsFatal(ex)

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -298,20 +308,16 @@ public override void SetMappings(string schema, string viewModel)
 
   public override void ClearMappings(List<string> ids)
   {
-    foreach (var id in ids)
+    foreach (string id in ids)
```
BovineOx, line 311: So much better!

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Utils.cs
```diff
@@ -92,48 +120,62 @@ public static bool FindObjectBySelectedId(RhinoDoc doc, string id, out object ob
...
-        }
-
-        return doc.Layers.FindIndex(newIndex);
+        newLayer.ParentLayerId = parentLayer.Id;
       }
       catch (Exception e)
       {
-        return null;
+        throw new InvalidOperationException("Could not set layer parent id.", e);
       }
```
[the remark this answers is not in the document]
BovineOx: I think wrapping it is fine, I don't know as we esablished what the pattern should be.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Previews.cs
```diff
@@ -195,16 +196,15 @@ public override async Task<StreamState> PreviewReceive(StreamState state, Progre
...
+    PreviewConduit = new PreviewConduit(Preview);
+    if (PreviewConduit.Preview.Count == 0)
     {
-      PreviewConduit = new PreviewConduit(Preview);
-    }
-    catch (Exception e)
-    {
-      progress.Report.OperationErrors.Add(new Exception($"Could not create preview: {e.Message}"));
+      SpeckleLog.Logger.Information("No previewable geometry was found.");
+      progress.Report.OperationErrors.Add(new Exception($"No previewable objects found."));
```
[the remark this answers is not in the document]
clairekuang: I couldn't determine if this was throwing, but did add some needed safety checks within the `PreviewConduit` constructor. Would be good to target preview receives during testing

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -672,11 +674,7 @@ private void SetUserInfo(Base obj, ObjectAttributes attributes, ApplicationObjec
 
     // set application id
     var appId = parent != null ? parent.applicationId : obj.applicationId;
-    try
-    {
-      attributes.SetUserString(ApplicationIdKey, appId);
-    }
-    catch { }
```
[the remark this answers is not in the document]
clairekuang: Alright, couldn't identify any issues with illegal chars or duplicate keys - leaving this as is for now.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
clairekuang: Yes preexisting, but a good and easy change to make

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Selection.cs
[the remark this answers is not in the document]
clairekuang: ah I just forgot to take out this comment while refactoring

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Utils.cs
clairekuang: Actually just noticed there are no references to this method - going to remove it!

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/SpeckleMappingsCommandWin.cs
clairekuang: yup - another toplevel exception

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
clairekuang: Can confirm that `.Objects` should never return null, but it's definitely not great to use `ActiveDoc` as per rhino documentation: `WARNING!! Do not use the ActiveDoc if you don't have to. Under Mac Rhino the ActiveDoc can change while a command is running. Use the doc that is passed to you in your RunCommand function or continue to use the same doc after the first call to ActiveDoc.` 

This it out of scope for this ticket however so I'm leaving it as is unless @AlanRynne knows if the call to `ActiveDoc` would ever return null (eg when called from rhino.inside?)

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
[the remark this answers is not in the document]
clairekuang: > would this argument exception be because `result` is null? or because `x.Name` was?
> 
> If we could prevent it from happening instead i think that would be preferable (adding some null checks in key places)
> 
> I'm aware this is a mappings class so i'd tread lightly before asking for too many changes 😅

`schemas` wouldn't be null, but not sure about `x.name` - similarly I'm a bit afraid to touch anything related to mappings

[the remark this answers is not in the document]
clairekuang: I'm removing the trycatch with a null check🤞

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
clairekuang: Due to complications with Avalonia and DUI2, I'm not entirely comfortable completely removing entire blocks wrapped in a trycatch - I can rethrow with `IsFatal()` as previously suggested however since this not a topmost exception

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
BovineOx: It certainly feels like connector/converter logic requires a context object which would contain the document in context for the current calling context...  I need to play around with this in autofac so we can embody this in some DI in the future

BovineOx (commented):
Apologies I think I added more comments only now I see them as pending 😨

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
BovineOx: It probably just needs the !IsFatal() patern then in the case

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -28,14 +28,18 @@ public override MappingSelectionInfo GetSelectionInfo()
...
+      }
 
+      var result = new List<Schema>();
       foreach (var obj in selection)
       {
-        var schemas = GetObjectSchemas(obj);
+        List<Schema> schemas = GetObjectSchemas(obj);
 
-        if (!result.Any())
+        if (result.Count == 0)
```
BovineOx, line 42: I still think this if statement is unnecessary optimisation

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -58,8 +66,9 @@ public override MappingSelectionInfo GetSelectionInfo()
     }
     catch (Exception ex)
```
BovineOx, line 67: I get we are throwing, but do we want to catch something more specific or wrap if !NotFatal()

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
BovineOx: Can we remove the TODOs

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
AlanRynne: Interestingly... I see no warning for this on my IDE 🤔 

EDIT: Issue with my mac IDE... non related to this. I'm resolving this as it's been fixed

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
```diff
@@ -74,8 +83,7 @@ private List<Schema> GetObjectSchemas(RhinoObject obj)
 
     try
     {
-      var existingSchema = GetExistingObjectSchema(obj);
-      if (existingSchema != null)
+      if (GetExistingObjectSchema(obj) is Schema existingSchema)
```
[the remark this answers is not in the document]
AlanRynne: Not sure what you mean... this is a type check on the result of a method call.

Granted it's a bit odd... but it's essentially

```csharp
var schemaResult = GetExistingObjectSchema(obj);
if(schemaResult  is Schema existingSchema)
{
  ...
}
```

AlanRynne (commented):
Fixed most of the pending comments from @BovineOx except 2

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
AlanRynne: ActiveDoc **can and will be null** when running on `Rhino.Compute`, among other scenarios.

The need for a context object is real, there is contextual information that the connector needs to pass to the converter for it to do it's work properly (document, conversion caches, etc...) and some sub-conversions require different contextual information (i.e. Units.None instead of document units) to do their work properly.

ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/MappingBindingsRhino.cs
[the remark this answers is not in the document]
clairekuang: Marking this as resolved as I don't think there's a formatting issue here

AlanRynne (approved):
Approving after chat with Claire last night

If anything else pops up, we can deal with it in future tickets.

PR #3114, opened 2023-12-18
feat(Core): CNX-8570, CNX-8393 Transporter exception cleanup and IsFatal(this Exception) function

JR-Morgan:
Lets merge https://github.com/specklesystems/speckle-sharp/pull/3109 first as this one is built off the last


Core/Core/Credentials/StreamWrapper.cs
```diff
@@ -319,13 +321,14 @@ public async Task<Account> GetAccount()
...
-      catch (Exception e)
+      catch (Exception ex) when (!ex.IsFatal())
       {
-        err = e;
+        err.Add(new SpeckleException($"Account {acc} failed to auth stream wrapper", ex));
       }
     }
 
-    throw err;
+    AggregateException inner = new(null, err);
```
JR-Morgan, line 330: I noticed here, we would only include the exception from the last account that failed. I decided it might be useful to have the exceptions from all of them.

Core/Core/Helpers/Http.cs
```diff
@@ -128,23 +131,24 @@ public static async Task<bool> Ping(string hostnameOrAddress)
   }
 
   /// <summary>
-  /// Pings and tries gettign data from a specific address to verify it's online. Retries 3 times.
+  ///
   /// </summary>
-  /// <param name="address">The address to use</param>
-  /// <returns>True if the the status code is successful, false otherwise.</returns>
-  public static async Task<bool> HttpPing(string address)
```
JR-Morgan, line 135: Quick refactor of this function to throw instead of return true/false, just so we have keep the stack trace + exception message

Core/Core/Models/Base.cs
```diff
@@ -286,4 +274,35 @@ public Base ShallowCopy()
 
     return myDuplicate;
   }
+
+  #region Obsolete
+  /// <inheritdoc cref="GetId(bool)"/>
+  [Obsolete("Serializer v1 is now deprecated, use other overload(s)")]
+  public string GetId(SerializerVersion serializerVersion)
```
JR-Morgan, line 281: Oops. I accidently included some serialzierv1 deprecation.
I can unstage this if preferred.

Core/Core/Serialisation/DeserializationWorkerThreads.cs
```diff
@@ -57,6 +57,11 @@ protected override void ThreadMain()
       catch (Exception ex)
       {
         tcs.SetException(ex);
+
+        if (ex.IsFatal())
+        {
+          throw;
```
JR-Morgan, line 63: Not sure what happens (in practice) when we throw here. I have a feeling it becomes an unobserved task exception. Not ideal, so I thought it would be useful to make sure the task completion source `tcs` stays in a faulted state

Core/Core/Transports/TransportExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+  /// <summary>
+  /// <inheritdoc cref="ITransport.SaveObject(string, string)"/>
+  /// Retrieves the object from the provided transport
+  /// </summary>
+  /// <param name="destination">The destination</param>
+  /// <param name="id"><inheritdoc cref="ITransport.SaveObject(string, string)"/></param>
+  /// <param name="source">The transport from where to retrieve it.</param>
+  /// <exception cref="TransportException">Failed to save object</exception>
+  /// <exception cref="System.OperationCanceledException"><see cref="System.Threading.CancellationToken"/> requested cancel</exception>
+  public static void SaveObject(this ITransport destination, string id, ITransport source)
```
JR-Morgan, line 14: All our transports implemented this function the same. So I've made it a helper function, rather than included in the `ITransport`. This should be a non-source breaking change (since its in the same namespace)

AlanRynne:
Merged! I think we need a rebase of this now :)

JR-Morgan:
Back merge from dev complete. Ready for review

Core/Core/Credentials/StreamWrapper.cs
AlanRynne: Oh this is great

Core/Core/Helpers/Http.cs
```diff
@@ -52,23 +52,26 @@ public static AsyncRetryPolicy<HttpResponseMessage> HttpAsyncPolicy(IEnumerable<
...
+      //Perform a quick ping test e.g. to cloudflaire dns, as is quicker than pinging server
+      if (await Ping("1.1.1.1").ConfigureAwait(false))
+      {
+        return true;
+      }
+
+      defaultServer = AccountManager.GetDefaultServerUrl();
+      Uri serverUrl = new(defaultServer);
+      await HttpPing(serverUrl).ConfigureAwait(false);
       return true;
```
AlanRynne, line 67: Changes are good, but this function is doing more than it should (possibly another ticket?).
`UserHasInternet` as it stands requires your default account server to be online too, which is not what the function says it will do. Checking if the "computer user" has internet is not the same as checking if the "speckle user" is valid against the server, specially since the default server url can be other than our own.

This will result in "You are not connected to the internet" problems that may not be caused by that.

Core/Core/Helpers/Http.cs
AlanRynne: This is was always intended as internal so i'm fine with this.

Core/Core/Logging/ExceptionHelpers.cs
```diff
@@ -0,0 +1,47 @@
...
+      or ThreadAbortException
+      or InvalidProgramException
+      or AccessViolationException
+      or AppDomainUnloadedException
+      or BadImageFormatException
+        => true,
+      _ => false,
+    };
+  }
+}
```
AlanRynne, line 47: I'm adding this to the guidance already 👍🏼

Core/Core/Models/Base.cs
AlanRynne: Hmmm... if it's not a huge ask lets roll this back and open an extra PR for this 🙏🏼

Core/Core/Serialisation/DeserializationWorkerThreads.cs
AlanRynne: I'm going to have to defer this to @BovineOx 😅

Core/Core/Transports/ITransport.cs
```diff
@@ -66,16 +66,6 @@ public interface ITransport
...
 
-  /// <summary>
-  /// <inheritdoc cref="SaveObject(string, string)"/>
-  /// Retrieving its serialised version from the provided transport.
-  /// </summary>
-  /// <param name="id"><inheritdoc cref="SaveObject(string, string)"/></param>
-  /// <param name="sourceTransport">The transport from where to retrieve it.</param>
-  /// <exception cref="TransportException">Failed to save object</exception>
-  /// <exception cref="OperationCanceledException"><see cref="CancellationToken"/> requested cancel</exception>
-  public void SaveObject(string id, ITransport sourceTransport);
```
AlanRynne, line 77: This is the one that was moved to a shared implementation right?

Not sure how many people would have their own transport but technically this is a breaking change if your transport already has `SaveObject` implemented right? But also we're saying a transport "doesn't save objects" when that's it's whole point of existing.

Since the extension lives in the same namespace as the interface, it must cause a conflict. I'm not sure we should remove this from here, though the shared implementation should stay.

It would be a bit more verbose to just keep this and call the shared implementation on every transport class.

Core/Core/Transports/TransportExtensions.cs
AlanRynne: Oh, you learn something new every day, i never tried to extend an interface 😅

Core/Tests/Kits/KitManagerTests.cs
AlanRynne, line 1: Interesting... this file appears as unchanged

AlanRynne (changes requested):
Mostly great! 🙌🏼 I have questions, pinged @BovineOx to weigh in on some, and some requests/opinionated thoughts

Core/Core/Transports/TransportExtensions.cs
AlanRynne: Forgot to mention here, i added a comment about this in the deleted interface method before realising this was here!

Core/Core/Transports/ITransport.cs
JR-Morgan: There shouldn't be any conflicts. If you call `SaveObject` on an `ITransport`, it'll use the extension method. If you call `SaveObject` on your concrete type, then it will use the member function.
Case in point, I didn't touch `DiskTransport`, and the tests still pass! (side note: should I update `DiskTransport` too? )

The only potential issues would be if somehow you're relying on some custom behaviour in `SaveObject(string, ITransport)`

If you think that's possible, then I can revert this change.

> But also we're saying a transport "doesn't save objects" when that's it's whole point of existing.

Not sure I quite follow here... maybe I just revert and we chat in the new year?

Core/Core/Models/Base.cs
JR-Morgan: reverted

Core/Core/Helpers/Http.cs
JR-Morgan: 100%, I was holding back on re-writing this code. We're relying on either `1.1.1.1` ping succeeding (sometimes blocked in corporate firewall/DNS) or your default server being online and your account being valid for it.

Core/Core/Credentials/AccountManager.cs
```diff
@@ -125,9 +125,9 @@ private static async Task<ActiveUserServerInfoResponse> GetUserServerInfo(string
 
       return res.Data;
     }
-    catch (Exception e)
+    catch (Exception ex) when (!ex.IsFatal())
     {
-      throw new SpeckleException(e.Message, e);
+      throw new SpeckleException(ex.Message, ex);
```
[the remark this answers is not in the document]
JR-Morgan: Done!

Core/Core/Transports/ITransport.cs
AlanRynne: I meant that the interface is in essence our contract to other developers detailing what a class "should do", even if all our Transport classes do the same thing, I don't think it's safe to say that "every transport" will do the same?

We have no knowledge of other transports existing but that doesn't mean they don't 😅

I would revert the deletion in the interface and just do

`public void SaveObject(...) => TransportExtensions.SaveObject(...)`

Obviously `TransortExtensions` naming would no longer make sense and may need revisiting... maybe `TransportUtilities` or `TransportBase`? 🤔 Naming is hard...

This way `TransportExtensions.SaveObject` doesn't need to be an extension method either, and we don't break the existing contract (even though I'm convinced after your comment it is not really breaking)

Core/Core/Helpers/Http.cs
AlanRynne: Resolving, added ticket to re-evaluate this in the backlog
https://spockle.atlassian.net/browse/CNX-8643

Core/Core/Api/GraphQL/Client.cs
BovineOx, line 1: Why are we choosing to log for some exceptions but not all?

Core/Core/Api/GraphQL/Client.cs
```diff
@@ -188,7 +191,7 @@ public async Task<T> ExecuteGraphQLRequest<T>(GraphQLRequest request, Cancellati
     }
     // we log and wrap anything that is not a graphql exception.
     // this makes sure, that any graphql operation only throws SpeckleGraphQLExceptions
-    catch (Exception ex)
+    catch (Exception ex) when (!ex.IsFatal())
```
BovineOx, line 194: Ahh I see this is the only change, I was curious about why we log for SpeckleGraphQLException<T> but not the others...  but probably beyoond the scope

Core/Core/Serialisation/BaseObjectSerializerV2.cs
```diff
@@ -65,7 +66,19 @@ public string Serialize(Base baseObj)
     {
       _stopwatch.Start();
       _isBusy = true;
-      IDictionary<string, object?> converted = PreserializeBase(baseObj, true)!;
+      IDictionary<string, object?> converted;
```
BovineOx, line 69: Would SDK consumers need to catch ArgumentException? This is one of those times when I wonder if a SpeckleArgumentException might be more easily handled?

Core/Core/Serialisation/DeserializationWorkerThreads.cs
BovineOx: This one will need some discussion. I suggest we add anew ticket, technical debt and discuss after the Christmas break

Core/Core/Transports/ServerUtils/ParallelServerAPI.cs
```diff
@@ -236,6 +237,11 @@ protected override void ThreadMain()
       catch (Exception ex)
       {
         tcs.SetException(ex);
+
```
BovineOx, line 240: Similar category

Core/Core/Transports/ServerV2.cs
```diff
@@ -415,7 +398,13 @@ private async void SendingThreadMain()
           _sendBuffer.Clear();
           _exception = ex;
         }
-        return;
+
```
BovineOx, line 401: Okay but this whole class has some issue as you are probably aware. locking will hold the thread and not await, thread.sleep similarly, should be Task.Delay() which is awaitable. I presume there is some queue here and it may be we can engineer this a bit better.

BovineOx:
@JR-Morgan I think it's broadly fine - a couple of issues we should consider but if you can raise a ticket and reference that in those places in the code and then it can go.

Core/Core/Serialisation/BaseObjectSerializerV2.cs
JR-Morgan: Partially agree, I think I should switch this to a `SpeckleException`. `ArgumentException` is not an appropriate exception here for the same reason why I don't think `SpeckleArgumentException` is a good idea.

---

`ArgumentException` implies an argument value is not valid for the method call. These validation rules should be consistent and "Functionally pure", i.e. the same input always gives the same result (exception throw or not) regardless of external (runtime) state.

The caller should be-able to easily avoid runtime `ArgumentException` by ensuring the arguments they pass are valid. 
Either by using pre-call checks, or by handling the `ArgumentException` separately.

In this case, we are throwing `ArgumentException` reactively, and its not clear how the caller could avoid this exception if they wanted to.

---

In general I think `ArgumentException` is perfectly valid for us to throw (as long as we don't misuse it). It fits with .NET standards and Microsoft guidance.
If we use `SpeckleArgumentException`, what would this inherit from?
 - If its `ArgumentException`, then what advantage does this give, will we ever want to handle `SpeckleArgumentException` differently from `ArgumentException` I don't think so, and if not, then what value does it add?
- If its `SpeckleException`, we have the opposite question, won't we always want to handle them differently? likely, a developer wants to know if they are responsible for passing around invalid objects or breaking call contracts.  Having `SpeckleArgumentException` inherit from `SpeckleException` introduces a risk that this goes unnoticed, and accidently handled in the same way other general runtime failures (where the caller likely doesn't have control).
We'd be mixing exceptions that are always the caller's fault, with exceptions that are the fault of some external state/service.
- What about `ArgumentNullException` do we replicate this also, leading to same questions as the above (and potential confusion on the callers side)

Core/Core/Transports/ITransport.cs
JR-Morgan: I've reverted the `ITransport` change, and will include it in a separate PR later. We can discuss that more when you're back!

Core/Core/Transports/ServerV2.cs
JR-Morgan: Well aware of the async related issues with our transports.
I'm not a fan of the implementation, but gonna require some more involved refactors to address.

(see https://spockle.atlassian.net/browse/CNX-8613)

For now, resolving this item, it's not something we can solve right away.

Core/Core/Api/GraphQL/Client.cs
```diff
@@ -138,6 +138,9 @@ internal async Task<T> ExecuteWithResiliencePolicies<T>(Func<Task<T>> func)
     return await graphqlRetry.ExecuteAsync(func).ConfigureAwait(false);
   }
 
+  /// <exception cref="SpeckleGraphQLForbiddenException{T}">"FORBIDDEN" on "UNAUTHORIZED" response from server</exception>
```
BovineOx, line 141: Would be inclined to add the rest of the method documentation in

Core/Core/Helpers/Http.cs
```diff
@@ -52,23 +52,26 @@ public static AsyncRetryPolicy<HttpResponseMessage> HttpAsyncPolicy(IEnumerable<
...
     {
+      //Perform a quick ping test e.g. to cloudflaire dns, as is quicker than pinging server
+      if (await Ping("1.1.1.1").ConfigureAwait(false))
+      {
+        return true;
+      }
+
+      defaultServer = AccountManager.GetDefaultServerUrl();
+      Uri serverUrl = new(defaultServer);
+      await HttpPing(serverUrl).ConfigureAwait(false);
```
BovineOx, line 66: This throws if inaccessible I guess?

Core/Core/Helpers/Http.cs
```diff
@@ -128,23 +131,24 @@ public static async Task<bool> Ping(string hostnameOrAddress)
   }
 
   /// <summary>
-  /// Pings and tries gettign data from a specific address to verify it's online. Retries 3 times.
+  ///
```
BovineOx, line 134: We lost the summary?

Core/Core/Helpers/Http.cs
```diff
@@ -128,23 +131,24 @@ public static async Task<bool> Ping(string hostnameOrAddress)
...
+      SpeckleLog.Logger.Warning("Successfully pinged {uri}", uri);
+      return response;
     }
-    catch (Exception ex)
+    catch (HttpRequestException ex)
     {
-      SpeckleLog.Logger.Warning(ex, "Exception while pinging: {message}", ex.Message);
-      return false;
+      SpeckleLog.Logger.Warning(ex, "Ping to {uri} was unsuccessful: {message}", uri, ex.Message);
+      throw new HttpRequestException($"Ping to {uri} was unsuccessful", ex);
```
BovineOx, line 151: Do we need to weap the HttpRequestException with this to provide the context?

Core/Core/Logging/ExceptionHelpers.cs
BovineOx, line 1: Super :D

Core/Core/Models/Base.cs
```diff
@@ -278,7 +280,7 @@ public Base ShallowCopy()
       {
         myDuplicate[kvp.Key] = kvp.Value;
       }
-      catch (Exception)
+      catch (Exception ex) when (!ex.IsFatal())
```
BovineOx, line 283: I guess there's no opportunity for logging anything here, eh?

Core/Core/Models/CommitObjectBuilder.cs
```diff
@@ -105,7 +105,7 @@ protected void ApplyRelationships(IEnumerable<Base> toAdd, Base rootCommitObject
       {
         ApplyRelationship(c, rootCommitObject);
       }
-      catch (Exception ex)
+      catch (Exception ex) when (!ex.IsFatal())
```
BovineOx, line 108: Feel free to resolve but I gotta wonder when I see these sorts of things, but what if this gets caught, to the caller ApplyRelationships() worked :O

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
```diff
@@ -131,7 +131,7 @@ private List<(string, int)> GetClosures(string rootObjectJson)
       }
       return closureList;
     }
-    catch (Exception)
+    catch (Exception ex) when (!ex.IsFatal())
```
BovineOx, line 134: Same, I know there't not part of this work..  GetClosures() exploded and we go, ahh clearly there were none :D

Core/Core/Serialisation/BaseObjectSerializerV2.cs
BovineOx: Well we're stating we can throw it, I do feel it's a pain to have to catch multiple types when calling into an API. But happy for you to resolve it.

Core/Core/Serialisation/DeserializationWorkerThreads.cs
```diff
@@ -57,6 +56,11 @@ protected override void ThreadMain()
       catch (Exception ex)
       {
         tcs.SetException(ex);
+
```
BovineOx, line 59: You might want to share this pattern more widely as I think it came up yesterday and this look like a better solution.

DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
```diff
@@ -63,44 +63,56 @@ internal void ReloadUsers()
...
+          DropDownOpen = true;
+        }
+      }
+      else
+      {
+        userSearchDebouncer();
       }
     }
-    else
+    catch (Exception ex) when (!ex.IsFatal())
```
BovineOx, line 108: If we throw Fatal error up, is that going to cause the host app issues or will DUI handle it gracefully?

Core/Core/Helpers/Http.cs
JR-Morgan: Since this is method effectively a helper wrapper around `httpClient.GetAsync`, I wanted to keep it to using the `HttpResponseMessage` return OR  `HttpRequestException` throw as this fits with other HttpClient functions.

This method is internal, so we aren't too concerned with public API. But easy enough change if you'd prefer me to use a different type.

I think it could be quite dangerous to start using `SpeckleException` everywhere inside our private/internal implementation (e.g. if a `SpeckleException` thrown from private/internal implementation bubbles up too high by mistake (to public callers) we'd never really know, and this could hide bugs. Using lower level types might allow us to catch this category of mistake)

Core/Core/Helpers/Http.cs
JR-Morgan: I'll add a summary (tho, I didn't remove this one by accidently)

The comment about retries is (as far as I can tell) misleading/inaccurate.
And this is an internal function, so the comment is just for us.

But the importance is that this is a `HTTP Get` request (not a ICMP/ping). Worthy of a comment

DesktopUI2/DesktopUI2/ViewModels/CollaboratorsViewModel.cs
JR-Morgan: I will remove this

Core/Core/Models/CommitObjectBuilder.cs
JR-Morgan: I think this was to avoid some edge cases where we've setup relationships for objects that are invalid, and we wanted to ignore those objects.
But perhaps there's a cleaner way we can check & throw for this.

Core/Core/Serialisation/DeserializationWorkerThreads.cs
JR-Morgan: Not convinced this pattern is needed outside of manually manipulating threads (which we should avoid doing, even inside core imo)

But if you have an example of where else this could be applied, happy to re-consider. You mentioned @clairekuang came a across some place that sounded similar?

Core/Core/Serialisation/BaseObjectSerializerV2.cs
JR-Morgan: We seem on the same page now, I've updated this to a `SpeckleException`

Core/Core/Models/Base.cs
JR-Morgan: I'll add a warning log. But I have a feeling we can avoid this entirely / make the exception type more specific

Core/Core/Serialisation/BaseObjectDeserializerV2.cs
JR-Morgan: Agreed, but needs some future clean-up/refactor for us to properly fix.

BovineOx (approved):

PR #3115, opened 2023-12-19
Jrm/tests/warnings

JR-Morgan:
1. Resolved easy to fix warnings in core tests
1. Move Unit tests into folder structure matching what we did for automate

new structure
```
├───Core
├───Examples
├───Tests
│   ├───Speckle.Core.Tests.Integration
│   ├───Speckle.Core.Tests.Performance
│   └───Speckle.Core.Tests.Unit
└───Transports
```


Core/Tests/Directory.Build.props
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
JR-Morgan, line 1: @AlanRynne for the test projects (Unit + integration + performance)

I think we'd want something like this, with far fewer of the analysers running. 
My thought is, we often are unit testing things that (like how our serializer deals with obsolete functions, nested classes, classes etc.) all things that trigger a warning.
Let me know what you think, this is just a first pass

Core/Core/Core.csproj
JR-Morgan, line 1: In Core, Why do we have
```xml
<RuntimeIdentifiers>win-x64;osx-x64;osx-arm64</RuntimeIdentifiers>
```
We already run on other platforms...

Core/Transports/MongoDBTransport/MongoDB.cs
```diff
@@ -21,6 +21,7 @@ public enum Field
 
 // Question: the benefit of noSQL is the use of unstructured collections of variable documents.
 // Explore storing partially serialized Speckle objects with dynamically generated fields instead of just a content string?
+[Obsolete("This Transport is no longer maintained or routinely tested, use with caution")]
```
JR-Morgan, line 24: Should discuss this as a team before merge, and check how many users are using this (if any).
We haven't tested this in over a year. I think we should either re-vitalise it, or deprecate it. 

I may compile a list of some other parts of core we could consider deprecating (serializerV1) but we'll need to make sure all the team has had the chance to voice their thoughts.

Core/Tests/Speckle.Core.Tests.Integration/.vscode/launch.json
AlanRynne, line 1: do we need this in this folder? I would argue we should just use All.sln and it's filters, and have .vscode at the folder root?

Happy to be talked out of this opinion though 🤣

Core/Tests/Speckle.Core.Tests.Integration/.vscode/tasks.json
AlanRynne, line 1: Same as above :)

AlanRynne (changes requested):
Flagging this as "changes requested" but it's mostly great!

Only thing is the test project `Directory.Build.props` that we may want to re-use elsewhere and may be better directly imported instead.

Core/Tests/Directory.Build.props
JR-Morgan: OK cool. I did have this though that we may want to reuse. Glad to hear its easy to pull off!

Core/Tests/Speckle.Core.Tests.Integration/.vscode/launch.json
JR-Morgan: I'll have to double check what's happen here. this may not have been an intentional change

Core/Tests/Speckle.Core.Tests.Integration/Speckle.Core.Tests.Integration.csproj
```diff
@@ -4,13 +4,14 @@
     <TargetFramework>net7.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-
     <IsPackable>false</IsPackable>
-    <NoWarn>$(NoWarn);CA2007</NoWarn>
+    <RootNamespace>Speckle.Core.Tests.Integration</RootNamespace>
```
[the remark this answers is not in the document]
JR-Morgan: Good point. I'll double check and remove if possible

Core/Tests/Speckle.Core.Tests.Integration/.vscode/launch.json
JR-Morgan: Looks like these were existing files in the existing folder structure. 
Lets have a chat about this, I'm not immediately sure what's needed here

JR-Morgan: https://spockle.atlassian.net/browse/CNX-8660

Core/Tests/Directory.Build.props
JR-Morgan: Implemented a solution with alan that uses the `IsTestProject` prop rather than unique props files.

AlanRynne (approved):
All new changes look good 🚀

PR #3116, opened 2023-12-19
feat(civil3d): CNX-8013 add alignment properties design speed

clairekuang:
## Description & motivation

Adjusts the previous pr's implementation of design speeds to include design speeds that may not be on an alignment's existing stations. 
This implementation stores each design speed as a `Base` with 4 properties: the speed `number`, `station`, `value`, and `comment` (if exists). The list of speeds is stored as a detached `List<Base>` on the alignment.

[https://speckle.xyz/streams/b53a53697a/commits/89b9c3e6f5](https://speckle.xyz/streams/b53a53697a/commits/89b9c3e6f5)

## Changes:

- Civil3D converter

### DIFF
--- a/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
+++ b/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
@@ -46,7 +46,7 @@ public partial class ConverterAutocadCivil
   }
 
   // stations
-  public Station StationToSpeckle(CivilDB.Station station, double? designSpeed)
+  public Station StationToSpeckle(CivilDB.Station station)
   {
     var speckleStation = new Station
     {
@@ -55,11 +55,6 @@ public partial class ConverterAutocadCivil
       number = station.RawStation,
       units = ModelUnits
     };
-    
-    if (designSpeed is not null)
-    {
-      speckleStation["designSpeed"] = designSpeed.Value;
-    }
 
     return speckleStation;
   }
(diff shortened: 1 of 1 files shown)


Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -189,6 +185,28 @@ public CivilAlignment AlignmentToSpeckle(CivilDB.Alignment alignment)
...
     return speckleAlignment;
   }
+
+  private List<Base> DesignSpeedsToSpeckle(DesignSpeedCollection designSpeeds)
+  {
+    List<Base> speckleDesignSpeeds = new();
+
+    foreach (DesignSpeed designSpeed in designSpeeds)
+    {
+      Base speckleDesignSpeed = new();
```
AlanRynne, line 195: Only consideration here would be we may want to just have an object type for this instead? Once we decide it's `Base`, undoing this will be a PITA 😅

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -331,13 +349,13 @@ public ApplicationObject AlignmentToNative(Alignment alignment)
...
 
     // set design speeds if any
-    if (civilAlignment["@stations"] is List<object> stations)
+    if (civilAlignment["@designSpeeds"] is List<object> speeds)
     {
-      foreach (Station station in stations.Cast<Station>())
+      foreach (object speed in speeds)
       {
-        if (station["designSpeed"] is double designSpeed)
+        if (speed is Base speedBase && speedBase["station"] is double station && speedBase["value"] is double value)
```
AlanRynne, line 356: Having a typed class will allow to do a typecheck here instead of 3

AlanRynne (commented):
I'm not sure if you already discussed this before the implementation and I'm speaking out of turn... so i'll leave this as a comment 😄

AlanRynne (approved):
Approving after quick chat with Claire

PR #3118, opened 2023-12-20
fix:(tekla): CNX-8391 Handle warnings CA1031

jsdbroughton:
## THIS PR SHOULD BE MERGED AFTER ExceptionHelpers has merged.

Changes to:

### Connector
MainForm
StreamStateManager

### Converter
ConverterTeklaStructureUtils
ConvertMeshToNative

In general I think from my non-codebase owner eye, there are many areas where more exceptions may be raised and are not handled. This PR addresses the generic cases where they were.


ConnectorTeklaStructures/ConnectorTeklaStructuresShared/StreamStateManager/StreamStateManager.cs
```diff
@@ -41,31 +43,66 @@ public static void WriteStreamStateList(Model model, List<StreamState> streamSta
...
       GetOrCreateSpeckleFilePath(model);
     }
 
-    FileStream fileStream = new(_speckleFilePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite);
+    if (_speckleFilePath == null || !File.Exists(_speckleFilePath))
+    {
+      return;
+    }
+
+    FileStream fileStream = null;
```
BovineOx, line 92: so much better!

Objects/Converters/ConverterTeklaStructures/ConverterTeklaStructuresShared/PartialClasses/ConvertDirectShapeMesh.cs
```diff
@@ -31,13 +47,24 @@ public void MeshToNative(Base @object, List<GE.Mesh> displayValues)
...
       }
-      catch (Exception ex) { }
+      catch (InvalidOperationException ex)
+      {
+        throw new ConversionException("The converted Mesh could not be inserted.", ex);
+      }
+      catch (ArgumentException ex)
+      {
+        // if the exception message contains "BRep already exists", swallow the error, else rethrow
+        if (!ex.Message.Contains("The BRep geometry already exists"))
```
BovineOx, line 63: This is a bit brittle, is there a better way to identify this? Can we avoid trying the Insert() if we know it already exists?

BovineOx (commented):
A few things to consider, few questions

Objects/Converters/ConverterTeklaStructures/ConverterTeklaStructuresShared/PartialClasses/ConvertDirectShapeMesh.cs
jsdbroughton: It could be I was being careful to maintain the current behaviour/swallow pattern but this is a good point

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainForm.cs
```diff
@@ -24,41 +24,89 @@ public partial class MainForm : PluginFormBase
...
-        var viewModel = new MainViewModel(Bindings);
-        MainWindow = new DesktopUI2.Views.MainWindow { DataContext = viewModel };
-
-        Bindings.OpenTeklaStructures();
+        SubscribeToEvents();
+        BuildAndShowMainWindow();
+        SetTeklaAsOwner();
       }
-      catch (Exception ex)
+      catch (Exception ex) when (!ex.IsFatal())
```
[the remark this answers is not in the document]
jsdbroughton: This is the example I was referring to - most of the DUI c# connectors have approximately the same general Exception catch and some previously swallowed. Probably a candidate for coalescing.

Objects/Converters/ConverterTeklaStructures/ConverterTeklaStructuresShared/PartialClasses/ConvertDirectShapeMesh.cs
jsdbroughton: The answer is no there are some internal methods for ShapeExists and no documentation of generating that `geohash` string 
image

Their code examples in documentation suggest the try it and see method for `Insert()`

I am 100% open to throwing or logging and moving on for all exceptions - there are several that could be thrown:
image

I was keeping that single exception allowable as it seemed to represent the previous comment of why the error should be swallowed.

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/StreamStateManager/StreamStateManager.cs
```diff
@@ -41,31 +43,66 @@ public static void WriteStreamStateList(Model model, List<StreamState> streamSta
       GetOrCreateSpeckleFilePath(model);
     }
 
+    if (_speckleFilePath == null)
+      return;
```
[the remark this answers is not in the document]
jsdbroughton: done. damn i missed that one

Objects/Converters/ConverterTeklaStructures/ConverterTeklaStructuresShared/PartialClasses/ConvertDirectShapeMesh.cs
BovineOx: This is why I am keen we wrap our exceptions at some point when uses call our SDK code so that users can at least say, okay this was a SpeckleException, thing I tried didn't work.

Actually here I would catch ArgumentException, InvalidOperationException and ArgumentOutOfRangeException if that is the code we are calling into. We can then wrap it with our own context if throwing upwards - cannot see from this view.

jsdbroughton: I started with that - but collapsed a little as `ArgumentOutOfRange` is an `ArgumentException` all we'd be doing is making a distinction where Tekla SDK could not. The pasted switch-case code is Teklas, not ours.

I am happy to do whatever we collectively want, I picked this Tekla ticket up as it has no owner :D

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/StreamStateManager/StreamStateManager.cs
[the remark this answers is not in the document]
BovineOx: Looking at this again, it would be clearer if the 2nd check occurs within the first if block. It initially looked like a mistake in the code, but I see GetOrCreateSpeckleFilePath() tries to create the paht also

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/StreamStateManager/StreamStateManager.cs
```diff
@@ -1,142 +1,210 @@
...
       GetOrCreateSpeckleFilePath(model);
     }
 
-    FileStream fileStream = new(_speckleFilePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite);
+    if (s_speckleFilePath == null)
+    {
+      return;
+    }
+
+    FileStream fileStream = new(s_speckleFilePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite);
```
BovineOx, line 54: This can throw, probably ought to be in the try/catch

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/StreamStateManager/StreamStateManager.cs
```diff
@@ -1,142 +1,210 @@
...
-        streamWriter.Flush();
+        return;
       }
+
+      streamWriter.Write(streamStateString);
+      streamWriter.Flush();
+    }
+    catch (Exception ex) when (!ex.IsFatal())
+    {
+      SpeckleLog.Logger.Error("Failed to write stream state list to file");
```
BovineOx, line 76: Does the caller get any feedback?

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainForm.cs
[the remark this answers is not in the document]
BovineOx: Yes we have a ticket to handle the top-level catch block in the backlog, but for now it's probably worth catching Exception here and disabling the warning, unless you're saying something that calls this WILL catch(Exception)

Objects/Converters/ConverterTeklaStructures/ConverterTeklaStructuresShared/PartialClasses/ConvertDirectShapeMesh.cs
BovineOx: All good then

Objects/Converters/ConverterTeklaStructures/ConverterTeklaStructuresShared/PartialClasses/ConvertDirectShapeMesh.cs
```diff
@@ -48,23 +74,33 @@ public void MeshToNative(Base @object, List<GE.Mesh> displayValues)
           shapeName = matchingShape.Name;
           result = true;
         }
+        else
```
BovineOx, line 77: I'd rather add a ticket labelled with Tech debt and reference that here than add more TODOs.

BovineOx:
Just a few things to consider, but minor, once you've done so happy for you to resolve the comments - ping me if it needs an approval still

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/StreamStateManager/StreamStateManager.cs
jsdbroughton: Actually, DUI catches errors when calling this, so can rethrow instead.

ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainForm.cs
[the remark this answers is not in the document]
jsdbroughton: No catch exists above this. A failure to create the plugin window shouldn't disable the host application.

AlanRynne (approved):
Approving since all conversations are resolved and CI is ✅

PR #3119, opened 2023-12-21
refactor(autocadcivil): CNX-8389 fixes CA1031 warnings in autocad and civil3d projects

clairekuang:
## Description & motivation

Fixes CA1031 in Autocad and Civil3D.

## Changes:

- Autocad and Civil3d Connectors and Converters

## To-do before merge:

- [x] Check for Speckle Logging initialization in Connector
- [x] Replace `!IsFatal() TODOS`


ConnectorAutocadCivil/ConnectorAutocadCivil/Entry/App.cs
```diff
@@ -243,21 +242,54 @@ private RibbonToolTip CreateToolTip(string title, string content)
...
       string resource = GetType().Assembly
         .GetManifestResourceNames()
         .Where(o => o.EndsWith(sourceName))
         .FirstOrDefault();
-      Assembly assembly = Assembly.GetExecutingAssembly();
-      Stream stream = assembly.GetManifestResourceStream(resource);
-      PngBitmapDecoder decoder = new(stream, BitmapCreateOptions.PreservePixelFormat, BitmapCacheOption.Default);
-      ImageSource source = decoder.Frames[0];
-      return source;
+      if (string.IsNullOrEmpty(resource))
```
BovineOx, line 259: Can we have a new line between the end of the LINQ and the if please?

ConnectorAutocadCivil/ConnectorAutocadCivil/Entry/SpeckleAutocadCommand.cs
```diff
@@ -114,7 +115,10 @@ public static void SpeckleCommunity()
         true
       );
     }
-    catch { }
+    catch (System.Exception ex)
```
BovineOx, line 118: Are these all top-level exceptions? If we didn't catch this, would the host app have to handle it (and possibly crash?) if so, all good - we have a ticket for better handling for TL-exceptions

ConnectorAutocadCivil/ConnectorAutocadCivil/Storage/SpeckleStreamManager.cs
```diff
@@ -60,17 +62,18 @@ public static List<StreamState> ReadState(Document doc)
...
-      try
+      if (tr.GetObject(id, OpenMode.ForRead) is Xrecord record)
       {
-        //Try to decode here because there is old data
-        value = Base64Decode(value);
+        string value = GetXrecordData(record);
+        if (!string.IsNullOrEmpty(value))
+        {
+          //Try to decode here because there is old data
+          if (Base64Decode(value, out value))
```
BovineOx, line 71: Should it be valid to fail the base64 decode? I appreciate we swallowed it before, but I imagine if we got this far we expected it to work? Should we at least log something?

ConnectorAutocadCivil/ConnectorAutocadCivil/Storage/SpeckleStreamManager.cs
```diff
@@ -80,7 +83,6 @@ public static List<StreamState> ReadState(Document doc)
   /// Writes the stream states to the current document.
   /// </summary>
   /// <param name="doc"></param>
```
BovineOx, line 85: We should udate the parma list to include steamStates

ConnectorAutocadCivil/ConnectorAutocadCivil/Storage/SpeckleStreamManager.cs
```diff
@@ -148,10 +150,45 @@ private static string Base64Encode(string plainText)
...
   }
 
-  private static string Base64Decode(string base64EncodedData)
+  /// <summary>
+  /// Decodes base64 encoded string.
+  /// </summary>
+  /// <param name="base64EncodedData"></param>
+  /// <param name="decodedString">The decoded string</param>
+  /// <returns>True on success, false on failure</returns>
+  private static bool Base64Decode(string base64EncodedData, out string decodedString)
```
BovineOx, line 159: We should name this TryBase64Decode() I feel

ConnectorAutocadCivil/ConnectorAutocadCivil/Storage/SpeckleStreamManager.cs
```diff
@@ -148,10 +150,45 @@ private static string Base64Encode(string plainText)
...
   {
-    var base64EncodedBytes = Convert.FromBase64String(base64EncodedData);
-    return Encoding.UTF8.GetString(base64EncodedBytes);
+    decodedString = null;
+    if (string.IsNullOrEmpty(base64EncodedData))
```
BovineOx, line 162: perhaps IsNullOrWhiteSpace()

ConnectorAutocadCivil/ConnectorAutocadCivil/Storage/SpeckleStreamManager.cs
```diff
@@ -148,10 +150,45 @@ private static string Base64Encode(string plainText)
...
+    {
```
BovineOx, line 163: It's wordy but it's way better :D

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Events.cs
```diff
@@ -30,44 +32,44 @@ public void Application_LayerChanged(object sender, EventArgs e)
...
     }
-    catch { }
+
+    UpdateSavedStreams(streams);
+
+    MainViewModel.GoHome();
   }
 
   private void Application_DocumentActivated(object sender, DocumentCollectionEventArgs e)
   {
```
BovineOx, line 52: Similar to the above - should we retain and log if TL?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -223,29 +219,32 @@ string id
           // convert base (or base fallback values) and store in appobj converted prop
           if (commitObj.Convertible)
           {
-            converter.Report.Log(commitObj); // Log object so converter can access
-            try
+            if (StoredObjects.TryGetValue(commitObj.OriginalId, out Base obj))
             {
-              commitObj.Converted = ConvertObject(commitObj, converter);
+              converter.Report.Log(commitObj); // Log object so converter can access
+              commitObj.Converted = ConvertObject(obj, converter);
```
BovineOx, line 225: We're sure this call won't throw?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -519,23 +517,23 @@ void FlattenConvertedObject(object item)
 
             // add property sets if this is Civil3D
 #if CIVIL2021 || CIVIL2022 || CIVIL2023 || CIVIL2024
```
BovineOx, line 519: I know this is pre-existing, but maybe worth doing the greater than approach?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -519,23 +517,23 @@ void FlattenConvertedObject(object item)
...
-                    propertySets.Add(listObj as Dictionary<string, object>);
-                  }
-
-                  o.SetPropertySets(Doc, propertySets);
-                }
+                o.SetPropertySets(Doc, propertySets);
               }
               catch (Exception e)
               {
                 appObj.Log.Add($"Could not attach property sets: {e.Message}");
```
BovineOx, line 534: Is this top level catch? What log level would we see this as?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Selection.cs
```diff
@@ -71,24 +71,22 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
   {
-    var editor = Application.DocumentManager.MdiActiveDocument.Editor;
-    var currentSelection = editor.SelectImplied().Value?.GetObjectIds()?.ToList() ?? new List<ObjectId>();
-    foreach (var arg in args)
+    if (objs is not null && objs.Count > 0)
     {
-      try
+      Editor editor = Application.DocumentManager.MdiActiveDocument.Editor;
+
+      var currentSelection = editor.SelectImplied().Value?.GetObjectIds()?.ToList() ?? new List<ObjectId>();
```
BovineOx, line 80: Can SelectImplied() be null?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Selection.cs
```diff
@@ -71,24 +71,22 @@ public override List<ISelectionFilter> GetSelectionFilters()
...
         {
           if (Doc.Database.TryGetObjectId(handle, out ObjectId id))
           {
             if (deselect)
             {
-              if (currentSelection.Contains(id))
-              {
-                currentSelection.Remove(id);
-              }
+              currentSelection.Remove(id);
```
BovineOx, line 89: Can currentSelection actually be null here?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.cs
```diff
@@ -86,15 +87,36 @@ public override List<StreamState> GetStreamsInFile()
...
+    string docString;
+
     try
     {
-      path = GetDocPath(Doc);
+      // get the path (unsaved files will not have a path)
+      string path = GetDocPath(Doc);
+      docString = path + name;
+    }
+    catch // is there a way to check if the doc is saved before attempting to retrieve the doc path?
```
BovineOx, line 111: Probably requires the !IsFatal() pattern

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -254,95 +254,193 @@ public static bool Visible(this DBObject obj)
...
+    propSetDef.AppliesToAll = true;
+
+    // Create the definition for each property
+    foreach (var entry in propertySetDict)
+    {
+      var propDef = new PropertyDefinition();
+      propDef.SetToStandard(doc.Database);
+      propDef.SubSetDatabaseDefaults(doc.Database);
+      propDef.Name = entry.Key;
+      if (GetPropertySetType(entry.Value) is Autodesk.Aec.PropertyData.DataType dataType)
```
BovineOx, line 347: What if the type doesn't maych, we add a propDef with a null? DataType?

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -254,95 +254,193 @@ public static bool Visible(this DBObject obj)
...
+    try
+    {
+      // attempt to retrieve the property set.
+      // this will throw if the property set does not exist on the object.
+      // afaik, trycatch is necessary because there is no way to preemptively check if the set already exists.
+      temporaryId = PropertyDataServices.GetPropertySet(obj, propertySetId);
+    } 
+    catch (Autodesk.AutoCAD.Runtime.Exception)
+    {
+      // More than likely eKeyNotFound exception.
```
BovineOx, line 378: Do we not want to make sure this IS what we are seeing? What if it were not?

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -354,9 +452,9 @@ public static void SetPropertySets(this Entity entity, Document doc, List<Dictio
     {
       propertySets = PropertyDataServices.GetPropertySets(obj);
     }
-    catch (Exception e)
-    { }
-    if (propertySets == null)
+    catch (Autodesk.AutoCAD.Runtime.Exception) { }
```
BovineOx, line 455: What's the rationale for continuing? if continung is valid, we should add the rationale in the {}

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -606,7 +702,7 @@ string fileNameHash
           }
         }
       }
-      if (foundObjects.Any())
+      if (foundObjects.Count != 0)
```
BovineOx, line 705: I know this was pre-existing, but can we add a space between the end of one if block and this if statement plZ?

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -735,9 +847,9 @@ public static void SetStyle(Base styleBase, Entity entity, Dictionary<string, Ob
 
     if (styleBase["linetype"] is string lineType)
     {
-      if (lineTypeDictionary.ContainsKey(lineType))
+      if (lineTypeDictionary.TryGetValue(lineType, out ObjectId value))
```
BovineOx, line 850: What happens if we don't manage to parse this? will entity.LinetypeId be set to anything useful/correct?

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/Converter.AutocadCivil.Utils.cs
```diff
@@ -70,7 +66,7 @@ public static Base GetObjectExtensionDictionaryAsBase(this DBObject source)
 
 public partial class ConverterAutocadCivil
 {
-  public static string invalidAutocadChars = @"<>/\:;""?*|=,‘";
+  private const string INVALIDCHARS = @"<>/\:;""?*|=,‘";
```
BovineOx, line 69: I would be inclined to have INVALID_CHARS for readability

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -267,52 +264,52 @@ public ApplicationObject AlignmentToNative(Alignment alignment)
...
-          default:
-            try
-            {
-              id = CivilDB.Alignment.Create(civilDoc, name, site, layer, style, label);
-            }
-            catch
+            catch (ArgumentException)
             {
-              id = CivilDB.Alignment.Create(civilDoc, CivilDB.Alignment.GetNextUniqueName(name), site, layer, style, label);
+              goto default;
```
BovineOx, line 291: Really not keen on this. I'd suggest adding a CreateDefaultValue() private method and calling it from both places.

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -451,12 +448,12 @@ private Arc AlignmentArcToSpeckle(CivilDB.AlignmentSubEntityArc arc)
     Point2d midPoint = chordMid.Add(unitMidVector.MultiplyBy(sagitta));
     try
     {
-      if (arc.GreaterThan180) // sometimes this prop throws an exception??
+      if (arc.GreaterThan180) // this can throw
```
BovineOx, line 451: I think it is clearer if we but the comment before the line

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -451,12 +448,12 @@ private Arc AlignmentArcToSpeckle(CivilDB.AlignmentSubEntityArc arc)
...
       {
         midPoint = chordMid.Add(unitMidVector.Negate().MultiplyBy(2 * arc.Radius - sagitta));
       }
     }
-    catch { }
+    catch (InvalidOperationException){ } // The property gets an invalid value according to the entity's constraint type.
```
BovineOx, line 456: Not sur I follow the comment, are things in a usable/correct state if this exception has been caught?

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -949,35 +856,33 @@ public ApplicationObject TinSurfaceToNative(Mesh mesh, Base props)
...
+      ObjectIdCollection docStyles = new();
+      ObjectId style = ObjectId.Null;
       foreach (ObjectId styleId in civilDoc.Styles.SurfaceStyles)
       {
         docStyles.Add(styleId);
       }
-
-      var style = props["style"] as string != null ?
-        GetFromObjectIdCollection(props["style"] as string, docStyles) : civilDoc.Styles.SurfaceStyles.First();
+      if (docStyles.Count != 0 )
```
BovineOx, line 871: line space please :)

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -1142,6 +1052,7 @@ public Pipe PipeToSpeckle(CivilDB.Pipe pipe)
     // assign additional pipe props
     AddNameAndDescriptionProperty(pipe.Name, pipe.Description, specklePipe);
 
+    // TODO: use !IsFatal() here
```
BovineOx, line 1055: ouch, what is throwing here? Is it the ToString() or can some things be null?

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
```diff
@@ -871,13 +866,14 @@ public Curve SplineToSpeckle(Spline spline)
     }
 
     // handle the display polyline
-    try
+    if (spline.ToPolyline(false, true) is AcadDB.Curve poly)
```
BovineOx, line 869: Can this ToPolyline() ever not be AcadDB.Curve poly? If it were, what happens downsteam of this check?

AlanRynne (changes requested):
Setting this to "requested changes". @clairekuang is already working through them and I'll re-review once they're in

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
clairekuang: This chunk is trying to set the `displayValue` of a spline curve, so I'll modify this to log a (noncritical) error if it fails

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Selection.cs
clairekuang: None of these methods or properties including [`SelectImplied()` ](https://help.autodesk.com/view/OARX/2023/ENU/?guid=OARX-ManagedRefGuide-Autodesk_AutoCAD_EditorInput_Editor_SelectImplied) are documented to indicate they may return null or throw; I may have added these checks previously due to undocumented null exceptions but will make it consistent with `SelectImplied()`. `currentSelection` shouldn't be null since it's set to an empty list on null.

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
clairekuang: This will be done in a separate ticket to improve our preprocessor directives accross all sharp connectors

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
clairekuang: yes it returns an empty list if the conversion throws in the converter, it was only caught before because the converter method wasn't wrapped with a trycatch

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Events.cs
```diff
@@ -30,44 +32,44 @@ public void Application_LayerChanged(object sender, EventArgs e)
   //checks whether to refresh the stream list in case the user changes active view and selects a different document
   private void Application_WindowActivated(object sender, DocumentWindowActivatedEventArgs e)
   {
-    try
+    if (e.DocumentWindow?.Document == null || UpdateSavedStreams == null)
```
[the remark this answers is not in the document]
clairekuang: good point - will use the !IsFatal() pattern here

ConnectorAutocadCivil/ConnectorAutocadCivil/Storage/SpeckleStreamManager.cs
clairekuang: it may be valid to fail in the case of old files with streams saved in an incorrect format (has happened in alpha versions of the connector) in which case we'd want to override the saved streams. will add logging here however

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
clairekuang: `DataType` is an enum, so I believe it will default to `int`. This is a good catch though, I will skip this if the type cannot be determined and log it instead.

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
clairekuang: No exceptions were documented on the `GetPropertySet` method, from previous testing the runtime exception had an `eKeyNotFound` message. Added !IsFatal() here

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
clairekuang: Continuing because this may throw if the object simply doesn't have property sets - there is no way to check for property sets before attempting to retrieve them on an object with `GetPropertySets` afaik. adding it into comments

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
clairekuang: It will end up using the default linetype, which would be expected behavior

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
clairekuang: yes, clarified in the comments

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
clairekuang: Accessing some properties on the `pipe` object may throw for undocumented reasons (usually if the property doesn't exist or isn't applicable). It's worth trying to add as many as possible to the pipe, not knowing which may throw, thus the individual try catches

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/Converter.AutocadCivil.Utils.cs
```diff
@@ -126,14 +122,31 @@ public static void AddNameAndDescriptionProperty(string name, string description
...
       return false;
     }
+    catch (FormatException)
+    {
+      return false;
+    }
+    catch (OverflowException)
+    {
+      return false;
+    }
```
AlanRynne, line 146: No change needed, but we've been discussing doing something like:

```csharp
catch (Exception ex) when (ex is ArgumentException or FormatException or OverflowException) {...}
```
When the handling of all cases is in essence the same, but we haven't settled on a preference yet. Just FYI this can be done.

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/Converter.AutocadCivil.Utils.cs
```diff
@@ -233,6 +246,42 @@ public ObjectId GetFromObjectIdCollection(string name, ObjectIdCollection collec
...
+
+      LayerTableRecord newLayer = new() { Name = name };
+      try
+      {
+        layerTable.Add(newLayer);
+        Trans.AddNewlyCreatedDBObject(newLayer, true);
+        layer = newLayer;
+        return true;
+      }
+      catch { } // TODO: use !IsFatal() here
```
AlanRynne, line 279: Missing IsFatal

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
```diff
@@ -1556,14 +1509,16 @@ public PolyFaceMesh MeshToNativeDB(Mesh mesh)
...
           {
-            Color color = Color.FromArgb(mesh.colors[i]);
-            vertex.Color = Autodesk.AutoCAD.Colors.Color.FromRgb(color.R, color.G, color.B);
+            if (Color.FromArgb(mesh.colors[i]) is Color color)
+            {
+              vertex.Color = Autodesk.AutoCAD.Colors.Color.FromRgb(color.R, color.G, color.B);
+            }
           }
-          catch { }
+          catch { } // TODO: use !IsFatal here
```
AlanRynne, line 1521: Missing exception fix here, at least the `!ex.IsFatal()`

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
```diff
@@ -1634,8 +1589,7 @@ public PolyFaceMesh MeshToNativeDB(Mesh mesh)
         area = solid.Area;
         volume = solid.MassProperties.Volume;
       }
-      catch (Exception e) { }
-      ;
+      catch { } // TODO: use !IsFatal() here
```
AlanRynne, line 1592: And here 👍🏻

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
```diff
@@ -1709,7 +1663,7 @@ public PolyFaceMesh MeshToNativeDB(Mesh mesh)
           };
         }
       }
-      catch (Exception e)
+      catch (Exception e) // TODO: use !IsFatal() here
```
AlanRynne, line 1666: And here

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Other.cs
```diff
@@ -405,22 +403,11 @@ public ApplicationObject HatchToNativeDB(Hatch hatch)
...
-            convertedItem: entry.Key,
-            logItem: $"Could not delete loop {loopHandle}: {e.Message}"
-          );
-        }
+        newHatch.AppendLoop(entry.Value, new ObjectIdCollection() { entry.Key.ObjectId });
+        newHatch.EvaluateHatch(true);
+        entry.Key.Erase(); // delete created hatch loop curve
       }
-      catch (Exception e)
+      catch (Exception e) // TODO: use !IsFatal() here
```
AlanRynne, line 410: Missing IsFatal

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Other.cs
```diff
@@ -1428,95 +1386,91 @@ public AcadDB.Dimension AcadDimensionToNative(Dimension dimension)
...
+
         break;
       default:
         break;
     }
-    //if (_dimension != null)
-    // _dimension.TextPosition = PointToNative(dimension.textPosition);
-    return _dimension;
+
+    return autocadDimension;
```
AlanRynne, line 1473: We really should stop having 1000+ LOC in a single file 😢

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.cs
```diff
@@ -102,167 +104,189 @@ public Base ConvertToSpeckle(object @object)
...
+    }
+    catch (SpeckleException e)
+    {
+      SpeckleLog.Logger.Warning(e, "{exceptionMessage}");
+      reportObj?.Update(
+        status: ApplicationObject.State.Failed,
+        logItem: $"{@object.GetType()} unhandled conversion error: {e.Message}\n{e.StackTrace}"
+      );
+    }
+    catch (Exception e) // TODO: use !IsFatal here
```
AlanRynne, line 284: Missing IsFatal

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.cs
```diff
@@ -308,125 +332,138 @@ public object ConvertToNative(Base @object)
...
+    }
+    catch (SpeckleException e)
+    {
+      SpeckleLog.Logger.Warning(e, "{exceptionMessage}");
+      reportObj?.Update(
+        status: ApplicationObject.State.Failed,
+        logItem: $"{@object.GetType()} unhandled conversion error: {e.Message}\n{e.StackTrace}"
+      );
+    }
+    catch (Exception e) // TODO: use !IsFatal() here
```
AlanRynne, line 442: Missing IsFatal

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -1261,7 +1183,7 @@ public Base CorridorToSpeckle(CivilDB.Corridor corridor)
           surfaces.Add(mesh);
         }
       }
-      catch (Exception e) 
+      catch (Exception e) // TODO: use !IsFatal() here
```
AlanRynne, line 1186: Missing IsFatal

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.cs
```diff
@@ -102,167 +104,189 @@ public Base ConvertToSpeckle(object @object)
...
-          Report.UpdateReportObject(reportObj);
-        }
-        return @base;
+            default:
+#if ADVANCESTEEL
+              try
+              {
+                @base = ConvertASToSpeckle(obj, reportObj, notes);
+              }
+              catch (Exception)
```
AlanRynne, line 230: Potentially missing `IsFatal`, not sure we want to try and update the report when `OutOfMemory` for example.

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -354,9 +458,13 @@ public static void SetPropertySets(this Entity entity, Document doc, List<Dictio
     {
       propertySets = PropertyDataServices.GetPropertySets(obj);
     }
-    catch (Exception e)
-    { }
-    if (propertySets == null)
+    catch (Autodesk.AutoCAD.Runtime.Exception e) when (!e.IsFatal())
```
AlanRynne, line 461: I don't think we need `IsFatal` here as an `Autocad` specific exception doesn't overlap with any of the exceptions we consider "unrecoverable" in the IsFatal implementation (they all derive from System.Exception)

ConnectorAutocadCivil/ConnectorAutocadCivil/Utils.cs
```diff
@@ -254,95 +255,198 @@ public static bool Visible(this DBObject obj)
...
-          PropertyDataServices.AddPropertySet(entity, propSetDef.ObjectId);
-          tr.Commit();
-        }
+        dictPropSetDef.AddNewRecord(propSetDefName, propSetDef);
+        tr.AddNewlyCreatedDBObject(propSetDef, true);
+        // add property set to the object
+        AddPropertySetToObject(entity, propSetDef.ObjectId);
       }
-      catch { }
+      catch (Autodesk.AutoCAD.Runtime.Exception) { }
```
AlanRynne, line 440: Would we want to log this at least?

AlanRynne (commented):
Looks mostly good. Just some missing `IsFatal()` calls, a typo and a missing log call 👍🏻

AlanRynne (approved):
This is looking great now 🙌🏼

PR #3123, opened 2024-01-08
refactor(objects): handles nullability warnings

clairekuang:
Handles nullability warnings in objects by fixing explicit nullability on props and args.


Objects/Objects/BuiltElements/Floor.cs
```diff
@@ -95,9 +97,9 @@ namespace Objects.BuiltElements.Archicad
   public sealed class ArchicadFloor : Floor
   {
     // Element base
-    public string? elementType { get; set; } /*APINullabe*/
```
JR-Morgan, line 98: @jozseflkiss remind me, what were these `APINullable` comments for? 
And, are Claire's changes to this class correct? I seem to remember you were experimenting with keeping some of these props non-nullable for Archicad -> Speckle, but nullable for Archicad C# -> C++

JR-Morgan:
There's still a couple CS8625 violations perhaps were missed:
image

fyi, you can (temporarly) make a specific analyser an error to easily check if you've caught them all. 
```xml
<WarningsAsErrors>$(WarningsAsErrors);CS8625</WarningsAsErrors>
```

JR-Morgan:
Also, was the goal to solve other nullability warnings. I see a couple `CS8619`, `CS8766`, `CS8603`, and `CS8604`?
I think we were just focusing on CS8625, so no need to solve these yet.

Objects/Objects/Utils/MeshTriangulationHelper.cs
```diff
@@ -242,22 +242,16 @@ public Vector3(double x, double y, double z)
...
+    public static Vector3 operator +(Vector3 a, Vector3 b) => new(a.x + b.x, a.y + b.y, a.z + b.z);
 
-    public static Vector3 operator -(Vector3 a, Vector3 b)
-    {
-      return new Vector3(a.x - b.x, a.y - b.y, a.z - b.z);
-    }
+    public static Vector3 operator -(Vector3 a, Vector3 b) => new(a.x - b.x, a.y - b.y, a.z - b.z);
 
-    public double Dot(Vector3 v)
+    public readonly double Dot(Vector3 v)
```
JR-Morgan, line 249: Nice!

Objects/Objects/Structural/Material/Steel.cs
```diff
@@ -40,7 +40,6 @@ public class Steel : StructuralMaterial
     this.poissonsRatio = poissonsRatio;
     this.shearModulus = shearModulus;
     this.density = density;
-    thermalExpansivity = thermalExpansivity;
```
JR-Morgan, line 43: 🤦‍♂️

Objects/Objects/Geometry/Point.cs
```diff
@@ -150,50 +150,36 @@ public void Deconstruct(out double x, out double y, out double z)
...
 
   public static bool operator ==(Point? point1, Point? point2)
   {
     if (point1 is null && point2 is null)
     {
       return true;
     }
-
-    if (point1 is null ^ point2 is null)
+    else if (point1 is null ^ point2 is null)
```
JR-Morgan, line 174: Could we change this operator to be:

I don't know why we'd want to use the binary XOR operator... Since the previous case covers `&&`
And doing a `||` also makes the `?` in the property comparison unnecessary.

```cs
  public static bool operator ==(Point? point1, Point? point2)
  {
    if (point1 is null && point2 is null)
    {
      return true;
    }
    else if (point1 is null || point2 is null)
    {
      return false;
    }

    return point1.units == point2.units && point1.x == point2.x && point1.y == point2.y && point1.z == point2.z;
  }
  ```

File: Objects/Objects/Geometry/Line.cs [hunk 5]
```diff
@@ -14,15 +14,16 @@ public class Line : Base, ICurve, IHasBoundingBox, ITransformable<Line>
 {
   public Line() { }
 
-  public Line(double x, double y, double z = 0, string units = Units.Meters, string applicationId = null)
+  [Obsolete("Line should not use a constructor that only sets the start point. Deprecated in 2.18.", true)]
+  public Line(double x, double y, double z = 0, string units = Units.Meters, string? applicationId = null)
   {
     start = new Point(x, y, z);
-    end = null;
+    end = null; // constructor is marked obsolete because we don't want to allow setting a null end point here.
```
JR-Morgan, line 21: Is it worth us supressing the CA warning on this line? In core, we've supressed warnings in Obsolete functions (since they are legacy code, and can't/shan't be fixed)

Objects/Objects/BuiltElements/Room.cs
```diff
@@ -28,26 +28,26 @@ public Room(string name, string number, Level level, [SchemaMainParam] Point bas
     /// <summary>
     /// SchemaBuilder constructor for a Room
     /// </summary>
-    /// <remarks>Assign units when using this constructor due to <paramref name="height"/> param</remarks>
+    /// <remarks>Assign units when using this constructor due to <propref name="height"/> prop</remarks>
```
JR-Morgan, line 31: Is `propref` valid tag? Rider doesn't recognise it. I've been using `<see cref="height"/>` to reference properties

JR-Morgan (changes requested):

Objects/Objects/Utils/MeshTriangulationHelper.cs
JR-Morgan: Actually, I this has made me think, we could probably look into making the entire struct `readonly`...
(tho, nothing more needed in this PR)

clairekuang:
> Also, was the goal to solve other nullability warnings. I see a couple `CS8619`, `CS8766`, `CS8603`, and `CS8604`? I think we were just focusing on CS8625, so no need to solve these yet.

Mostly targetting CS8625 with this pr, but I did resolve most of the other ones as well - any left behind can be tackled in the `misc` isse for Object warnings

JR-Morgan (approved):

PR #3126, opened 2024-01-10
fix(grasshopper): CNX-8395 Handle CA1031 warnings in Grasshopper projects

AlanRynne:
https://spockle.atlassian.net/browse/CNX-8395


ConnectorGrasshopper/ConnectorGrasshopperShared/SchemaBuilder/CreateSchemaObject.cs
```diff
@@ -604,9 +605,17 @@ private object ConvertType(Type type, object value, string name)
       {
         return Convert.ChangeType(value, type);
       }
-      catch (Exception e)
+      catch (InvalidCastException e)
```
JR-Morgan, line 608: Would we prefer to use a `when` cause here instead (like elsewhere?)

```cs
catch (Exception ex) when (ex is InvalidCastException or FormatException or OverflowException)
```

Not that this is bad, I'm more curious if there's a reason you've used this pattern here, and the `when` pattern elsewhere

ConnectorGrasshopper/ConnectorGrasshopperShared/SchemaBuilder/CreateSchemaObjectBase.cs
```diff
@@ -573,9 +571,17 @@ private object ConvertType(Type type, object value, string name)
       {
         return Convert.ChangeType(value, type);
       }
-      catch (Exception e)
+      catch (InvalidCastException e)
```
JR-Morgan, line 574: Same as above

ConnectorGrasshopper/ConnectorGrasshopperShared/Streams/Deprecated/StreamListComponent.cs
```diff
@@ -109,12 +115,12 @@ public override void SolveInstanceWithLogContext(IGH_DataAccess DA)
...
           var client = new Client(account);
           // Save the result
-          var result = await client.StreamsGet(limit);
+          var result = await client.StreamsGet(limit).ConfigureAwait(false);
           streams = result
             .Select(stream => new StreamWrapper(stream.id, account.userInfo.id, account.serverInfo.url))
             .ToList();
         }
-        catch (Exception e)
+        catch (SpeckleGraphQLException e)
```
JR-Morgan, line 123: One of the tickets r.e. Core is ensuring that the client functions have clear exception paths. I'm 70% they will only throw `SpeckleGraphQlExceptions`. But I'm not 100% sure, since we've definitely   seen some suspicious null responses from the server.

JR-Morgan (changes requested):

ConnectorGrasshopper/ConnectorGrasshopperShared/Streams/Deprecated/StreamListComponent.cs
AlanRynne: I'm reverting this since this class is obsolete. No need to change it.

AlanRynne (commented):
I pushed my changes before posting my response to some of your comments so all this is outdated now. Leaving it for posterity.

ConnectorGrasshopper/ConnectorGrasshopperShared/SchemaBuilder/CreateSchemaObject.cs
AlanRynne: Resolved but the comment still stands

JR-Morgan (approved):

PR #3130, opened 2024-01-10
Chore(Tests): CNX-8640 core integration test project

JR-Morgan:
- Resolved naming violations in Integration test projects
- Updated editor config to warn when using the old nunit functions
- Updated both Unit and integration tests to use the new unit functions
- Removed ConfigureAwaits from test project:
    - My reasoning is. They aren't needed, don't provide value, just adds visual clutter.
    - Happy to discuss and revert if prefered.

### DIFF
--- a/Core/Tests/Speckle.Core.Tests.Unit/Api/GraphQLClient.cs
+++ b/Core/Tests/Speckle.Core.Tests.Unit/Api/GraphQLClient.cs
@@ -80,16 +80,14 @@ public sealed class GraphQLClientTests : IDisposable
         .ExecuteWithResiliencePolicies(
           async () =>
             await Task.Run(
-                async () =>
-                {
-                  await Task.Delay(1000).ConfigureAwait(false);
-                  return "foo";
-                },
-                tokenSource.Token
-              )
-              .ConfigureAwait(false)
-        )
-        .ConfigureAwait(false);
+              async () =>
+              {
+                await Task.Delay(1000);
+                return "foo";
+              },
+              tokenSource.Token
+            )
+        );
     });
     timer.Stop();
     var elapsed = timer.ElapsedMilliseconds;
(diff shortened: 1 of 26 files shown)


Core/Tests/Speckle.Core.Tests.Unit/Api/GraphQLClient.cs
```diff
@@ -64,9 +64,7 @@ public void TestExceptionThrowingFromGraphQLErrors(Type exType, Map extensions)
   [Test]
   public void TestMaybeThrowsDoesntThrowForNoErrors()
   {
-    _client.MaybeThrowFromGraphQLErrors(new GraphQLRequest(), new GraphQLResponse<string>());
-    // We're just checking that the prev function didn't throw
-    Assert.True(true);
```
AlanRynne, line 69: Ohh WTH 🤣

Core/Tests/Speckle.Core.Tests.Unit/Api/Operations/OperationsReceiveTests.Exceptional.cs
```diff
@@ -33,7 +33,7 @@ public void Receive_OperationCanceled_ExceptionThrown(string id)
     ctc.Cancel();
 
     MemoryTransport emptyTransport2 = new();
-    Assert.ThrowsAsync<OperationCanceledException>(async () =>
+    Assert.CatchAsync<OperationCanceledException>(async () =>
```
AlanRynne, line 36: Is `CatchAsync` the recommended pattern to assert an exception type here? Asking out of ignorance

AlanRynne (approved):
No complaints, just a question 👇🏼 Approving anyways

Core/Tests/Speckle.Core.Tests.Unit/Api/Operations/OperationsReceiveTests.Exceptional.cs
JR-Morgan: ThrowsAsync tests if the method throws the EXACT type
CatchAsync allows subtypes

In this case, the spirit of the test was to make sure it an `OperationCancelledException`, we don't care if its a subtype like `TaskCancelledException`, so I after learning of the existence of `CatchAsync`, I decided to update the functions where we would be ok with subtypes.

PR #3131, opened 2024-01-11
Feat (Automate): CNX-8636 automate sdk breaking change

oguzhankoral:
Read function run data from provided path via CLI argument on `run` command.

### DIFF
--- a/Automate/Speckle.Automate.Sdk/Runner.cs
+++ b/Automate/Speckle.Automate.Sdk/Runner.cs
@@ -1,9 +1,9 @@
 using System.CommandLine;
 using System.Diagnostics.CodeAnalysis;
+using Newtonsoft.Json.Schema;
 using Newtonsoft.Json.Schema.Generation;
 using Newtonsoft.Json.Serialization;
 using Speckle.Automate.Sdk.Schema;
-using Speckle.Newtonsoft.Json;
 
 namespace Speckle.Automate.Sdk;
 
@@ -21,7 +21,9 @@ public static class AutomationRunner
   )
     where TInput : struct
   {
-    var automationContext = await AutomationContext.Initialize(automationRunData, speckleToken).ConfigureAwait(false);
+    AutomationContext automationContext = await AutomationContext
+      .Initialize(automationRunData, speckleToken)
+      .ConfigureAwait(false);
 
     try
     {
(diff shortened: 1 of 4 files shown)


Automate/Speckle.Automate.Sdk/Runner.cs
```diff
@@ -1,9 +1,9 @@
 using System.CommandLine;
 using System.Diagnostics.CodeAnalysis;
+using Newtonsoft.Json.Schema;
 using Newtonsoft.Json.Schema.Generation;
 using Newtonsoft.Json.Serialization;
```
AlanRynne, line 5: Ping @JR-Morgan as I'm fairly certain we should not be using Newtonsoft.Json and instead use our own `Speckle.Newtonsoft.Json`

This may be an oversight that precedes this PR though, judging by the other 2 lines

Automate/Speckle.Automate.Sdk/Schema/FunctionRunDataParser.cs
```diff
@@ -0,0 +1,43 @@
...
+  /// Function run data parser from json file path./>
+  /// </summary>
+  /// <param name="inputLocation"> Path to retrieve function run data.</param>
+  /// <typeparam name="T"> Type for function inputs.</typeparam>
+  /// <returns>The data to be able run function.</returns>
+  /// <exception cref="SerializationException"> Throws if deserialized object is null.</exception>
+  public static FunctionRunData<T> FromPath<T>(string inputLocation)
+  {
+    string inputJsonString = ReadInputData(inputLocation);
+    FunctionRunData<T>? functionRunData = JsonConvert.DeserializeObject<FunctionRunData<T>>(inputJsonString);
```
AlanRynne, line 18: Note that `FromPath` can also throw a `FileNotFoundException` coming from the call `ReadInputData(...)` on line 17.
The xml docs should reflect this

Automate/Tests/Speckle.Automate.Sdk.Tests.Integration/SpeckleAutomate.cs
```diff
@@ -1,3 +1,4 @@
+using Newtonsoft.Json;
```
AlanRynne, line 1: Same, should this be `Speckle.Newtonsoft.Json`?

AlanRynne (commented):
Looks good, I've pinged Jedd regarding the use of Newtonsoft.Json

Automate/Speckle.Automate.Sdk/Schema/FunctionRunDataParser.cs
oguzhankoral: good catch

oguzhankoral: Added. Resolving conservation

Automate/Tests/Speckle.Automate.Sdk.Tests.Integration/SpeckleAutomate.cs
AlanRynne: Resolving, it is correct as is but having both `Speckle.Newtonsoft.Json` and `Newtosoft.Json` in the same project will get weird.

I'll open a ticket for this in the backlog

AlanRynne (approved):
🪄 Approving and merging in.

PR #3137, opened 2024-01-15
chore(objects): Resolve namespace warnings

clairekuang:
Adjusts folder structure for all namespace warnings in `Objects`. Organization is based on namespaces - no namespaces were adjusted.


JR-Morgan (approved):

Objects/Objects/BuiltElements/Revit/RevitRoof.cs
```diff
@@ -0,0 +1,96 @@
+using System.Collections.Generic;
+using Objects.Geometry;
+using Objects.Utils;
+using Speckle.Core.Kits;
+using Speckle.Core.Models;
+
+namespace Objects.BuiltElements.Revit.RevitRoof;
```
JR-Morgan, line 7: I see this one still generates a warning... But this namespace is questionable at best.
Perhaps we can use our "deprecated" namespacing feature of the serializer to deprecate + move these incorrectly namespace classes.

Perhaps to be done in another PR

clairekuang: Hmm for now I've created a `RevitRoof` folder for this file, but it does seem like the namespace is incorrect. We can fix these in another ticket

PR #3138, opened 2024-01-15
fix(core): CNX-8692 Fixed regression with AccountManager updating accounts while loading them from SQLite

JR-Morgan:


Core/Core/Transports/SQLite.cs
```diff
@@ -257,8 +257,10 @@ public void UpdateObject(string hash, string serializedObject)
   {
     CancellationToken.ThrowIfCancellationRequested();
 
+    using var c = new SqliteConnection(_connectionString);
```
JR-Morgan, line 260: Our AccountManager does some potentially sketchy operations calling `UpdateObject` while looping through the lazily enumerate return of `GetAllObjects`. This causes `GetAllObjects` to return forever, since it returns your updated objects also.

Reverting my change so `GetAllObjects` and `UpdateObject` use separate SQLite connections appears to fix things.
I'm a little concerned that we're actually relying on some sort of in-memory cache that the connections use.

I question if this is the behaviour we actually want, it seems to me the problem is with the way the `AccountManager` is enumerating over a lazy enumerable that its updating. I'll open a ticket for next release to decide on what our desired behaviour here should be.

AlanRynne: I agree. We should do some investigation as to why this is the way it works now, as this sounds super dodgy...

Core/Tests/Speckle.Core.Tests.Unit/Transports/SQLiteTransportTests.cs
```diff
@@ -91,6 +91,42 @@ public void SaveAndRetrieveObject_Sync()
     }
   }
 
+  [Test(
+    Description = "Tests that it is possible to enumerate through all objects of the transport while updating them, without getting stuck in an infinite loop"
+  )]
```
AlanRynne, line 96: Nice ♥️ I'd love you to run me by this just to be 100% sure I'm interpreting this test right

AlanRynne (approved):
Looking good! Live reviewed with Jedd

AlanRynne:
Merging in! Remember to open a ticket for the unresolved conversation above @JR-Morgan 🙌🏼

PR #3139, opened 2024-01-15
Chore(core): CNX-8689 - Example App Removed

JR-Morgan:
Removed the example app after agreeing with Team that it is old, out of date, and does not serve as a useful example suite to keep around.

If we decide we need a proper suite of examples, then we will open the discussion on what that would best look like.


AlanRynne (approved):
YES and thank you 🥳

PR #3141, opened 2024-01-16
fix(grasshopper): CNX-8642 resolving account by id from the accounts list will return the first account found in a server running fe1 fe2

AlanRynne:
Fixes ambiguity between accounts by no longer trying to find an account exclusively by it's user ID.

The accounts are now distinguished by their "Local Identifier", which is basically just the server URL with an `id` query parameter holding the user ID.

`https://speckle.xyz/?id=USER_ID`

This guarantees uniqueness of each user+server pair.

Changes in this PR include:
- `Account` class: Addition of internal `GetLocalIdentifier`
- `AccountManager` class:Addition of `GetLocalIdentifierForAccount` and `GetAccountForLocalIdentifier`
- `Accounts.ListAccounts` GH component: Now uses the above AccountManager methods everywhere it's needed. It also now outputs actual `Account` instances instead of the previous user ID, so the solution is now more robust.
- `GH_SpeckleAccount` class: No longer requires casting from strings, as the only place this was used was in the AccountList component, and this no longer outputs strings.

### DIFF
--- a/Core/Core/Credentials/Account.cs
+++ b/Core/Core/Credentials/Account.cs
@@ -1,6 +1,9 @@
 #nullable disable
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
+
 using Speckle.Core.Api;
 using Speckle.Core.Helpers;
 using Speckle.Core.Logging;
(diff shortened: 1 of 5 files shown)


Core/Core/Credentials/AccountManager.cs
```diff
@@ -353,6 +353,44 @@ public static void ChangeDefaultAccount(string id)
...
+  /// Retrieves the local identifier for the specified account.
+  /// </summary>
+  /// <param name="account">The account for which to retrieve the local identifier.</param>
+  /// <returns>The local identifier for the specified account in the form of "SERVER_URL?u=USER_ID".</returns>
+  /// <remarks>
+  /// Notice that the generated Uri is not intended to be used as a functioning Uri, but rather as a
+  /// unique identifier for a specific account in a local environment. The format of the Uri, containing a query parameter with the user ID,
+  /// serves this specific purpose. Therefore, it should not be used for forming network requests or
+  /// expecting it to lead to an actual webpage. The primary intent of this Uri is for unique identification in a Uri format.
+  /// </remarks>
```
AlanRynne, line 366: @JR-Morgan can I inherit this remarks from the Account.GetLocalIdentifier doc?

Core/Core/Credentials/Account.cs
```diff
@@ -89,4 +89,30 @@ public override bool Equals(object obj)
...
+  ///   <code>
+  ///     Uri localIdentifier = GetLocalIdentifier();
+  ///     Console.WriteLine(localIdentifier);
+  ///   </code>
+  ///   For a fictional `User ID: 123` and `Server: https://speckle.xyz`, the output might look like this:
+  ///   <code>
+  ///     https://speckle.xyz?id=123
+  ///   </code>
+  /// </example>
+  internal Uri GetLocalIdentifier() => new($"{serverInfo.url}?id={userInfo.id}");
```
AlanRynne, line 117: @JR-Morgan I'm aware that this may potentially throw, as we've seen cases where the serverInfo or userInfo is not populated (i.e. just do `new Account()` 😅)

Not sure if I should code defensively against it. WDYT

JR-Morgan (approved):

clairekuang (approved):

PR #3142, opened 2024-01-16
Fix(Logger): CNX-8443 speckle connectors overwrite static logger instance

oguzhankoral:
Previously we set out logger instance onto static `Log.Logger` and properties onto static `GlobalLogContext`. Now we set global properties on `LoggerConfiguration` instance before creating instance from `ILogger`.

Update: We only set `id` property after creating `ILogger`. It was causing infinite loop on the case of having no account.


Core/Core/Logging/SpeckleLog.cs
```diff
@@ -271,23 +292,16 @@ private static void AddUserIdToGlobalContextFromDefaultAccount()
...
+      // To log it after Logger initialized as deferred action.
+      s_isMachineIdUsed = true;
     }
-    GlobalLogContext.PushProperty("id", id);
-
-    SentrySdk.ConfigureScope(scope =>
-    {
-      scope.User = new User { Id = id };
-    });
+    return id;
```
AlanRynne, line 298: Should `s_isMachineIdUsed` also be added when `defaultAccount == null` (opposite condition than line 288)

Core/Core/Logging/SpeckleLog.cs
```diff
@@ -236,11 +241,27 @@ SpeckleLogConfiguration logConfiguration
...
+
+    // Configure scope after logger created.
+    SentrySdk.ConfigureScope(scope =>
+    {
+      scope.User = new User { Id = id };
+    });
+    SentrySdk.ConfigureScope(scope =>
+    {
+      scope.SetTag("hostApplication", hostApplicationName);
+    });
```
AlanRynne, line 253: I'm assuming this scope is global as it has no reference to the logger instance.

This would collide with anybody else using Sentry on the same AppDomain

Core/Core/Logging/SpeckleLog.cs
AlanRynne, line 298: Should s_isMachineIdUsed also be added when defaultAccount == null (opposite condition than line 288)

Core/Core/Logging/SpeckleLog.cs
```diff
@@ -236,11 +241,27 @@ SpeckleLogConfiguration logConfiguration
...
+
     if (logConfiguration.LogToFile && !canLogToFile)
     {
       logger.Warning("Log to file is enabled, but cannot write to {LogFilePath}", logFilePath);
     }
 
+    if (s_isMachineIdUsed)
+    {
+      logger.Warning("Cannot set user id for the global log context.");
```
AlanRynne, line 262: Would we want to add more than just a warning log here?

I can see adding a ´.Enrich.WithProperty("isMachineId", true)` would allow us to separate logs that came from users with valid computed IDs versus the ones that come from users with machine ID.

Not sure if this would bring much benefit though. @BovineOx WDYT?

Core/Core/Logging/SpeckleLog.cs
```diff
@@ -271,23 +292,16 @@ private static void AddUserIdToGlobalContextFromDefaultAccount()
     }
```
AlanRynne, line 292: Note that, as far as I'm aware, this part of the code is not the same ID we're computing in `Account`. @JR-Morgan not sure what to make of this as "it has been working", but seems like a potential for missmatch.

Core/Core/Logging/SpeckleLog.cs
```diff
@@ -171,9 +166,19 @@ SpeckleLogConfiguration logConfiguration
...
       .Enrich.FromLogContext()
+      .Enrich.WithProperty("id", id)
+      .Enrich.WithProperty("version", fileVersionInfo.FileVersion)
+      .Enrich.WithProperty("productVersion", fileVersionInfo.ProductVersion)
+      .Enrich.WithProperty("hostOs", DetermineHostOsSlug())
+      .Enrich.WithProperty("hostOsVersion", Environment.OSVersion)
+      .Enrich.WithProperty("hostOsArchitecture", RuntimeInformation.ProcessArchitecture.ToString())
+      .Enrich.WithProperty("runtime", RuntimeInformation.FrameworkDescription)
+      .Enrich.WithProperty("hostApplication", $"{hostApplicationName}{hostApplicationVersion ?? ""}")
       .Enrich.FromGlobalLogContext();
```
AlanRynne, line 182: If we enrich from the global log context, and this context has been polluted by a another plugin, does this mean we'd be getting properties that are not ours?

We should prevent this as much as preventing our properties leaking to other Sentry consumers.

AlanRynne (changes requested):
Flagging as requesting changes... but it's more to prevent accidental merging while we discuss these points.

AlanRynne:
@oguzhankoral it seems you've introduce an infinite loop in your `GetUserIdFromDefaultAccount` changes, methods that are used to initialise the logger cannot call the logger themselves, or the logger will call Initialise automatically.

Screenshot 2024-01-16 at 14 01 52

Core/Core/Logging/SpeckleLog.cs
oguzhankoral: That's a valid point, I'll test it without this enrichment.

Core/Core/Logging/SpeckleLog.cs
AlanRynne: Just realised this was supposed to be a multi-line comment but it's not... unhelpful

Core/Core/Logging/SpeckleLog.cs
oguzhankoral: When I remove this, nothing really effected on logs

JR-Morgan (changes requested):
From several places in our code, We're pushing scoped logs via the `LogContext.PushProperty` (see `Operations.Receive`) for an example.

Discussed briefly with Alan and Oguzhan, we should find an alternative for these to use `SpeckleLog` explicitly

Core/Core/Logging/SpeckleLog.cs
oguzhankoral: you're right! done

Core/Core/Logging/SpeckleLog.cs
oguzhankoral: it is make sense and added.

Core/Core/Logging/SpeckleLog.cs
oguzhankoral: I have to update this comment, because I was checking it on Seq. It should have meaningful impact on sentry.

JR-Morgan (approved):

oguzhankoral:
Pulled from `dev` again and @JR-Morgan's tests are passed with this PR. @AlanRynne if you don't have an objection it should be good to go.

Core/Core/Logging/SpeckleLog.cs
AlanRynne: Resolving in favour of https://spockle.atlassian.net/browse/CNX-8750

AlanRynne (approved):

PR #3143, opened 2024-01-16
chore(objects): CNX-8709 suppress warnings in obsolete classes, methods, and fields

clairekuang:
- All existing obsolete tags now throw an error
- Converters using obsolete classes, fields, and methods are updated
- Warnings in obsolete tagged objects are now suppressed
- `Network` class and related classes are now marked as `Obsolete` (instead of just the constructor).
- Includes some other minor warning resolutions


Objects/Objects/Other/Instance.cs
```diff
@@ -39,7 +39,7 @@ protected virtual IEnumerable<Base> GetTransformableGeometry()
     var displayValueRule = TraversalRule
       .NewTraversalRule()
       .When(DefaultTraversal.HasDisplayValue)
-      .ContinueTraversing(_ => DefaultTraversal.displayValueAndElementsPropAliases);
+      .ContinueTraversing(_ => DefaultTraversal.GeometryPropAliases);
```
JR-Morgan, line 42: Oh no! 
There's a mistake in my Obsolete tag for `displayValueAndElements`
The correct name is `DisplayValueAndElementsPropAliases.

image

Please can you correct this, and the `Obsolete` message in Core

Objects/Objects/BuiltElements/Revit/RevitCeiling.cs
```diff
@@ -62,7 +66,7 @@ public class RevitCeiling : Ceiling
   public double slope { get; set; }
   public Line? slopeDirection { get; set; }
 
-  [Obsolete("Offset property is now captured in parameters to match the behavior of other Revit objects")]
+  [Obsolete("Offset property is now captured in parameters to match the behavior of other Revit objects", true)]
```
JR-Morgan, line 69: Are we happy with this being `error: true`?

Objects/Objects/Geometry/BrepTrim.cs
```diff
@@ -32,6 +32,8 @@ int endIndex
     IsoStatus = isoStatus;
     TrimType = trimType;
     IsReversed = reversed;
+    StartIndex = startIndex;
```
JR-Morgan, line 35: Why were we not doing this before?

JR-Morgan (changes requested):

Objects/Objects/BuiltElements/Revit/RevitCeiling.cs
clairekuang: I think yes in this case, given that there were no references to `offset` in the converters

Objects/Objects/Other/Instance.cs
clairekuang: Yup will do!

Objects/Objects/Geometry/BrepTrim.cs
clairekuang: I'm not sure, @AlanRynne was there a reason you can remember for not setting the `StartIndex` and `EndIndex` props even though they are arguments passed to the constructor?

AlanRynne: I find no reason why we would ever do this, I'll dig in a bit more after standup.

AlanRynne: I double checked the usages of this constructor and it is definitely a bug/oversight.

This does not affect conversions for the most part, as the start/end vertex index is rarely used... but I'm sure this would fix some odd Brep edge-cases we may have been unaware of.

It will also ensure breps always have the same id (as missing to set start/end vertex would in fact result in a different id for that brep)

AlanRynne (approved):
This looks good. I added my 2 cents on the BrepTrim discussion, but it's safe to merge!

JR-Morgan (approved):

PR #3145, opened 2024-01-16
fix(dynamo): CNX-8394 Handle CA1031 warnings in Dynamo projects

AlanRynne:
https://spockle.atlassian.net/browse/CNX-8394

Here goes nothing! I have to confess I've had to held back on refactoring half the codebase here... so I've made heavy use of `IsFatal` as there were places where it was unclear what could be thrown.

PR removes all CA1031 warnings from the Dynamo project, and a minor refactor of 1 horrible method with an very old angry comment above it 😅

### DIFF
--- a/ConnectorDynamo/ConnectorDynamo.slnf
+++ b/ConnectorDynamo/ConnectorDynamo.slnf
@@ -23,7 +23,7 @@
       "Objects\\Converters\\ConverterRevit\\ConverterRevit2024\\ConverterRevit2024.csproj",
       "Objects\\Converters\\ConverterRevit\\ConverterRevitShared\\ConverterRevitShared.shproj",
       "Objects\\Objects\\Objects.csproj",
-      "Objects\\Tests\\Objects.Tests.Unit\\Objects.Tests.Unit.csproj"
+      "Objects\\Tests\\Objects.Tests.Unit\\Objects.Tests.Unit.csproj",
       "ConnectorRevit\\RevitSharedResources2020\\RevitSharedResources2020.csproj",
       "ConnectorRevit\\RevitSharedResources2021\\RevitSharedResources2021.csproj",
       "ConnectorRevit\\RevitSharedResources2022\\RevitSharedResources2022.csproj",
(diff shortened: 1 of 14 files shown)


ConnectorDynamo/ConnectorDynamo.slnf
```diff
@@ -23,7 +23,7 @@
       "Objects\\Converters\\ConverterRevit\\ConverterRevit2024\\ConverterRevit2024.csproj",
       "Objects\\Converters\\ConverterRevit\\ConverterRevitShared\\ConverterRevitShared.shproj",
       "Objects\\Objects\\Objects.csproj",
-      "Objects\\Tests\\Objects.Tests.Unit\\Objects.Tests.Unit.csproj"
```
JR-Morgan, line 26: Damn, I may have made this mistake elsewhere. I'll double check and PR if necessary.

ConnectorDynamo/ConnectorDynamo/CreateStreamNode/CreateStream.cs
```diff
@@ -156,22 +157,10 @@ internal void DoCreateStream()
 
       OnNodeModified(true);
     }
-    catch (Exception ex)
+    catch (Exception ex) when (!ex.IsFatal())
     {
-      //someone improve this pls :)
```
JR-Morgan, line 161: Nice response to Matteo's 4 year old bounty 🤣

ConnectorDynamo/ConnectorDynamoExtension/SpeckleExtension.cs
```diff
@@ -35,7 +35,10 @@ public void Loaded(ViewLoadedParams viewLoadedParams)
 
       Setup.Init(HostApplications.Dynamo.GetVersion(HostAppVersion.vRevit), HostApplications.Dynamo.Slug);
     }
-    catch (Exception e) { }
+    catch (Exception ex) when (!ex.IsFatal())
+    {
+      SpeckleLog.Logger.Error(ex, "Failed to load Speckle extension");
```
JR-Morgan, line 40: Should this become a `Fatal`
Also, is this considered a "Top level" exception handler?
If so, would we want to apply the DUI2 strategy of a normal try catch here? (with no `ex.IsFatal()` check)?

JR-Morgan (approved):

ConnectorDynamo/ConnectorDynamo.slnf
JR-Morgan: Just double checked, the `Objects.Tests.Unit.csproj` is the last item in every other slnf except this one. so only this one was broken. Thanks for fixing.

PR #3147, opened 2024-01-18
Test(core): CNX-8745 - Added unit test for logger

JR-Morgan:
Not exhaustive, just testing the console logging + context properties


AlanRynne (approved):
This looks great! We could be more exhaustive but it is a good start as a reference for breaking behaviour

PR #3151, opened 2024-01-26
feat(dui): adds support for updating stream cards to FE2

teocomi:
When loading stream cards into DUI:

- first tries matching them by user ID and Stream URL
- then by just user ID

This will allow upgrading (and eventually downgrading too) stream cards from FE1 to FE2 on any server.
This is guaranteed by the fact that the User ID is consistent across frontends.


AlanRynne:
We wen't for a different solution regarding Account Manager instead of updating stream cards which is a breaking change. Closing this

PR #3154, opened 2024-01-30
fix(Archicad): Receive slowed down in Archicad 26

jozseflkiss:
## Description & motivation

https://spockle.atlassian.net/browse/CNX-8760

## Changes:

It seems that Archciad 26 has a bug. When e.g. passing `%lf` to `GS::String::SPrintf`, it does not iterate on all the format specifier options available for both double and float, just takes the first one, which in this case is `%f`. After that it compares the sizes: `%f `--> `float` (4 bytes), the passed argument was `double` (8 bytes). This size mismatch creates a log entry in the log file which takes so much time per vertex of a mesh being imported. 
In Archicad 27 this seems to be solved.
Since the string representation of the vertex written into GDL was not and should not be so precise (enough to have 8 digits), the fix is to cast the parameters to `float`-s.

## Checklist:


jozseflkiss:
@AlanRynne Bulk of the changes are formatting, and be able to build the add-on with debug version of Archicad.

AlanRynne (approved):
I'll trust you on the c++ side... we'll give it a good test

PR #3150, opened 2024-01-23
chore(archicad): CNX-8304 fix ide0005 in archicad

JR-Morgan:
Archicad:
- Fixed warning as error violations in Archicad projects

Build/CI:
 - Added desktopbuild false for dotnet publish builds
 - Added all level 1 warnings (Except those that had violations) to our build props  (enforced for all builds)

PR #3157, opened 2024-01-31
Chore(build) Lowered several XML warnings from errors

JR-Morgan:
#3150 made all Level 1 warnings errors that didn't have existing violations.
However, it accidently included a couple that have violations.

Weirdly, it appears some of these analysers don't get run as errors on build anyway, so this was how it slipped past me.
Further investigation required, but for now, this PR means Rider doesn't show squiggly red files despite MSBuild compiling successfully


AlanRynne (approved):
Oh nice! In it goes

PR #3158, opened 2024-01-31
feat(core): CNX-8776 move is frontend2 logic

JR-Morgan:
Moves `IsFrontend2` logic so that it will always be set when calling `GetServerInfo` or `GetUserServerInfo`.

This enables us to rely upon this result anywhere else and just check if the account has `Account.serverInfo.frontend2` set to true or false.

Minor changes in 2 grasshopper components that created non-local.


oguzhankoral (approved):
Seems neat!

AlanRynne (approved):
Tested changes in Grasshopper. It all seems to be working as before 🙌🏼

KatKatKateryna (approved):

JR-Morgan:
Changes included in https://github.com/specklesystems/speckle-sharp/pull/3164

PR #3159, opened 2024-02-01
fix(core): CNX-8831 Fix StreamWrapper.ToString() so that it is not FE1 hardcoded

AlanRynne:
https://spockle.atlassian.net/browse/CNX-8831
Based on Account.serverInfo.frontend2 first, then matching regex to originalInput, then defaulting to FE1

Also added NotSupportedException for federation models


AlanRynne:
Draft to discuss with @JR-Morgan

JR-Morgan (approved):

PR #3164, opened 2024-02-05
Fix (FE2): CNX-8776 get header from front end

oguzhankoral:
I've been following up issue that retrieving x-speckle-header-2 header from serverUrl. Here my findings,
- We were trying to query /graphql for header but actually server doesn't know anything about this header.
- I've changed to way of retrieving it on IsFrontend2Server function by pinging /. @Jedd I'm having PR onto your branch/PR for changes.
- Now some unit tests are failing because local server localhost:3000 doesn't proxy to any frontend to ping and get header from it. This is also a bad sign for mutating ServerInfo from different URL requests.. I've added unit tests for real servers like https://app.speckle.systems/, https://speckle.xyz/. but just in case we need to do proper testing with some alpha version when the time comes


Core/Tests/Speckle.Core.Tests.Integration/Credentials/UserServerInfoTests.cs
```diff
@@ -15,17 +15,37 @@ public async Task Setup()
...
+    Assert.That(result, Is.Not.Null);
     Assert.That(result.frontend2, Is.False);
   }
 
+  /// <remarks>
+  /// We get ServerInfo from "http://localhost:3000/graphql",
+  /// Then we mutate the `frontend2` property of ServerInfo by trying to fetch header from "http://localhost:3000/",
+  /// This is not doable in local server because there is no end-point on this to ping.
+  /// This is a bad sign for mutation.
+  /// </remarks>
```
AlanRynne, line 40: Doesn't the local server do the opposite? i.e. it is the frontend that proxies the backend. So maybe it should be `localhost:8080/graphql` for it to work? 🤔

But just the fact that we're discussing this indicates that relying on this header may not be as "reliable" as we were lead to believe...

AlanRynne (dismissed):
So in essence we're rolling back `isFrontend2` implementation to the way it used to be 🤔 is this correct?

Core/Tests/Speckle.Core.Tests.Integration/Credentials/UserServerInfoTests.cs
oguzhankoral: On local server we don't have proxy for frontend. i.e. we run FE2 on `localhost:8081`. So we can't get this header from same server url which is `localhost:3000` on our integration tests
image

Core/Core/Credentials/AccountManager.cs
```diff
@@ -705,30 +703,42 @@ await response.Content.ReadAsStringAsync().ConfigureAwait(false)
...
+        return false;
+      }
 
-    string? headerValue = values.FirstOrDefault();
+      string? headerValue = values.FirstOrDefault();
 
-    if (!bool.TryParse(headerValue, out bool value))
+      if (!bool.TryParse(headerValue, out bool value))
+      {
+        throw new ArgumentException(
```
JR-Morgan, line 730: We should think of a different Exception type to throw here, since the exception is no longer about the server argument

Core/Core/Credentials/AccountManager.cs
```diff
@@ -705,30 +703,42 @@ await response.Content.ReadAsStringAsync().ConfigureAwait(false)
...
+  /// <exception cref="ArgumentException"><paramref name="server"/> contained FE2 header, but the value was <see langword="null"/>, empty, or not parseable to a <see cref="Boolean"/></exception>
+  private static async Task<bool> IsFrontend2Server(Uri server)
   {
-    const string HEADER = "x-speckle-frontend-2";
-    if (!headers.TryGetValues(HEADER, out IEnumerable<string> values))
+    using var httpClient = Http.GetHttpProxyClient();
+
+    var response = await Http.HttpPing(server).ConfigureAwait(false);
+
+    if (response.EnsureSuccessStatusCode() != null)
```
JR-Morgan, line 717: Why compare the return against null? `EnsureSucessCode` will throw for non-success. I'm not sure I understand the intention here

Core/Core/Credentials/AccountManager.cs
```diff
@@ -705,30 +703,42 @@ await response.Content.ReadAsStringAsync().ConfigureAwait(false)
   }
 
   /// <summary>
-  /// Check the <paramref name="headers"/> for a <c>"x-speckle-frontend-2"</c> <see cref="Boolean"/> value
+  /// Check the <paramref name="server"/> for having a <c>"x-speckle-frontend-2"</c> <see cref="Boolean"/> header value
```
JR-Morgan, line 706: Maybe we can re-word this summary also. We aren't checking the server url.. we're making a request to the server. Same goes for the `returns` and `exception` tag

JR-Morgan (changes requested):

Core/Core/Credentials/AccountManager.cs
AlanRynne: I agree, it should be very clear what is the behaviour of this.

Core/Core/Credentials/AccountManager.cs
JR-Morgan: If we're feeling lazy, we could just do a `SpeckleException` here

Core/Core/Credentials/AccountManager.cs
oguzhankoral: Seems bad, agree. Not sure how I end up with this

JR-Morgan (approved):

AlanRynne (approved):

PR #3169, opened 2024-02-09
CNX-8866 Grasshopper nodes format result URL in Stream/Branch format regardless of the input format

AlanRynne:
https://spockle.atlassian.net/browse/CNX-8866

Fixes how Grasshopper duplicates StreamWrapper's by also propagating the original account.
This ensures that the logic to expose FE2 or FE1 logic gets also preserved after the streamwrapper was validated.

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopperShared/Ops/Operations.VariableInputSendComponent.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopperShared/Ops/Operations.VariableInputSendComponent.cs
@@ -695,7 +695,7 @@ public class NewVariableInputSendComponentWorker : WorkerInstance
               );
               wrapper.CommitId = commitId;
               wrapper.SetAccount(client.Account);
-              
+
               OutputWrappers.Add(wrapper);
             }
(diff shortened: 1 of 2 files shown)


ConnectorGrasshopper/ConnectorGrasshopperShared/Ops/Operations.VariableInputSendComponent.cs
```diff
@@ -689,8 +689,13 @@ public override void DoWork(Action<string, double> ReportProgress, Action Done)
               var commitId = await client.CommitCreate(commitCreateInput, CancellationToken);
 
               var wrapper = new StreamWrapper(
-                $"{client.Account.serverInfo.url}/streams/{((ServerTransport)transport).StreamId}/commits/{commitId}?u={client.Account.userInfo.id}"
+                ((ServerTransport)transport).StreamId,
+                client.Account.userInfo.id,
+                client.Account.serverInfo.url
               );
+              wrapper.CommitId = commitId;
```
clairekuang, line 696: object initialization could be cleaner here, but not important

clairekuang (approved):

PR #3172, opened 2024-02-12
Fix(GH) : account token component not finding correct account

connorivy:
## Description & motivation

Account token component was broken by this PR https://github.com/specklesystems/speckle-sharp/pull/3141 which changed the behavior of accounts

This PR applies the same fix that was used for the other components that depended on the account component.

## Checklist:

### DIFF
--- a/ConnectorGrasshopper/ConnectorGrasshopperShared/Accounts/Accounts.GetAccountToken.cs
+++ b/ConnectorGrasshopper/ConnectorGrasshopperShared/Accounts/Accounts.GetAccountToken.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Drawing;
 using System.Linq;
+using ConnectorGrasshopper.Extras;
 using ConnectorGrasshopper.Properties;
 using Grasshopper;
 using Grasshopper.Kernel;
@@ -46,12 +47,7 @@ public class Accounts_GetAccountToken : GH_SpeckleComponent
 
   protected override void RegisterInputParams(GH_InputParamManager pManager)
   {
-    pManager.AddTextParameter(
-      "Account",
-      "A",
-      "Account to get the auth token from. Expects the `userId`",
-      GH_ParamAccess.item
-    );
+    pManager.AddParameter(new SpeckleAccountParam());
   }
 
   protected override void RegisterOutputParams(GH_OutputParamManager pManager)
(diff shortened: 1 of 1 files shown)


ConnectorGrasshopper/ConnectorGrasshopperShared/Accounts/Accounts.GetAccountToken.cs
```diff
@@ -46,12 +47,7 @@ public Accounts_GetAccountToken()
 
   protected override void RegisterInputParams(GH_InputParamManager pManager)
   {
-    pManager.AddTextParameter(
-      "Account",
-      "A",
-      "Account to get the auth token from. Expects the `userId`",
-      GH_ParamAccess.item
-    );
+    pManager.AddParameter(new SpeckleAccountParam());
```
clairekuang, line 50: This changes the input parameter type, which would be a breaking change for users with GH scripts with this component that are passing in a text parameter. Are we okay with this @AlanRynne ?

clairekuang: Nevermind, looks like the input node was an `Accounts` node which previously passed the user id as the output now passes the `Account` object. Looks good to go ✌️

clairekuang (approved):

AlanRynne (changes requested):
These changes do the trick except for updating both name and description of the account input. Built with this branch, i tested an opening a file with an existing "Get Token From Account" component

Screenshot 2024-02-14 at 09 58 11

While the a newly placed component will look like this:

Screenshot 2024-02-14 at 09 58 15

I'm not sure this is a big enough issue to request changing this.
The solution would involve:
- Leaving the original component intact
- Creating a brand new component with these changes
- Create a class that implements IGH_UpgradeObject to define what is involved in "upgrading" an existing component to the new one.

This would also have the side-effect of forcing the users to upgrade their existing nodes in each file (which is Grasshopper recommended approach) as this will only affect descriptions and names of pre-existing components, but the components **will work as expected**

AlanRynne (approved):
Approving after discussion during standup.

Opened ticket to polish off the UI quirks in my comment https://spockle.atlassian.net/browse/CNX-8923

PR #3177, opened 2024-02-15
Fix(Revit) : getting project info can throw and kills conversion

connorivy:
## Description & motivation

Retrieving project info can throw if the project info object has custom attributes with characters that are not allowed for base property names. This pr fixes that issue.

We may want to consider wrapping this whole operation in a try catch because throwing here will kill the entire send operation

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -1417,4 +1417,17 @@ public partial class ConverterRevit
 
     return headOffset;
   }
+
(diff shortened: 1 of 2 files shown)


AlanRynne (changes requested):
Shouldn't we have a way to clean the name of illegal characters? I'm fairly certain Rhino and ACAD already do this for layer names.

Since project info is "send only" i.e we do nothing with it on receive in Revit, i think it's safe to send a modified property name

connorivy:
made the changes per your comment @AlanRynne. Would be good to get another look at it

AlanRynne (approved):
Much nicer ♥️

PR #3111, opened 2023-12-15
fix(rhino): CNX-8572 Handles ca1031 in Rhino converter files

clairekuang:
## Description & motivation

Handles ca1031 warnings in the rhino converter. Mostly concerns layer handling, and `CollectionToNative` conversions, including a minor fix for missing invalid chars and minor refactoring to remove duplicated local methods.

## Changes:

- Rhino converter

PR #3184, opened 2024-02-20
Fix(rhino): restores previous collection to native method and no longer disposes created layer in connector

clairekuang:
- Restores the previous collection to native method (prior to [exception handling pr](https://github.com/specklesystems/speckle-sharp/pull/3111/files)) 
- no longer disposes created layers in connector utils 

Results when receiving Revit objects testing commit:
image


oguzhankoral (approved):
Reviewed/debugged locally, works fine.

JR-Morgan (approved):

PR #3191, opened 2024-02-21
fix(Archicad): CNX-9031 CNX-9040 Missing elements in Archicad from revit/objects stream and between localized versions of Archicads

jozseflkiss:
## Description & motivation

https://spockle.atlassian.net/browse/CNX-9031
https://spockle.atlassian.net/browse/CNX-9040

## Changes:

1. C# Connector: Improved exception handling around native-to-native conversion from Revit to Archicad.
2. C++ Add-on: Layer attribute search error ignored

## Checklist:

### DIFF
--- a/ConnectorArchicad/ConnectorArchicad/Converters/Converters/BeamConverter.cs
+++ b/ConnectorArchicad/ConnectorArchicad/Converters/Converters/BeamConverter.cs
@@ -42,17 +42,19 @@ public sealed class Beam : IConverter
 
             // upgrade (if not Archicad beam): Objects.BuiltElements.Beam --> Objects.BuiltElements.Archicad.ArchicadBeam
             {
-              var baseLine = (Line)beam.baseLine;
-              var newBeam = new Objects.BuiltElements.Archicad.ArchicadBeam
+              if (beam.baseLine is Line baseLine)
               {
-                id = beam.id,
-                applicationId = beam.applicationId,
-                archicadLevel = Archicad.Converters.Utils.ConvertLevel(beam.level),
-                begC = Utils.ScaleToNative(baseLine.start),
-                endC = Utils.ScaleToNative(baseLine.end)
-              };
-
-              beams.Add(newBeam);
+                beams.Add(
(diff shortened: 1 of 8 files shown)


ConnectorArchicad/ConnectorArchicad/Converters/Converters/BeamConverter.cs
```diff
@@ -42,17 +42,19 @@ CancellationToken token
...
+                  new Objects.BuiltElements.Archicad.ArchicadBeam
+                  {
+                    id = beam.id,
+                    applicationId = beam.applicationId,
+                    archicadLevel = Archicad.Converters.Utils.ConvertLevel(beam.level),
+                    begC = Utils.ScaleToNative(baseLine.start),
+                    endC = Utils.ScaleToNative(baseLine.end)
+                  }
+                );
+              }
```
AlanRynne, line 57: I'm assuming from this that if a `Beam` doesn't have a line (has something else) we will silently fail?
At least something should be added to the report in DUI I guess?

AlanRynne (changes requested):
Flagging this as "Request changes" but I'm not sure if this is within scope, or rather should be another ticket.

It seems like the previous logic (and current one too) will silently fail a conversion if the geometry is not of the expected type.

Does Archicad not use the Report tab in DUI at all? It seems like we should be notifying the user if a beam/floor/etc is skipped for whatever reason

ConnectorArchicad/ConnectorArchicad/Converters/Converters/BeamConverter.cs
jozseflkiss: @AlanRynne You are right about reporting. In these cases we silently fail and the report will be "Unknown". However currently the reporting is implemented on C++ side, so I think this should be another ticket (it is not a regression).

jozseflkiss: Ticket added: https://spockle.atlassian.net/browse/CNX-9060

PR #3193, opened 2024-02-21
fix(DUI2): CNX-9024 bug d ui2 branch keeps re setting to main when opening stream card

JR-Morgan:


BovineOx (approved):

clairekuang (approved):
Added comments noting the branch id/name info so we don't forget next time, and a small tweak to the conditional. Looks good to me!

PR #3200, opened 2024-02-27
feat(DUI2): Cnx 9068 update dui2 to use new fe2 terminology

clairekuang:
## Description & motivation
A button that serves as a global config toggle for fe2 terminology was previously implemented in the home view of DUI2, and used in conjunction with `fe2valueconverter` to toggle terminology in the home view. In the stream view, the account was used to determine the value of `UseFe2` irrespective of the config value.

This pr:
- removes functionality for converting fe2 terminology in both views and viewmodels
- All user-facing strings have been updated to reflect new fe2 terminology


AlanRynne (approved):

DesktopUI2/DesktopUI2/ViewModels/MappingTool/MappingsViewModel.cs
```diff
@@ -293,7 +293,7 @@ private void GetTypesAndLevels(Base model)
         () =>
           Dialogs.ShowMapperDialog(
             "No types available",
-            "The selected stream does not contain any Revit types.\nMake sure to send Project Information > Families & Types from Revit\nusing the latest version of the connector.\n\n👉 And no worries, you can keep using Speckle Mapper with default types!",
+            $"The selected project does not contain any Revit types.\nMake sure to send Project Information > Families & Types from Revit\nusing the latest version of the connector.\n\n👉 And no worries, you can keep using Speckle Mapper with default types!",
```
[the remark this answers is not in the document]
clairekuang: Yes - it's the mapper, which doesn't have access to config

DesktopUI2/DesktopUI2/Utils.cs
```diff
@@ -69,6 +69,22 @@ public static IDialogWindow<DialogResult> SendReceiveDialog(string header, objec
 
 public static class Formatting
 {
+  // User-facing terminology for streams, branches, and commits
+  public static string ReplaceTerminology(bool replace, string text)
```
JR-Morgan, line 73: we could mark this function as pure

DesktopUI2/DesktopUI2/Models/ConfigManager.cs
```diff
@@ -25,12 +25,16 @@ public static Config Load()
...
 
-      if (!string.IsNullOrEmpty(newConfig))
+      Config deserializedConfig = !string.IsNullOrWhiteSpace(newConfig)
+        ? JsonConvert.DeserializeObject<Config>(newConfig)
+        : JsonConvert.DeserializeObject<Config>(oldConfig);
+
+      if (deserializedConfig is not null)
       {
-        return JsonConvert.DeserializeObject<Config>(newConfig);
+        deserializedConfig.UseFe2 = true;
```
JR-Morgan, line 34: Am I correct in saying that `UseFe2` will always be true after this.
No more FE1 terminology.

If that's the case, would we be able to kill the `Fe2TextValueConverter` and `Formatting.ReplaceTerminology`

Maybe not now, but we may want to do this eventually. Are there any advantages to keep it?

clairekuang: This just sets the intial state of the FE2 terminology button to true - users can still toggle the state if they wish. I believe we decided that leaving this option here for now is okay, if using new terminology is the default.

clairekuang: Update to the decision: we'll open a separate ticket to deprecate terminology-switching code

clairekuang: Update #2: I've decided to remove everything because it was about the same amount of work amending the previous pr anyways

DesktopUI2/DesktopUI2/ViewModels/HomeViewModel.cs
```diff
@@ -535,7 +535,7 @@ private void GenerateMenuItems()
         new MenuItemViewModel(LaunchManagerCommand, "Manage accounts in Manager", MaterialIconKind.AccountCog)
       );
 
-      menu.Items.Add(new MenuItemViewModel(RefreshCommand, "Refresh streams & accounts", MaterialIconKind.Refresh));
+      menu.Items.Add(new MenuItemViewModel(RefreshCommand, $"Refresh projects & accounts", MaterialIconKind.Refresh));
```
JR-Morgan, line 538: A few of these strings have an unnecessary $
Perhaps we can quickly remove them before his pr is merged

DesktopUI2/DesktopUI2/Views/Controls/StreamEditControls/Receive.xaml
```diff
@@ -199,7 +195,7 @@
       <StackPanel>
         <TextBlock
           IsVisible="{Binding !Commits.Count}"
-          Text="{Binding UseFe2, Converter={StaticResource Fe2TextValueConverter}, ConverterParameter='This branch has no commits. Commits are generated what data is sent to a Branch in a Stream.&#x0a;&#x0a;Please select a different Branch or Stream.'}"
+          Text="This model has no version. Versions are generated what data is sent to a model in a project.&#x0a;&#x0a;Please select a different model or project."
```
JR-Morgan, line 198: Change in pluralisation. Was this intended?
"This branch has no commit**s**"
"This model has no version" 

(I think either is fine, just want to make sure this change was intentional

JR-Morgan (approved):

PR #3201, opened 2024-02-28
CNX-9043 update server info return type and queries to include new migration property

connorivy:
## Description & motivation

Add new account properties to track if a user's account has been upgraded from an FE1 account

## Changes:

Add movedTo and movedFrom props to ServerInfo object
Modify graphQL queries to include the movedFrom and movedTo info

## Checklist:

### DIFF
--- a/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
+++ b/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
@@ -1,5 +1,4 @@
 using GraphQL;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Threading;
 using GraphQL.Client.Http;
--- a/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
+++ b/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
@@ -40,7 +40,7 @@ public static class GraphQLHttpClientExtensions
       );
     }
 
-    return response.Data.serverInfo.version.Contains("dev")
+    return response.Data.serverInfo.version == "dev"
       ? new System.Version(999, 999, 999)
       : new System.Version(Regex.Replace(response.Data.serverInfo.version, "[-a-zA-Z]+", ""));
   }
--- a/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
+++ b/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
@@ -3,6 +3,7 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Threading;
 using GraphQL.Client.Http;
+using System.Linq;
 
 namespace Speckle.Core.Api.GraphQL;
 
@@ -40,8 +41,17 @@ public static class GraphQLHttpClientExtensions
       );
     }
 
+    if (string.IsNullOrWhiteSpace(response.Data.serverInfo.version))
+    {
+      throw new SpeckleGraphQLException<ServerInfoResponse>(
+        $"Query {nameof(GetServerVersion)} did not provide a valid server version",
+        request,
+        response
+      );
+    }
+
     return response.Data.serverInfo.version == "dev"
       ? new System.Version(999, 999, 999)
-      : new System.Version(Regex.Replace(response.Data.serverInfo.version, "[-a-zA-Z]+", ""));
+      : new System.Version(response.Data.serverInfo.version.Split('-').First());
   }
 }
--- a/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
+++ b/Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
@@ -1,0 +1,47 @@
+using GraphQL;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using System.Threading;
+using GraphQL.Client.Http;
+
+namespace Speckle.Core.Api.GraphQL;
+
+public static class GraphQLHttpClientExtensions
+{
+  /// <summary>
+  /// Gets the version of the current server. Useful for guarding against unsupported api calls on newer or older servers.
+  /// </summary>
+  /// <param name="cancellationToken">[Optional] defaults to an empty cancellation token</param>
+  /// <returns><see cref="Version"/> object excluding any strings (eg "2.7.2-alpha.6995" becomes "2.7.2.6995")</returns>
+  /// <exception cref="SpeckleGraphQLException"></exception>
+  public static async Task<System.Version> GetServerVersion(
+    this GraphQLHttpClient client,
+    CancellationToken cancellationToken = default
+  )
+  {
+    var request = new GraphQLRequest
+    {
+      Query =
+        @"query Server {
+                    serverInfo {
+                        version
+                      }
+                  }"
+    };
+
+    var response = await client.SendQueryAsync<ServerInfoResponse>(request, cancellationToken).ConfigureAwait(false);
+
+    if (response.Errors != null)
+    {
+      throw new SpeckleGraphQLException<ServerInfoResponse>(
+        $"Query {nameof(GetServerVersion)} failed",
+        request,
+        response
+      );
+    }
+
+    return response.Data.serverInfo.version.Contains("dev")
+      ? new System.Version(999, 999, 999)
+      : new System.Version(Regex.Replace(response.Data.serverInfo.version, "[-a-zA-Z]+", ""));
+  }
+}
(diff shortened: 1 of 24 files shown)


Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+  /// <summary>
+  /// Gets the version of the current server. Useful for guarding against unsupported api calls on newer or older servers.
+  /// </summary>
+  /// <param name="cancellationToken">[Optional] defaults to an empty cancellation token</param>
+  /// <returns><see cref="Version"/> object excluding any strings (eg "2.7.2-alpha.6995" becomes "2.7.2.6995")</returns>
+  /// <exception cref="SpeckleGraphQLException"></exception>
+  public static async Task<System.Version> GetServerVersion(
+    this GraphQLHttpClient client,
+    CancellationToken cancellationToken = default
+  )
```
AlanRynne, line 20: Why an extension method?

Core/Core/Credentials/AccountManager.cs
```diff
@@ -53,8 +54,23 @@ public static async Task<ServerInfo> GetServerInfo(Uri server, CancellationToken
       httpClient
     );
 
+    System.Version version = await gqlClient
+      .GetServerVersion(cancellationToken: cancellationToken)
+      .ConfigureAwait(false);
+    System.Version serverMigrationVersion = new(2, 18, 5);
```
AlanRynne, line 60: These are some magical numbers, I wish there was a better way to specify this. If not, we should add a comment saying why 2.18.5 is the threshold (we know now... but whoever comes fix this in the future may not)

Core/Core/Credentials/AccountManager.cs
```diff
@@ -53,8 +54,23 @@ public static async Task<ServerInfo> GetServerInfo(Uri server, CancellationToken
...
+    string queryString;
+    if (version >= serverMigrationVersion)
+    {
+      queryString = "query { serverInfo { name company migration { movedFrom movedTo } } }";
+    }
+    else
+    {
+      queryString = "query { serverInfo { name company } }";
+    }
+
```
AlanRynne, line 71: This is for backwards compatibility only, we need to discuss with @BovineOx how long should this stay.

Core/Core/Credentials/AccountManager.cs
```diff
@@ -195,6 +223,25 @@ public static string GetDefaultServerUrl()
...
+    if (!account.serverInfo.url.Contains("https://speckle.xyz"))
+    {
+      throw new SpeckleAccountManagerException($"Can only upgrade accounts on speckle.xyz");
+    }
+
+    account.serverInfo.url = DEFAULT_SERVER_URL;
+    s_accountStorage.UpdateObject(account.id, JsonConvert.SerializeObject(account));
+  }
+
```
AlanRynne, line 244: This implementation is **not correct** ⚠️.

We should be using `migration.migratedTo` to determine if an account "can be upgraded" and if so, use said url as the one to swap.

Also, this is out of scope for this ticket, or did I miss something in the past days while i was lost in Digicert? 🙂

Core/Core/Helpers/Http.cs
```diff
@@ -134,7 +134,7 @@ public static async Task<bool> Ping(string hostnameOrAddress)
   /// </summary>
   /// <param name="uri">The URI that should be pinged</param>
   /// <exception cref="HttpRequestException">Request to <paramref name="uri"/> failed</exception>
-  internal static async Task<HttpResponseMessage> HttpPing(Uri uri)
+  public static async Task<HttpResponseMessage> HttpPing(Uri uri)
```
AlanRynne, line 137: 🤔 Why are we making this public?

Core/Core/Api/GraphQL/Models.cs
```diff
@@ -396,6 +396,23 @@ public class ServerInfo
...
+  /// One of the base fields for our account upgrade mechanism.
+  /// Shows the url that this account now lives at after upgrading to FE2
+  /// </summary>
+  public Uri movedTo { get; set; }
+
+  /// <summary>
+  /// One of the base fields for our account upgrade mechanism.
+  /// Shows the url that this account used to live at before being upgraded to FE2
+  /// </summary>
+  public Uri movedFrom { get; set; }
```
AlanRynne, line 415: This docs are deceiving. `ServerMigration` class knows nothing of any Account, it's purely server metadata.

`movedFrom` and `movedTo` show the "new URL where this server is now deployed" and "the old RUL where this server used to be deployed" respectively.

Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+      throw new SpeckleGraphQLException<ServerInfoResponse>(
+        $"Query {nameof(GetServerVersion)} failed",
+        request,
+        response
+      );
+    }
+
+    return response.Data.serverInfo.version.Contains("dev")
+      ? new System.Version(999, 999, 999)
+      : new System.Version(Regex.Replace(response.Data.serverInfo.version, "[-a-zA-Z]+", ""));
```
AlanRynne, line 45: I think you need to run me through this 🤔 If we're checking if it contains `dev`, why are we then removing all A-Z characters with regex.

Do we know if `dev` is the only possible edge-case?

AlanRynne (changes requested):
I've got many questions hehe 😅🙌🏼

Core/Core/Credentials/AccountManager.cs
JR-Morgan: I think this might just be matteo original implementation, Connors pr is still draft, so I imagine there's more to come.

Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
connorivy: I like extensions methods. They are reusable, discoverable with intellisense, and organized so we'll never end up with massive utils files. I didn't want to add it as a method on line 850 of AccountManager so this is what I came up with, but I'm happy to change it if you have something else in mind.

Core/Core/Credentials/AccountManager.cs
connorivy: this is indeed only for backwards compatibility

Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
connorivy: This logic was taken from Client.ServerOperations.cs

Unfortunately, I could just use that implementation because it only applies to a Speckle.Core.Api.Client which is coupled to our Account class. Where this method is being used currently is a static context that is ignorant of any accounts so we can't use the existing logic.

Core/Core/Api/GraphQL/Models.cs
connorivy: You're right this is misleading

Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
AlanRynne: yeah i do agree 850 lines seems like enough. But extension methods were really created to extend a class you don't have access to (i.e from a dependency), this seems a bit of a misuse, but I accept your reasoning :)

Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
AlanRynne: So you're saying we now have this weird logic in 2 places? 😅

I think we should check with someone from the server team what are the edge-cases expected in the serverInfo.version return value.

Core/Core/Credentials/AccountManager.cs
AlanRynne: I'm opening a ticket for this and letting Ian know

Core/Core/Api/GraphQL/GraphQLHttpClientExtensions.cs
JR-Morgan: Can we have a quick chat on this today? It would be nice if we could avoid duplicating this logic

JR-Morgan: From our discussion with @didimitrie and @gjedlicska :
 - The other server version check can be killed, was only for 2.6.0 backwards compact
 - The `.Contains("Dev")` can be replaced by a `== "Dev"`

There's a couple other topics I want chat about next week in Prague r.e. our `Client` class. But for now, just the above ^ affects this PR.

AlanRynne (approved):

PR #3204, opened 2024-03-01
Fix(Revit) : [CNX-8711] initialize aggregate cache when document is added

connorivy:
## Description & motivation

Devs using our nugets have been facing this issue for a while where the converter will fail due to a null "revitDocumentAggregateCache". This PR initializes the aggregate cache with an empty one when the document is passed into the converter. It still isn't the prettiest solution, but it seems pretty good until we can get a di solution to work.

## Changes:

- move document cache and converted objects cache to the revit shared resources folder
- initialize empty objects cache in converter by default
- create empty document cache when document is passed to "SetContextDocument" method if no document cache has been initialized yet

## Validation of changes:

- tested sending and receiving still works as it did

## Checklist:


AlanRynne (changes requested):
This makes some changes that break the pattern separation. `Converter` shouldn't be aware of the specific type we expect as a `IRevitDocumentCache` (or any other interface for that matter).

Granted, the pattern was already broken when we started using `SetContextDocument` for other things, but not sure if initialising a blank cache makes much sense either 🤔

It also seems that we're using `RevitSharedResources` as a dumping ground of things to avoid the Converter/Connector separation, and that's not really what it was intended either 🙇🏼‍♂️

Since we're going to re-engineer converters for DUI3/DI efforts, we should consider not replicating this pattern

AlanRynne:
Closing this as we decided in yesterday's standup it is not a pressing matter

didimitrie:
PS: I've copy pasted the relevant parts in dui3/dev some time ago :)

PR #3210, opened 2024-03-07
CNX-8975: mixpanel metrics for grasshopper and dynamo are incorrectly logged as rhino revit

JR-Morgan:


AlanRynne (approved):
In addition to the necessary changes in Rhino/Grasshopper and Revit/Dynamo, I have fixed a disposal bug on the Analytics class. 

Disposal and async calls were running into a race condition. I also stopped using the proxy HTTP client and instead used a generic one, logging performance metrics for analytics calls is overkill.

Core/Core/Logging/Analytics.cs
```diff
@@ -221,17 +221,20 @@ public enum Events
...
 
         string json = JsonConvert.SerializeObject(new { @event = eventName.ToString(), properties });
 
         var query = new StreamContent(new MemoryStream(Encoding.UTF8.GetBytes("data=" + HttpUtility.UrlEncode(json))));
 
-        using HttpClient client = Http.GetHttpProxyClient();
+        using HttpClient client = new();
         client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("text/plain"));
         query.Headers.ContentType = new MediaTypeHeaderValue("application/json");
-        client.PostAsync(MIXPANEL_SERVER + "/track?ip=1", query);
```
JR-Morgan, line 234: There's a similar call to PostAsync in the AddConnectorToProfile.

I imagine this one will need the same change.

Core/Core/Logging/Analytics.cs
```diff
@@ -221,17 +221,20 @@ public enum Events
...
+        var res = await client.PostAsync(MIXPANEL_SERVER + "/track?ip=1", query).ConfigureAwait(false);
```
JR-Morgan, line 237: If we aren't using the variable `res`, we should omit it.

Is there any value in calling `EnsureSuccessStatusCode`, it would be nice to know in our logs that there was a failure case, especially if there is no logging performed by the client.

PR #3211, opened 2024-03-08
bug(revit): [CNX-7509] revit wall by face doesnt work in revit 2022

clairekuang:
## Description & motivation

We switched to using a new facewall method for 2023, but this method throws a generic exception when called in 2022.
Reverting back to the old method for 2022 and 2021.

## Changes:

- revit converter


AlanRynne (approved):
Reviewed IRL with Claire.

PR #3212, opened 2024-03-09
MVP: check for conflicts with other Revit plugins - DO NOT MERGE

teocomi:
Sample code to check for dll conflicts when Speckle is launched for the first time.
This code could also live in a completely separate addin that is shipped with our Connector if we wanted to keep things clean and ensure it can run every time.

We might want to subscribe to dll loaded events to ensure lazy-loaded references of other addin are checked too.

Just a draft on top of which to build a proper solution.


AlanRynne:
@connorivy is implementing this elsewhere, closing this PR for now but we can come back to it if needed.

PR #3216, opened 2024-03-12
Cnx 8920 use account migration when selecting accounts in sdk

BovineOx:
- Ensuring accounts migrated to a new server are returned first
- Ensuring we have no duplicate accounts (by ID)
- Unit test added from Jedd's branch (and reviewed)

### DIFF
--- a/Core/Core/Credentials/AccountManager.cs
+++ b/Core/Core/Credentials/AccountManager.cs
@@ -266,7 +266,22 @@ public static class AccountManager
   /// <returns></returns>
   public static IEnumerable<Account> GetAccounts(string serverUrl)
   {
-    return GetAccounts().Where(acc => acc.serverInfo.url == serverUrl);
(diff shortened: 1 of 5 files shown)


JR-Morgan (approved):

Core/Core/Credentials/AccountManager.cs
```diff
@@ -266,7 +266,26 @@ public static void UpgradeAccount(string id)
   /// <returns></returns>
```
JR-Morgan, line 266: Maybe we should update this summary to reflect these changes. Since we no longer "Get all accounts for a given server."

JR-Morgan (changes requested):

Core/Core/Credentials/AccountManager.cs
BovineOx: agree!

JR-Morgan (approved):

PR #3221, opened 2024-03-13
Feat(Core) : Remove account from db and save new one

connorivy:
## Description & motivation

Because the account id is changing, we no longer need to update the object. Instead, we delete the old object from the db and then save a new one.

## Checklist:

### DIFF
--- a/Core/Core/Credentials/AccountManager.cs
+++ b/Core/Core/Credentials/AccountManager.cs
@@ -259,7 +259,8 @@ public static class AccountManager
     // setting the id to null will force it to be recreated
     account.id = null;
 
-    s_accountStorage.UpdateObject(account.id!, JsonConvert.SerializeObject(account));
+    RemoveAccount(id);
+    s_accountStorage.SaveObjectSync(account.id, JsonConvert.SerializeObject(account));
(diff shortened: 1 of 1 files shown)


Core/Core/Credentials/AccountManager.cs
```diff
@@ -259,7 +259,9 @@ public static void UpgradeAccount(string id)
     // setting the id to null will force it to be recreated
     account.id = null;
 
-    s_accountStorage.UpdateObject(account.id!, JsonConvert.SerializeObject(account));
+    RemoveAccount(id);
+    s_accountStorage.SaveObject(account.id, JsonConvert.SerializeObject(account));
+    await s_accountStorage.WriteComplete().ConfigureAwait(false);
```
BovineOx, line 264: Seems fine, just unsure about this change to async and why WriteComplete() is necessary, seems a bit janky to me, but that's probably more about the storage?

BovineOx (approved):
I think it is fine but would like @AlanRynne or @JR-Morgan to cast an eye before merging through

JR-Morgan (approved):

PR #3231, opened 2024-03-15
Restructured DUI3 projects + enforced warnings for new projects

JR-Morgan:
Restructured repo slightly, Were going to keep the CNX folder for now (maybe renamed)

1. Enabled `TreatWarningsAsErrors` by default for all projects in the CNX folder
2. Resolved some warnings in existing projects. But the following projects still have warnings
  `Speckle.Objects.Common` <- A lot of legacy stuff + property naming that could be suppressed
  `Speckle.Connectors.DUI` <- Mostly easy to clean stuff + nullability warnings,
  `Speckle.Connectors.Revit2023` <- mostly unused variables, I'll assume Ian's still in the experimentaiton phase
3. Upgraded the Revit `Speckle.Connectors.Revit2023` to the newer SDK style csporj (actually this was @AlanRynne)
4. Fixed many of the easy to fix warnings and PlatformTarget mishap


CNX-8892/Sdk/Speckle.Autofac/DependencyInjection/AutofacContainer.cs
```diff
@@ -13,10 +13,8 @@ namespace Speckle.Autofac.DependencyInjection;
 // POC: wrap the IContainer or expose it?
 public class AutofacContainer
 {
-  public delegate void PreBuildEventHandler(object sender, ContainerBuilder containerBuilder);
-
   // Declare the event.
-  public event PreBuildEventHandler PreBuildEvent;
+  public event EventHandler<ContainerBuilder> PreBuildEvent;
```
JR-Morgan, line 17: This was another interesting change to reduce boiler plate @BovineOx 
Not sure what you think about this, would be happy to revert if prefered.

CNX-8892/Sdk/Speckle.Autofac/DependencyInjection/IFactory.cs
```diff
@@ -5,8 +5,8 @@ namespace Speckle.Autofac.DependencyInjection;
 // Interceptors?
 
 // POC: this might be somehting that could go in a wholly converter agnostic project
-public interface IFactory<K, T>
-  where T : class
+public interface IFactory<in TKey, out TValue>
```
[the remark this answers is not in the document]
BovineOx: Fair enough, I grew up with C++ templates so that was a bit lazy :D

CNX-8892/Sdk/Speckle.Autofac/DependencyInjection/AutofacContainer.cs
BovineOx: Right, this looks fine, I confess I think I've used this syntax before but event syntax I haven't used in ages tbh

BovineOx (approved):
Wondered if we should settle on <LangVersion> consistency as well, but we can probably tackle that when we get the shape together, along with global depends

AlanRynne (approved):

PR #3232, opened 2024-03-15
Fix for issue #3222

### LINKED ISSUE #3222: SetRelationship (in CommitObjectBuilder) throws "an item with the same key has already been added" exception
I am currently trying to rebuild the functionality of converting a whole Revit document into speckle objects. 
During the conversion, I got a lot "an item with the same key has already been added" exceptions. 

Specifically, this code throws the exception:

https://github.com/specklesystems/speckle-sharp/blob/9a3e68930367949c15a77370f5c4f83582f05634/Core/Core/Models/CommitObjectBuilder.cs#L74-L97

The logic seems flawed. If the `Converted` dictionary does **not** contain the key, a value will be assigned for this key, else the key and the value are added - shouldn't it be the other way round?
So instead of:
```if (!Converted.ContainsKey(appId))```
the if statement would make more sense to me like this:
```if (Converted.ContainsKey(appId))```

Same for `_nestingInstructions`. 

Let me know if you need more context or more information.

andreaslennartz:
As discussed in https://github.com/specklesystems/speckle-sharp/issues/3222


JR-Morgan:
Thanks 🙌,

I want to discuss this change a little with @connorivy (who's currently on holiday, so it will be next week we get round to reviewing it unfortunately).

andreaslennartz:
Sure, no rush. Thanks for looking into this!

JR-Morgan (approved):

connorivy (approved):

JR-Morgan:
Hi @andreaslennartz
We've merged this into our `dev` branch, the change will be part of our 2.19 nugets (I don't have a release date for you just yet, it will be a at least a couple weeks away)

Due to the way we merge commits to `dev` first (with a squash merge), you're fork's `main` branch is now out of sync with ours (just as a heads up if you ever wanted to contribute another pr in future, you may want to reset your `main` branch to avoid conflicts)

andreaslennartz:
Hi @JR-Morgan ,
thanks for the merge and the heads-up!
Keep up the good work!

PR #3238, opened 2024-03-19
feat(rhino) : CNX-9145 DI and creation of send operation services

connorivy:
## Description & motivation

Add proper di to send binding and move send operation out of binding into it's own class which is decoupled from DUI3 implementation

## Changes:

- change ISendBinding interface to return a Task instead of void
- refactor RhinoSendBinding to receive dependencies through constructor
- remove send operation functionality from RhinoSendBinding because it is not testable. Moved into separate send operation class
- split send operation into two parts: 
  - RootBaseObjectBuilder which queries the document and converts the selected objects
  - BaseObjectSenderToServer which deals with serializing the base object and sending to the server
    - This happens through an IBaseObjectSender interface so we can mock the send process during testing
- Most of the conversion logic (and commented code) was copied from existing code)

## Checklist:


DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/SendOperation.cs
```diff
@@ -0,0 +1,40 @@
...
+  {
+    Base commitObject = _baseBuilder.Build(sendFilter, onOperationProgressed, ct);
+
+    // base object handler is separated so we can do some testing on non-production databases
+    // exact interface may want to be tweaked when we implement this
+    string versionId = await _baseObjectSender
+      .Send(commitObject, accountId, projectId, modelId, onOperationProgressed, ct)
+      .ConfigureAwait(false);
+
+    onVersionIdCreated?.Invoke(versionId);
```
AlanRynne, line 52: Is there a reason why we're not returning the versionId directly, as it seems to be the result of this operation.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
```diff
@@ -50,14 +51,16 @@ public List<ISendFilter> GetSendFilters()
...
+  public Task Send(string modelCardId)
   {
     SpeckleTopLevelExceptionHandler.Run(
       () => HandleSend(modelCardId),
       HandleSpeckleException,
       HandleUnexpectedException,
       HandleFatalException
     );
+
+    return Task.CompletedTask;
```
AlanRynne, line 63: 🤔 Although `async void` should never exist, I would have expected `SpeckleTopLevelExceptionHandler.Run` to be awaited or return a task instead of us just returning `Task.CompletedTask`.

We're basically running an async operation but not awaiting it and immediately saying it completed. This is still boilerplate in Revit but we should be aware that the toplevel exception handler should have both `Run` and `RunAsync` available.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Bindings/RhinoSendBinding.cs
```diff
@@ -49,29 +38,32 @@ public class RhinoSendBinding : ISendBinding, ICancelable
...
     DocumentModelStore store,
     RhinoIdleManager idleManager,
-    RhinoSettings rhinoSettings,
     IBridge parent,
     IBasicConnectorBinding basicConnectorBinding,
-    IScopedFactory<ISpeckleConverterToSpeckle> speckleConverterToSpeckleFactory,
-    RhinoContext rhinoContext
+    IEnumerable<ISendFilter> sendFilters,
+    Func<SendOperation> sendOperationFactory,
+    SendBindingUICommands.Factory sendBindingFactory,
```
AlanRynne, line 50: Do we need this to be a factory?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
```diff
@@ -47,11 +51,35 @@ protected override void Load(ContainerBuilder builder)
     builder.RegisterType<RhinoSelectionBinding>().As<IBinding>().SingleInstance();
     builder.RegisterType<RhinoSendBinding>().As<IBinding>().SingleInstance();
 
+    // binding dependencies
+    builder.RegisterType<CancellationManager>().InstancePerDependency();
+    builder.RegisterType<SendBindingUICommands>().InstancePerDependency();
```
AlanRynne, line 56: I'm not sure we need to register this type directly. Maybe we should interface it out so it can be mocked 🤔 Binding commands and bindings are tightly coupled by definition anyway: if the SendBinding is not registered, using the `SendBindingUICommands` will do nothing at best.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/BaseObjectSenderToServer.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    onOperationProgressed?.Invoke("Uploading...", null);
+
+    // TODO: FETCHING ACCOUNTS BY ID ONLY IS UNSAFE
+    Account account =
+      AccountManager.GetAccounts().FirstOrDefault(acc => acc.id == accountId)
+      ?? throw new SpeckleAccountManagerException();
+
+    ITransport transport = _transportFactory(account, projectId);
+    var sendResult = await SendHelper.Send(commitObject, transport, true, null, ct).ConfigureAwait(true);
```
AlanRynne, line 45: Why are we `ConfigureAwait(true)` here? 🤔 Default is false and should be ok in this context.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
```diff
@@ -0,0 +1,154 @@
...
+namespace Speckle.Connectors.Rhino7.Operations.Send;
+
+/// <summary>
+/// Stateless builder object to turn an <see cref="ISendFilter"/> into a <see cref="Base"/> object
+/// </summary>
+public class RootBaseObjectBuilder
+{
+  private readonly Func<ISpeckleConverter> _converterFactory;
+
+  public RootBaseObjectBuilder(Func<ISpeckleConverter> converterFactory)
```
AlanRynne, line 20: Do we need a factory here? Can't we just resolve the ISpeckleConverter directly?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
```diff
@@ -0,0 +1,154 @@
...
+    converter.SetContextDocument(RhinoDoc.ActiveDoc);
+    Base commitObject = ConvertObjects(rhinoObjects, converter, onOperationProgressed, ct);
+
+    return commitObject;
+  }
+
+  private static Collection ConvertObjects(
+    List<RhinoObject> rhinoObjects,
+    ISpeckleConverter converter,
+    Action<string, double?>? onOperationProgressed = null,
```
AlanRynne, line 53: We should re-evaluate how we report "onOperationProgressed". Is that `double` supposed to be `0 to 1`, `0 to 100`? It was never very clear

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/SendBindingUICommands.cs
```diff
@@ -12,6 +12,8 @@ public class SendBindingUICommands
 
   private IBridge _bridge;
 
+  public delegate SendBindingUICommands Factory(IBridge bridge);
```
AlanRynne, line 15: Not sure if we need this 🤔

AlanRynne (commented):
These are mostly open questions that I'm aware we may not have the answers for yet 👍🏻

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
connorivy: This does need to be a factory here with the current converter. This is because the rootBaseObjectBuilder is designed to be a stateless singleton that is reused for all send operations. Therefore, it can't reuse the same converter instance because our converter is full of state.

If the new converter is being designed to be stateless then we could obviously change this

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
connorivy: I just used the existing implementation for this. The progress reporting is actually not working on dui3/alpha at the moment so someone will have to look at this

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/SendBindingUICommands.cs
connorivy: I am currently using this in the RhinoSendBinding constructor to create the sendBindingUiCommands with the same instance of the IBridge as the SendBinding. We could remove it one of two ways,
1. Inject Func<IBridge, SendBindingUiCommands> instead of this factory
2. Create a new container scope for each IBinding and then register the IBridge as scoped.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
connorivy: We would only need to interface out this type if we want to test the RhinoSendBinding class. In my opinion this isn't necessary (or at least it isn't urgent) because this class contains none of the Send logic. Also, it has a bunch of DUI3 and Rhino dependencies that would be tedious to abstract out like RhinoIdleManager and RhinoDocument events

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Bindings/RhinoSendBinding.cs
connorivy: You were right, send operation did not need to be a factory. However, dependencies down the line do kind of need to be factories. What would you use instead of Func?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/BaseObjectSenderToServer.cs
connorivy: good catch. This was definitely autocomplete using the wrong bool and me not checking.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/SendOperation.cs
```diff
@@ -0,0 +1,52 @@
...
+    string modelId,
+    Action<string, double?>? onOperationProgressed = null,
+    CancellationToken ct = default
+  )
+  {
+    Base commitObject = _baseBuilder.Build(sendFilter, onOperationProgressed, ct);
+
+    // base object handler is separated so we can do some testing on non-production databases
+    // exact interface may want to be tweaked when we implement this
+    string versionId = await _baseObjectSender
```
oguzhankoral, line 46: It would be nice to keep in mind we will not return only `versionId` from seriliazer after improvements on caching `ObjectReference`s. It means that we need to think where we will store these objects as state

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/BaseObjectSenderToServer.cs
```diff
@@ -0,0 +1,78 @@
...
+    string modelId,
+    Action<string, double?>? onOperationProgressed = null,
+    CancellationToken ct = default
+  )
+  {
+    ct.ThrowIfCancellationRequested();
+
+    onOperationProgressed?.Invoke("Uploading...", null);
+
+    // TODO: FETCHING ACCOUNTS BY ID ONLY IS UNSAFE
```
BovineOx, line 39: Why is it unsafe? I think this is giving us the behaviour we want? I think the list of matching IDs is ordered with NWA/migrated accounts coming first?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
```diff
@@ -46,11 +50,35 @@ protected override void Load(ContainerBuilder builder)
...
+
     // Register converter factory
+    //builder
```
BovineOx, line 58: we're meant to be using this :D

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
```diff
@@ -0,0 +1,154 @@
...
+using Speckle.Core.Models;
+using Speckle.Core.Kits;
+using System.Threading;
+
+namespace Speckle.Connectors.Rhino7.Operations.Send;
+
+/// <summary>
+/// Stateless builder object to turn an <see cref="ISendFilter"/> into a <see cref="Base"/> object
+/// </summary>
+public class RootBaseObjectBuilder
```
BovineOx, line 16: I mean I know it returns a base, but I don't find it's descriptive enough as to what it is doing, is this the commit builder, maybe we drop Base and then it is RootObjectBuilder, that might be nicer...

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
```diff
@@ -0,0 +1,154 @@
...
+    ISpeckleConverter converter,
+    Action<string, double?>? onOperationProgressed = null,
+    CancellationToken ct = default
+  )
+  {
+    var rootObjectCollection = new Collection { name = RhinoDoc.ActiveDoc.Name ?? "Unnamed document" };
+    int count = 0;
+
+    Dictionary<int, Collection> layerCollectionCache = new();
+    // TODO: Handle blocks.
```
BovineOx, line 61: Let's get out of the habit of adding more TODOs. Will also suggest to the team but let's use POC: instead as it's easy to distinguish existing to new

BovineOx (changes requested):
I think there are some naming changes and comments to review. Maybe we can discuss some changes or perhaps we can take the branch into mine and I can do some refactoring once you have made a few changes. I'm not sure why you want to the trouble of plumbing in the whole of the ISpeckleConverter, perhaps you didn't appreciate that was not coming across?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Bindings/RhinoSendBinding.cs
```diff
@@ -50,29 +39,32 @@ public sealed class RhinoSendBinding : ISendBinding, ICancelable
...
-    _rhinoContext = rhinoContext;
-    _toSpeckleConverter = _speckleConverterToSpeckleFactory.ResolveScopedInstance();
-
+    _sendFilters = sendFilters.ToList();
+    _sendOperation = sendOperation;
+    _cancellationManager = cancellationManager;
     Parent = parent;
-    Commands = new SendBindingUICommands(parent);
+    Commands = sendBindingFactory(parent);
+    SubscriptToRhinoEvents();
```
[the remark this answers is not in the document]
AlanRynne: Typo on `Subscript` should be `Subscribe`

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/BaseObjectSenderToServer.cs
AlanRynne: This was my comment. After our last changes it is "less unsafe" but still unsafe to assume there will only be 1 account with the same id at any given time.

Since we now ensure the "upgraded account" is the one returned first, this is less of an issue now but still I would very much push for

Filter accounts by server -> then pick by ID

This ensures we 100% pick the right account always.

The problem is that it doesn't seem like the modelCardId holds this data at all

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
AlanRynne: POC comment exists on the SpeckleTopLevelExceptionHandler to add async/await later on.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
AlanRynne: No longer needed after my last PR to this branch. Well get to inject these later on

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RootBaseObjectBuilder.cs
AlanRynne: Fixed! 🙌🏼 The issue was using the commands from the BasicBinding in the SendBinding. Since the BasicBinding.Commands was using a different bridge than the send binding... things were not working as intended.

AlanRynne (approved):
All comments are resolved except one regarding the future change tracking feature. Leaving it unresolved so it can be found if need be

PR #3244, opened 2024-03-21
exporting group parameters from Revit

samberger:
## Description & motivation

Currently the parameters of Revit's model groups aren't exported to Speckle at all. Similar to other built elements, the added code aims to add the parameters property object to the exported speckle element.

[as dicussed in the Speckle Community Forum](https://speckle.community/t/revits-built-in-parameter-all-model-type-name-saves-null/8256/12?u=samberger)

## Changes:

- added function `AddGroupParamsToDict()` to `ConversionUtils.cs`
- added parameters export to `GroupToSpeckle()` in `ConvertGroup.cs`

## To-do before merge:

- Unfortunately, I wasn't able to test the code myself, but most of it was copied from other converters and adapted accordingly for groups. So maybe do an extra double check. Sorry.

## Screenshots:

The exported data for `groups` before:

grafik

There should be an additional `parameters{}` property after.

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -343,8 +343,10 @@ public partial class ConverterRevit
     List<string> exclusions = null
   )
   {
-    
-    if (group == null) { return; }
+    if (group == null)
+    {
+      return;
+    }
 
     exclusions ??= new();
     using var parameters = group.Parameters;
@@ -359,9 +361,7 @@ public partial class ConverterRevit
       var speckleParam = ParameterToSpeckle(param, internalName);
       paramDict[internalName] = speckleParam;
     }
-    
     // no need for checking for type parameters
-    
   }
 
   /// <summary>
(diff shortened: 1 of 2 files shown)


bimgeek:
Hey @samberger ,

Thanks for the PR. I have scheduled a ticket to get this in as part of our next sprint (starting Wednesday).

connorivy:
Hey @samberger,

This functionality makes a lot of sense so it would be good to include it. I have a couple questions though. The "AddGroupParamsToDict" seems (at a glance) to be identitcal to our existing "AddElementParamsToDict" method. What is the reason that this method is needed? And similarly, the code that you added to the group conversion method seems similar to existing code as well.

Typically, we would just add the line "GetAllRevitParamsAndIds()" to the conversion. Does that not work for groups?

samberger:
Hi @connorivy,

Yes, most is copy/pasted to a separate function in order to avoid unintended effects on how it works for other elements, since I'm not really into the rest of the code and couldn't get a feeling for that.

But you are completely right! I had another look, and I just noticed that `Group` actually extends `Element`, so it should work the same way... Makes sense of course, but I guess I just haven't seen it.

That's why I thought I would need a separate function and also copied anything else, which seem to apply for groups, from `GetAllRevitParamsAndIds()` directly into `GroupToSpeckle()` 

Long story short, the PR was more thought of as a proposal (starting to get into the connectors' code), and you're right.. so do it the way you proposed and don't merge. :) I'll already be happy about being able to export the group parameters.

Thanks! :)

connorivy:
@samberger 

Perfect. Thanks for the push on this. I have made a beautiful one-line PR https://github.com/specklesystems/speckle-sharp/pull/3293 that will get included in 2.19

PR #3246, opened 2024-03-25
feat(autocad): CNX-9132 dui3 dx autocad create baseline project and bring up to include di

clairekuang:
## Description & motivation

Adds the Autocad Connector and Converter to DUI3-DX. Includes a few basic geometry conversions.


DUI3-DX.slnf
```diff
@@ -25,6 +29,7 @@
       "DUI3-DX\\Sdk\\Speckle.Connectors.Utils\\Speckle.Connectors.Utils.csproj",
       "DUI3-DX\\Sdk\\Speckle.Converters.Common.DependencyInjection\\Speckle.Converters.Common.DependencyInjection.csproj",
       "DUI3-DX\\Sdk\\Speckle.Converters.Common\\Speckle.Converters.Common.csproj",
+      "DesktopUI3\\DUI3\\DUI3.csproj",
```
JR-Morgan, line 32: This is the old DUI3 project, and shoulden't be included in the DX slnf

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/ISendBinding.cs
```diff
@@ -27,6 +28,9 @@ public interface ISendBinding : IBinding
 public class SenderModelCard : ModelCard
 {
   public ISendFilter SendFilter { get; set; }
+
+  [JsonIgnore]
+  public HashSet<string> ChangedObjectIds { get; set; } = new();
```
JR-Morgan, line 33: What was this needed for?

JR-Morgan (changes requested):
There are several analyser warnings still to be resolved. Please ping me tomorrow and we can solve these

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/ISendBinding.cs
didimitrie: This is needed for change detection on send, and it has been added a few weeks back to dui3/dev, but didn't make its way to dx.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadBasicConnectorBinding.cs
```diff
@@ -0,0 +1,137 @@
...
+{
+  public string Name { get; set; } = "baseBinding";
+  public IBridge Parent { get; }
+
+  private readonly DocumentModelStore _store;
+  private readonly AutocadSettings _settings;
+
+  public BasicConnectorBindingCommands Commands { get; }
+
+  private static Document Doc => Application.DocumentManager.MdiActiveDocument;
```
BovineOx, line 25: This should come from the ConversionContext

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadBasicConnectorBinding.cs
```diff
@@ -0,0 +1,137 @@
...
+
+  public string GetConnectorVersion() => Assembly.GetAssembly(GetType()).GetNameAndVersion().Version;
+
+  public string GetSourceApplicationName() => _settings.HostAppInfo.Slug;
+
+  public string GetSourceApplicationVersion() => _settings.HostAppVersion.ToString();
+
+  public Account[] GetAccounts() => AccountManager.GetAccounts().ToArray();
+
+  public DocumentInfo GetDocumentInfo()
```
BovineOx, line 47: When Doc is being injected in the context, it would be good to decide whether this should stay here or become IAutocadDocumentInfo interface, depending on where it is used.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadBasicConnectorBinding.cs
```diff
@@ -0,0 +1,137 @@
...
+    if (Doc == null)
+    {
+      return;
+    }
+    var objectIds = Array.Empty<ObjectId>();
+
+    var model = _store.GetModelById(modelCardId);
+    if (model == null)
+    {
+      return; // TODO: RECEIVERS
```
BovineOx, line 90: Are these TODOs we inherited? If new, then can we use POC: so we can distinguish between pre-existing gaps and new things. The reasoning being that we want to avoid adding gaps, I'd rather push out an improvement with a CNX ticket link in the comments.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadBasicConnectorBinding.cs
```diff
@@ -0,0 +1,137 @@
...
+
+    if (model is SenderModelCard senderModelCard)
+    {
+      List<(DBObject obj, string applicationId)> dbObjects = Doc.GetObjects(senderModelCard.SendFilter.GetObjectIds());
+
+      objectIds = dbObjects.Select(tuple => tuple.obj.Id).ToArray();
+    }
+
+    // TODO: enable it when IReceiveBindingUICommands enabled!
+    // if (
```
BovineOx, line 101: We should address the commented out code also...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadSelectionBinding.cs
```diff
@@ -0,0 +1,76 @@
...
+    Application.DocumentManager.DocumentActivated += (sender, e) => OnDocumentChanged(e.Document);
+  }
+
+  private void OnDocumentChanged(Document document)
+  {
+    if (document == null)
+    {
+      return;
+    }
+    if (!_visitedDocuments.Contains(document))
```
BovineOx, line 31: I'd really like whitespace between these if statements

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadSelectionBinding.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+      PromptSelectionResult selection = doc.Editor.SelectImplied();
+      if (selection.Status == PromptStatus.OK)
+      {
+        using var tr = doc.TransactionManager.StartTransaction();
+        foreach (SelectedObject obj in selection.Value)
+        {
+          var dbObject = tr.GetObject(obj.ObjectId, OpenMode.ForRead);
+          if (
+            dbObject == null /*|| !dbObject.Visible()*/
```
BovineOx, line 61: is this pre-exsting commenting from copy/paste? Does something need doing here? Does it need a POC: comment?

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/Extensions/EditorExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public static class EditorExtensions
+{
+  public static void Zoom(this Editor editor, Extents3d ext)
+  {
+    if (editor == null)
+    {
+      throw new ArgumentNullException(nameof(editor));
+    }
+    using ViewTableRecord view = editor.GetCurrentView();
```
BovineOx, line 15: Can we get a NL after the if statement

BovineOx: Also, wat is going on with this using statement? isn't view being disposed before we start using it?

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadContext.cs
```diff
@@ -0,0 +1,3 @@
+namespace Speckle.Connectors.Autocad.HostApp;
+
+public class AutocadContext { }
```
BovineOx, line 3: We should probably take this through if this is intended to be part of that ConversionContext pattern, or it may be for ACad it isn't needed in the way the RevitContext is needed?

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentManager.cs
```diff
@@ -0,0 +1,148 @@
...
+{
+  private const string SPECKLE_KEY = "Speckle_DUI3";
+  private const string SPECKLE_MODEL_CARDS_KEY = "Speckle_DUI3_Model_Cards";
+
+  /// <summary>
+  /// Returns all the speckle model cards present in the current document.
+  /// </summary>
+  /// <param name="doc"></param>
+  /// <returns></returns>
+  public string? ReadModelCards(Document doc)
```
BovineOx, line 17: Are we in conrol of this? If doc can be null we should be showing this with nullability, if it cannot then the null check makes no sense...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentManager.cs
```diff
@@ -0,0 +1,148 @@
...
+  {
+    if (doc == null)
+    {
+      return null;
+    }
+
+    using (TransactionContext.StartTransaction(doc))
```
BovineOx, line 24: We really need to get that ConversionContext through so we can remove these statics

BovineOx: This will perhaps require some thought, because I don't see the context in which this AutocadDocumentManager is being used. It does not implement an interface and isn't being injected so we need to consider the relationship with lifetimes - it may be we require named lifetimes...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentModelStore.cs
```diff
@@ -0,0 +1,87 @@
+using Speckle.Connectors.DUI.Models;
+using Speckle.Connectors.DUI.Models.Card;
+using Speckle.Newtonsoft.Json;
+
+namespace Speckle.Connectors.Autocad.HostApp;
+
+public class AutocadDocumentStore : DocumentModelStore
+{
+  private static Document Doc { get; set; }
```
BovineOx, line 9: statics :cry

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentModelStore.cs
```diff
@@ -0,0 +1,87 @@
...
+  private static string s_previousDocName;
```
BovineOx, line 10: why do we need this? possibly it will be apparent below...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadIdleManager.cs
```diff
@@ -0,0 +1,38 @@
+using System.Collections.Concurrent;
+
+namespace Speckle.Connectors.Autocad.HostApp;
+
+public class AutocadIdleManager
+{
+  private readonly ConcurrentDictionary<string, Action> _sCalls = new();
+  private bool _hasSubscribed;
```
BovineOx, line 8: volatile keyword should be added

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadIdleManager.cs
```diff
@@ -0,0 +1,38 @@
...
+  public void SubscribeToIdle(Action action)
+  {
+    _sCalls[action.Method.Name] = action;
+
+    if (_hasSubscribed)
+    {
+      return;
+    }
+
+    _hasSubscribed = true;
```
BovineOx, line 23: this setting and clearing represents a code-level race-condition, but I wonder if it is real world. I think you inherited this pattern from existing DUI3 but we do need to decide whether this is a problem, possibly it is...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/TransactionContext.cs
```diff
@@ -0,0 +1,27 @@
...
+using Autodesk.AutoCAD.DatabaseServices;
+
+namespace Speckle.Connectors.Autocad.HostApp;
+
+public sealed class TransactionContext : IDisposable
+{
+  private DocumentLock? _documentLock;
+  private Transaction? _transaction;
+
+  public static TransactionContext StartTransaction(Document document) => new(document);
```
BovineOx, line 16: as mentioned, we should be able to refactor this to remove these statics...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Plugin/AutocadExtensionApplication.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Speckle.Connectors.Autocad.Plugin;
+
+public class AutocadExtensionApplication : IExtensionApplication
+{
+  public void Initialize() => AppDomain.CurrentDomain.AssemblyResolve += new ResolveEventHandler(OnAssemblyResolve);
+
+  public void Terminate()
+  {
+    // TBD
```
BovineOx, line 13: let's concsistently use POC: so we can find all gaps easily

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConversionContextStack.cs
```diff
@@ -0,0 +1,21 @@
...
+
+// POC: Suppressed naming warning for now, but we should evaluate if we should follow this or disable it.
+[SuppressMessage(
+  "Naming",
+  "CA1711:Identifiers should not have incorrect suffix",
+  Justification = "Name ends in Stack but it is in fact a Stack, just not inheriting from `System.Collections.Stack`"
+)]
+public class AutocadConversionContextStack : ConversionContextStack<Document, UnitsValue>
+{
+  public AutocadConversionContextStack(IHostToSpeckleUnitConverter<UnitsValue> unitConverter)
```
BovineOx, line 15: Ahh so we do have this, but we'd need to look at the resolutuon of other elements relying on the document, because we would hope to be able to inject the appropriate IConversionContextStack and access the Document that way

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConverterToSpeckle.cs
```diff
@@ -0,0 +1,54 @@
...
+      }
+
+      var convertedObject = objectConverter.Convert(dbObject);
+
+      return convertedObject;
+    }
+    catch (SpeckleConversionException e)
+    {
+      Console.WriteLine(e);
+      throw; // Just rethrowing for now, Logs may be needed here.
```
BovineOx, line 51: POC: <-- I keep adding this, but we should assume if there is a gap, i.e. a "we should change this in the future" it should be tagged as POC: so before we say this pattern is good, they can be easily found and either fixed or CNX tickets created to do so

BovineOx (changes requested):
Let's discuss, I think if we can tighten up use of POC: we can accept most of this and I can add tickets (we already have one tbh) to review these and action improvements or turn them  into tickets.

Most of my concern is over having the Document everywhere which is pinning us to statics, but...  as I see we can take this in refactor shortly...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentManager.cs
clairekuang: Doc is not null here, so I've removed the unecessary check, but the method is returning null on a few conditions and a caught exception, so I'm leaving the nullable string as is.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadBasicConnectorBinding.cs
clairekuang: I see that currently we aren't using the doc from the `ConversionContext` in other connector basic bindings, and instead just calling the current active doc instead (which has its own problems, especially in rhino). Would you prefer that for Autocad, I inject the context and use the doc from there, or match the other basic bindings and call the active doc instead?

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadBasicConnectorBinding.cs
didimitrie: That to do is misleading, it's actually done in the dui3/dev (and implemented in the code below)

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/Extensions/EditorExtensions.cs
clairekuang: nope, c#8 doesn't require braces, view is disposed at the end of the scope in which it is declared here

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadSendBinding.cs
```diff
@@ -0,0 +1,293 @@
...
+    _cancellationManager = cancellationManager;
+    _sendFilters = sendFilters.ToList();
+
+    Parent = parent;
+    Commands = new SendBindingUICommands(parent);
+
+    Application.DocumentManager.DocumentActivated += (sender, args) => SubscribeToObjectChanges(args.Document);
+    if (Application.DocumentManager.CurrentDocument != null)
+    {
+      // NOTE: catches the case when autocad just opens up with a blank new doc
```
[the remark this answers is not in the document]
didimitrie: I think that was meant purely as an info for future readers tempted to remove that guard (it's my note)

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/Extensions/EditorExtensions.cs
BovineOx: oh I think this one is clearer with braces, much like single if statement body

BovineOx (approved):
I think the using should have braces, but going to take this and maybe add them later on - braces add clarity

DUI3-DX.slnf
BovineOx: I'll take this @JR-Morgan and we can I can add it to a list of outstanding issues

BovineOx: @clairekuang I think the warnings might need resolving @clairekuang before we can pull it in actually

PR #3248, opened 2024-03-26
Enforced CS8618

JR-Morgan:
Thanks @connorivy for pointing this out
We shouldn't really ever have been `NoWarn`ing [CS8618 - 
Nonnullable reference not initialized](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings#nonnullable-reference-not-initialized)

this PR removes CS8618  as a `NoWarn`  from the global build props
and adds as a `WarningsNotAsErrors` (for CI) for the `Core.csproj`, `Objects.csproj`, and `Speckle.Automate.Sdk.csproj`

In addition to this I have fixed the couple violations in our new DUI3 DX projects, and sure we are using null oblivious (`<nullable>disable</nullable>`) context in all our test projects.


Directory.Build.props
```diff
@@ -19,7 +19,7 @@
     <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
 
     <!-- False if running on CI, will prevent copying of files to local folders -->
-    <IsDesktopBuild Condition="'$(IsDesktopBuild)' == ''">true</IsDesktopBuild>
+    <IsDesktopBuild>false</IsDesktopBuild>
```
connorivy, line 22: was this change meant to be committed?

JR-Morgan: Oh good catch, no this absolutely wasn't meant to be staged (it was merely how I was testing locally) thank you

connorivy (approved):
Looks good. Thanks, Jedd

PR #3249, opened 2024-03-26
Added Directory.packages.props file for DUI3-DX dir

JR-Morgan:


DUI3-DX/Directory.Packages.props
```diff
@@ -0,0 +1,12 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageVersion Include="Serilog.Extensions.Logging" Version="7.0.0" />
```
BovineOx, line 6: That feels like the wrong version but I could be wrong?

BovineOx (changes requested):
Just one query but otherwise fine

DUI3-DX/Directory.Packages.props
JR-Morgan: It was the version we were using in the connector projects

JR-Morgan:
I was hoping that we could get away with mixing centrally managed versions, and ones hardcoded in a csproj file
It looks like this isn't the case (see errors in CI, although, locally the build is successful)

Things like RhinoCommon cannot be a centrally managed version, so I'm not sure how we can navigate this.
Perhaps, if we moved some of the shared code into a common connector project, we'll find we avoid referencing the same dependency in multiple places. (And therefore have no need for central package management)

AlanRynne (commented):
Centrally managed packages are useful for "widely used packages in our repo", i.e. things that we now 100% need to be the same version in every project and using a different one will result in unexpected behaviour.

I think for most `Microsoft.*` and things like logging it makes sense to centralise and lock it down.

For connectors it's tricky since the nuget's need to be different versions for each host app release, but I do still see this having value.

PR #3252, opened 2024-03-27
CNX-9175 Initial ArcGIS sending operation and filters

JR-Morgan:


KatKatKateryna:
CI is failing on "using statement" formatting
image

ConnectorArcGIS/Bindings/BasicConnectorBinding.cs
```diff
@@ -23,7 +23,8 @@ public BasicConnectorBinding(ArcGisDocumentStore store)
 
   public string GetSourceApplicationVersion() => "3";
 
-  public string GetConnectorVersion() => Assembly.GetAssembly(GetType()).GetNameAndVersion().Version;
+  public string GetConnectorVersion() =>
+    typeof(BasicConnectorBinding).Assembly.GetNameAndVersion().Version ?? "No version";
```
AlanRynne, line 27: This is odd. Even development versions of our connectors should have a version set, not having a version.

Was this failing for some reason?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Operations/Send/IRootObjectSender.cs
```diff
@@ -0,0 +1,22 @@
+using Speckle.Core.Models;
+
+namespace Speckle.Connectors.ArcGis.Operations.Send;
+
+//poc: duped from rhino
```
AlanRynne, line 5: If this is copy/pasted from Rhino as-is, then clearly it should live in a common project. @BovineOx any suggestions?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Operations/Send/SendOperation.cs
```diff
@@ -0,0 +1,44 @@
+﻿using ArcGIS.Desktop.Framework.Threading.Tasks;
+using Speckle.Core.Models;
+
+namespace Speckle.Connectors.ArcGis.Operations.Send;
+
+//POC: This file is a copy-paste from Rhino
```
AlanRynne, line 6: Also, if copy/pasted as-is from rhino we should move this to a shared project

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/ArcGISConverterToSpeckle.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace Speckle.Converters.ArcGIS3;
 
+//poc: dupe code from rhino
 public class ArcGISConverterToSpeckle : ISpeckleConverterToSpeckle
```
AlanRynne, line 9: Is this really a dupe from Rhino? Does it have no ArcGIS specifics?

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/PointToSpeckleConverter.cs
```diff
@@ -0,0 +1,15 @@
...
+using Speckle.Converters.Common;
+
+namespace Speckle.Converters.ArcGIS3.Geometry;
+
+[NameAndRankValue(nameof(MapPoint), NameAndRankValueAttribute.SPECKLE_DEFAULT_RANK)]
+internal class PointToSpeckleConverter : IHostObjectToSpeckleConversion, IRawConversion<MapPoint, Point>
+{
+  public Base Convert(object target) => RawConvert((MapPoint)target);
+
+  public Point RawConvert(MapPoint target) => new(target.X, target.Y, target.Z, "m");
```
AlanRynne, line 14: I'm assuming this is still pending having the units passed in as part of the context object.
We should add a POC comment here at least.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/VectorLayerToSpeckleConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+      while (rowCursor.MoveNext())
+      {
+        using (Row row = rowCursor.Current)
+        {
+          var element = _pointElementConverter.RawConvert(row);
+          speckleLayer.elements.Add(element);
+        }
+      }
+    }
```
AlanRynne, line 58: We could simplify this with inline using statements and a `foreach`

Objects/Objects/GIS/LineElement.cs
AlanRynne, line 1: Are these brand new objects? Or have these existed in Python for a while, and we're now just copying them over?

If they're new, I would like us to reconsider some of these.

Objects/Objects/GIS/RasterElement.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+  public int? band_count { get; set; }
+  public List<string>? band_names { get; set; }
+  public float? x_origin { get; set; }
+  public float? y_origin { get; set; }
+  public int? x_size { get; set; }
+  public int? y_size { get; set; }
+  public float? x_resolution { get; set; }
+  public float? y_resolution { get; set; }
+  public List<float?>? noDataValue { get; set; }
```
AlanRynne, line 17: Can a Raster element really have all it's properties be null?

Objects/Objects/GIS/CRS.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public class CRS : Base
+{
+  public string? name { get; set; }
+  public string? authority_id { get; set; }
+  public string? wkt { get; set; }
+  public string? units_native { get; set; }
+  public float? offset_x { get; set; }
+  public float? offset_y { get; set; }
+  public float? rotation { get; set; }
```
AlanRynne, line 13: Can a CRS be considered valid if all it's properties are null? By adding nullability we're saying that it's ok for a CRS to not have a `wkt` (or any other field), is this correct?

Objects/Objects/GIS/PointElement.cs
```diff
@@ -0,0 +1,11 @@
...
+using Objects.Geometry;
+using Speckle.Core.Models;
+
+namespace Objects.GIS;
+
+public class PointElement : Base
+{
+  public List<Point>? geometry { get; set; }
+  public Base? attributes { get; set; }
+}
```
AlanRynne, line 11: This PointElement, NonGeometryElement, LineElement classes make me think what we need is a wrapper class that holds some typed geometry and attributes.

Objects/Objects/GIS/VectorLayer.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Objects.GIS;
+
+public class VectorLayer : Collection
+{
+  public CRS? crs { get; set; }
+  public string? units { get; set; }
+  public Base? attributes { get; set; }
+  public string? geomType { get; set; }
+  public Dictionary<string, object>? renderer { get; set; }
```
AlanRynne, line 12: Same nullability question, can a vector layer have everything be null?

AlanRynne (changes requested):
Setting this as "request changes" mostly because of my questions around the new objects that have been added.

Added some other minor comments about refactoring some common code that has been copy/pasted so that it lives in a shared place.

Objects/Objects/GIS/PointElement.cs
KatKatKateryna: this is addressed in a new branch, I removed all these and created just "GisElement"

Objects/Objects/GIS/VectorLayer.cs
KatKatKateryna: regarding all nullability comments: these are the copies from specklepy, not sure we should specify them differently in core

ConnectorArcGIS/Bindings/BasicConnectorBinding.cs
JR-Morgan: This was a copy from Rhino.

The `version` property is nullable, so we needed to resolve the warning, so I copied what rhino was doing. I would prefer to throw here instead if we expect version to always be set.

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/ArcGISSendBinding.cs
```diff
@@ -87,77 +73,30 @@ public List<CardSetting> GetSendSettings()
...
-          new CommitCreateInput
-          {
-            streamId = modelCard.ProjectId,
-            branchName = modelCard.ModelId,
-            sourceApplication = "ArcGIS",
-            objectId = sendResult.rootObjId
-          },
+      string versionId = await _sendOperation
+        .Execute(
+          //modelCard.SendFilter,
```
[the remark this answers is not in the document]
JR-Morgan: Right now, no because I wanted to unblock kat while I am away, and it wasn't obvious how to get the send filter to work with the arcgis api (which works differently from Rhino). more investigation required on this.

ConnectorArcGIS/Bindings/BasicConnectorBinding.cs
AlanRynne: Fine, I accept for now :)

AlanRynne (approved):

PR #3253, opened 2024-03-27
CNX-9159 POC of Revit DI Converters

connorivy:
## Changes:
- Adds di conversions to wall, beams, and columns (and supporting objects)

### Send Binding
- Took send operation logic out of binding and put into designated send operation class

### Send Operation
- similar to existing operation logic that was in the binding
- starting a new container scope and resolving the RootObjectBuilder from that scope

### Connector
- Changed the way that types are registered. 
  - Registered by generic type in IRawConversion instead of by attribute with name of type

#### Extensions
- Copied existing extension methods from old converter

#### Helpers
- miscellaneous models that aren't exactly conversions
- some of this code is new, and some is existing from the old converter
- we could perhaps organize these things a bit more
  - I think I have several things in helpers that really could (should?) be moved to services

#### Services
- kind of like helpers although they have a more specific role. They perform a specific function during an operation such as scaling or caching while helpers may just be containers for data or state

#### ToSpeckle
- several conversions with a mix of new logic and existing logic
  - might be difficult to sort out what is production ready and what is not. For example, wall conversion is fairly close to done except for a couple edge cases like stacked walls and curtain walls.

#### IRawConversion
- I found it helpful to have an intermediate interface that exists between the fully generic conversion ```TOut RawConvert(TIn target)``` and the type-ignorant ```Base Convert(object? target)```
- This new interface, ```Base ConvertToBase(TIn target)``` helped me when I wanted to resolve a conversion but I didn't remember / care about the output type
  - For example, the curve conversion is registered as ```IRawConversion<DB.Curve, ICurve>```. While programming other conversions, I couldn't remember if this conversion return an ICurve or a Base, and I also didn't really care so I just resolved an ```IRawConversion<DB.Curve>```
  - This interface definitely isn't a necessity and can be removed if we feel like it complicates things instead of simplifies them.

### Global.json
- I had an issue where my IDE couldn't build properly while I had ```<IsDesktopBuild>false</>``` because my .net version of the compiler was higher than what the project was expecting
- This is because my latest .net version is 8.0.2 and the sdk settings were to roll forward to the latest major. 
- This change could be an issue if someone only has .net 8 installed and no version of .net 7

## Checklist:


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterValueExtractor.cs
```diff
@@ -0,0 +1,186 @@
...
+
+    return getParamValue(parameter);
+  }
+
+  public Dictionary<string, Parameter> GetAllRemainingParams(DB.Element revitElement)
+  {
+    var allParams = new Dictionary<string, Parameter>();
+    AddElementParamsToDict(revitElement, allParams);
+
+    //var elementType = revitElement.Document.GetElement(revitElement.GetTypeId());
```
BovineOx, line 149: Should this be commented out?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/RevitCategories.cs
```diff
@@ -0,0 +1,148 @@
...
+using System.Collections.Generic;
+using Autodesk.Revit.DB;
+
+namespace Speckle.Converters.RevitShared.Helpers;
+
+/// <summary>
+/// Contains predefined categories of revit objects as well as the types that make up the elements
+/// in that category
+/// </summary>
+public static class RevitCategories
```
BovineOx, line 11: Hmm I imagine this is pre-existing but the mass of statics feels a bit messy, we can perhaps look at it later. At the very least we ought to put some whitespace between each static

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/RevitCategoryInfo.cs
```diff
@@ -0,0 +1,66 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Autodesk.Revit.DB;
+
+namespace Speckle.Converters.RevitShared.Helpers;
+
+public class RevitCategoryInfo
```
BovineOx, line 8: Feels like we ought to inject this at some point or it will make unit testing anything that uses this more complicated

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/RevitConversionContextStack.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Speckle.Converters.RevitShared.Helpers;
+
+[System.Diagnostics.CodeAnalysis.SuppressMessage(
+  "Naming",
+  "CA1711:Identifiers should not have incorrect suffix",
+  Justification = "See base class justification"
+)]
+public class RevitConversionContextStack : ConversionContextStack<UI.UIDocument, ForgeTypeId>
+{
+  public double TOLERANCE { get; } = 0.0164042; // 5mm in ft
```
BovineOx, line 13: I feel this should be a constant unless it can change in any way?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
```diff
@@ -1,26 +1,62 @@
...
+    ParameterValueExtractor parameterValueExtractor
+  )
   {
-    _toSpeckle = toSpeckle;
+    _hostObjectConversions = hostObjectConversions;
+    _convertedObjectsCache = convertedObjectsCache;
+    _parameterValueExtractor = parameterValueExtractor;
   }
 
   public Base Convert(object target)
```
BovineOx, line 29: Probably we need some type check here. It feels like we should have some helper `method`
`
Helper.ThrowIfNot(type)
`
So we can add these kinds of checks easily and consistently

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
```diff
@@ -1,26 +1,62 @@
...
+    if (target is Element element)
+    {
+      _convertedObjectsCache.AddConvertedBase(element.UniqueId, result);
+      _parameterValueExtractor.RemoveUniqueId(element.UniqueId);
+    }
+
+    return result;
+  }
+
+  private IHostObjectToSpeckleConversion? RetrieveObjectConversion(Type objectType)
```
BovineOx, line 49: Interesting, we're looking for the specific type and then the subclass hmmm. I wonder if there is more here to exploit/scaffold

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/CachingService.cs
```diff
@@ -0,0 +1,86 @@
+using System;
+using System.Collections.Generic;
+using System.Collections;
+
+namespace Speckle.Converters.RevitShared.Services;
+
+public class CachingService : ICachingService
```
BovineOx, line 7: What is this caching? Is this pre-existing? We have caching in a few places so... just checking where/what

BovineOx: It's also an interesting class and I am curious about the Funcs being added

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ToSpeckleScalingService.cs
```diff
@@ -0,0 +1,29 @@
+using Autodesk.Revit.DB;
+using Speckle.Converters.RevitShared.Helpers;
+
+namespace Speckle.Converters.RevitShared.Services;
+
+public sealed class ToSpeckleScalingService
+{
+  private readonly double _defaultLengthConversionFactor;
```
BovineOx, line 8: How is this being used. I wonder if this scaling factor should be part of the conversion context stack?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/BaseConversionToSpeckle.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections.Generic;
+using Speckle.Converters.Common;
+using Speckle.Converters.Common.Objects;
+using Speckle.Core.Models;
+
+namespace Speckle.Converters.RevitShared.ToSpeckle;
+
+public abstract class BaseConversionToSpeckle<THost, TSpeckle>
```
BovineOx, line 8: Some of this might be worth having as a base generally?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/CurveConversionToSpeckle.cs
```diff
@@ -0,0 +1,34 @@
...
+using Objects;
+
+namespace Speckle.Converters.RevitShared.ToSpeckle;
+
+[NameAndRankValue(nameof(DB.Curve), 0)]
+public class CurveConversionToSpeckle : BaseConversionToSpeckle<DB.Curve, ICurve>
+{
+  private readonly IRawConversion<DB.Line, SOG.Line> _lineConverter;
+
+  //private readonly IRawConversion<DB.Arc, SOG.Line> _arcConverter;
```
BovineOx, line 12: Commented out section?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/ElementConversionToSpeckle.cs
```diff
@@ -0,0 +1,66 @@
...
+    //else if (baseGeometry is Geometry.Line line)
+    //{
+    //  speckleElement["baseLine"] = line;
+    //}
+
+    speckleElement.category = target.Category.Name;
+
+    speckleElement.displayValue = _displayValueExtractor.GetDisplayValue(target);
+
+    //GetHostedElements(speckleElement, target, out notes);
```
BovineOx, line 47: Commented out section

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/FamilyInstanceConversionToSpeckle.cs
```diff
@@ -0,0 +1,143 @@
...
+    _elementConverter = elementConverter;
+    _beamConverter = beamConverter;
+    _columnConverter = columnConverter;
+  }
+
+  public override Base RawConvert(FamilyInstance target)
+  {
+    Base? @base = null;
+
+    ////adaptive components
```
BovineOx, line 31: Commented out section

BovineOx: I wonder if some of these is a bit messy and we need a more specific conversion to avoid all of the code paths

BovineOx (changes requested):
I think some things are just copy and paste over. I think I would accept those but maybe run through and add some POC: comments for now, to capture that we might need to do something.

There's probably some low hanging fruit for things you can change now and I think some helpers could be injected, possibly should be injected.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterValueExtractor.cs
connorivy: In my opinion, yes, but this is how our existing converter does things so you're right. I put the type parameters back.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/RevitConversionContextStack.cs
connorivy: you are correct

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
connorivy: yeah, I'm wondering if we can make conversion "pipelines" or something like that where we could add all the Wall properties like length and height, and then continue passing the object down to more generic conversions that could add more generic info like parameters or display value.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/CachingService.cs
connorivy: yes this is existing. We do have caching in a ton of places. This was my attempt to bring all the caching into one object. It works well (I think) but we still have caching in a lot of places that we haven't moved over.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ToSpeckleScalingService.cs
connorivy: this is being used for speed. I once profiled a very large revit model and found that we were computing this conversion factor over 5 million times. So now we store that value and use it in our default "ScaleToSpeckle" method.

Yeah I should maybe talk to Alan about that. It seems like it kind of overlaps with the unit functionality of the conversion stack.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/CurveConversionToSpeckle.cs
connorivy: those conversion don't exist yet. Was thinking we could uncomment them as we implement them. Or I could just remove the comments, it wouldn't be much work to redo

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/ElementConversionToSpeckle.cs
connorivy: yeah, a lot of existing functionality that I haven't re-implemented

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/AutofacUIModule.cs
```diff
@@ -65,6 +66,8 @@ protected override void Load(ContainerBuilder builder)
     builder.RegisterType<SendSelection>().AsSelf().InstancePerLifetimeScope();
     builder.RegisterType<ToSpeckleConvertedObjectsCache>().AsSelf().InstancePerLifetimeScope();
     builder.RegisterType<RootObjectBuilder>().AsSelf().InstancePerLifetimeScope();
+    builder.RegisterType<ServerTransport>().As<ITransport>().InstancePerDependency();
```
BovineOx, line 69: Does this have state? What requires it? Will it get the appropriate scope?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterObjectAssigner.cs
```diff
@@ -0,0 +1,71 @@
...
+        speckleParam.isTypeParameter = isTypeParameter;
+        paramBase[kv.Key] = speckleParam;
+      }
+      catch (InvalidPropNameException)
+      {
+        //ignore
+      }
+      catch (SpeckleConversionException ex)
+      {
+        SpeckleLog.Logger.Warning(ex, "Error thrown when trying to set property named {propName}", kv.Key);
```
AlanRynne, line 67: Do we want this to silently fail? This seems like it would cause a lot of headaches (missing parameters in objects...).

I think we may want to store these exceptions until we've finished the loop and then throw an `AggregateException` of some sort?

As this is set, the caller of this method will be unable to know if something went wrong or not.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/RevitConversionContextStack.cs
```diff
@@ -0,0 +1,22 @@
...
+  "Naming",
+  "CA1711:Identifiers should not have incorrect suffix",
+  Justification = "See base class justification"
+)]
+public class RevitConversionContextStack : ConversionContextStack<UI.UIDocument, ForgeTypeId>
+{
+  public const double TOLERANCE = 0.0164042; // 5mm in ft
+
+  public RevitConversionContextStack(RevitContext context, IHostToSpeckleUnitConverter<ForgeTypeId> unitConverter)
+    : base(
```
AlanRynne, line 16: Is it ok to throw in this constructor? Since this entire thing will be dependency injected, when would we be expecting the Document to be null, only on Revit startup?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/SendSelection.cs
AlanRynne, line 1: I think I need to look into why we need this instead of the selection filter

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
```diff
@@ -0,0 +1,68 @@
...
+using Autodesk.Revit.DB;
+using Speckle.Converters.RevitShared.Helpers;
+using Speckle.Connectors.Revit.HostApp;
+using System.Linq;
+
+namespace Speckle.Connectors.Revit.Operations.Send;
+
+public class RootObjectBuilder
+{
+  private readonly SendSelection _sendSelection;
```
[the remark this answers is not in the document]
BovineOx: "having too many interfaces" <- there may be cases there we might not need the interface, but for many things, it is cheap to make the interface and have it, rather the build up a huge cost of having to change behavious and make interfaces for a type down the line, probably by that point multiple types.

Concrete types, bind services together - you cannot instantiate one service without another and this creates tight coupling. This tight coupling makes mocking and unit testing hard and makes it harder to replace one thing with another, following the Liskov subst. principal of SOLID. It also fulfils the O in SOLID, because it is open for expansion in this way and the I because it allows for inversion of control - if we have an ISendSelection then how it functions is opaque to the types that depend upon it, provided it returns the data required.

That said, I don't know what's in SendSelection without hunting it down, so it could be that in this case the concrete type isn't as important - I am a bit reluctant to accept that argument as this is being injected and so probably has dependencies of it's own or is hiding something else

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
AlanRynne: In both these cases I would say these exceptions should be thrown by the called method instead of returning null; as returning `null` clearly means "something went wrong", we might as well throw directly.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
AlanRynne: We should look into this idea Connor. I think it's achievable and makes more sense than having to add parameters on every conversion

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/RevitToSpeckleUnitConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+  public RevitToSpeckleUnitConverter()
+  {
+    _unitMapping[DB.UnitTypeId.Millimeters] = Units.Millimeters;
+    _unitMapping[DB.UnitTypeId.Centimeters] = Units.Centimeters;
+    _unitMapping[DB.UnitTypeId.Meters] = Units.Meters;
+    _unitMapping[DB.UnitTypeId.Inches] = Units.Inches;
+    _unitMapping[DB.UnitTypeId.Feet] = Units.Feet;
+  }
+
+  public string ConvertOrThrow(DB.ForgeTypeId hostUnit)
```
AlanRynne, line 21: This should just be `Convert`? If it throws it should just be documented on the XML comments

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ToSpeckleScalingService.cs
AlanRynne: Yep this sounds like a context thing to me :) lets chat later!

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/CurveConversionToSpeckle.cs
AlanRynne: Add POC comments on all these things so we can search for them later 😉

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/SendOperation.cs
```diff
@@ -0,0 +1,73 @@
...
+using Speckle.Core.Credentials;
+using Speckle.Core.Transports;
+using System.Linq;
+using Autofac;
+
+namespace Speckle.Connectors.Revit.Operations.Send;
+
+public sealed class SendOperation
+{
+  private readonly ILifetimeScope _scope;
```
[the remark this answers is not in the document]
BovineOx: Yeah the intention was that the ISpeckleConverterToSpeckle was the Unit of work and any dependency it required was either Singleton in InstancePerLifetimeScope. It may be that in your case RootObjectBuilder is where the unit of work should live and it would be its dependencies that are new for each instantiation of RootObjectBuilder. I don't know why ISpeckleConverterToSpeckle concrete types would have state, but I can see that legacy dependencies for  RootObjectBuilder might do. We can discuss, but I think I have a note to consider the UoW

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/MeshConversionToSpeckle.cs
```diff
@@ -0,0 +1,40 @@
...
+  )
+  {
+    _contextStack = contextStack;
+    _xyzToPointConverter = xyzToPointConverter;
+    _meshDataTriangulator = meshDataTriangulator;
+  }
+
+  public override List<SOG.Mesh> RawConvert(List<DB.Mesh> target)
+  {
+    MeshBuildHelper buildHelper = new();
```
AlanRynne, line 29: Not sure if we should inject this too, or if doing so would be overkill.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/CurveConversionToSpeckle.cs
BovineOx: Let's add a POC: comment here so we can hoover it up later

DUI3-DX/Sdk/Speckle.Connectors.Utils/Speckle.Connectors.Utils.csproj
```diff
@@ -9,4 +9,8 @@
     <ProjectReference Include="..\..\..\Core\Core\Core.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Exceptions\" />
+  </ItemGroup>
```
AlanRynne, line 14: Is this needed?

AlanRynne (commented):
Some minor comments and lots of questions. No blockers though 😉

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ToSpeckleScalingService.cs
BovineOx: Although creating and disposing of elements in short order 5 millions times might be bad. Was it a necessary conversion, or just that we were doing it in many unnecessary places?

BovineOx (commented):
I need to catch up with Alan on whether on how we should tackle this review. One approach might be that Alan and I pull it and Add POC: or CNX ticket comments where we have some changes. Let's be mindful of scope as we move forwards.

PR #3254, opened 2024-04-02
CNX-9182 receive binding for autocad

oguzhankoral:
## Description & motivation

Points are received with this PR by keeping their layers.

## Checklist:


DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/IReceiveBinding.cs
```diff
@@ -18,11 +19,22 @@ public interface IReceiveBinding : IBinding
...
+{
+  public string ModelCardId { get; set; }
+  public ReceiveResult ReceiveResult { get; set; }
+}
+
+public class ReceiveResult
+{
+  public List<string> BakedObjectIds { get; set; }
+  public bool Display { get; set; }
+}
```
AlanRynne, line 40: we should put these in their own files

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConverterToHost.cs
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public object Convert(Base target)
+  {
+    if (target is not Base speckleObject)
+    {
+      throw new NotSupportedException(
+        $"Conversion of {target.GetType().Name} to Host is not supported. Only objects that inherit from Base are."
+      );
+    }
```
AlanRynne, line 31: How would the `target` input not be a `Base`? It already expects it to be `Base` so this check seems redundant.

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConverterToHost.cs
```diff
@@ -0,0 +1,54 @@
...
+      }
+
+      var convertedObject = objectConverter.Convert(speckleObject);
+
+      return convertedObject;
+    }
+    catch (SpeckleConversionException e)
+    {
+      Console.WriteLine(e);
+      throw; // Just rethrowing for now, Logs may be needed here.
```
AlanRynne, line 51: Missing ``POC:``to the start of this comment 🙏🏼

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/Extensions/EntityExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Speckle.Connectors.Autocad.HostApp.Extensions;
+
+public static class EntityExtensions
+{
+  /// <summary>
+  /// Adds an entity to the autocad database model space record
+  /// </summary>
+  /// <param name="entity">Entity to add into database.</param>
+  /// <param name="layer"> Layer to append object.</param>
+  public static ObjectId Append(this Entity entity, string? layer = null)
```
AlanRynne, line 12: This method's logic is slightly odd. We return `ObjectId.Null` but is that an error we're swallowing? Should we prefer throwing an exception here instead?

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
```diff
@@ -0,0 +1,113 @@
...
+using Autodesk.AutoCAD.EditorInput;
+using Autodesk.AutoCAD.LayerManager;
+
+namespace Speckle.Connectors.Autocad.HostApp;
+
+public class AutocadLayerManager
+{
+  private readonly AutocadContext _autocadContext;
+  private readonly string _layerFilterName = "Speckle";
+  private Document Doc => Autodesk.AutoCAD.ApplicationServices.Core.Application.DocumentManager.MdiActiveDocument;
```
AlanRynne, line 11: Will this work if the user changes the Active Document in autocad?

Not sure how this works on Acad, just want to be sure we shouldn't be putting the Doc inside the `AutocadContext` instead.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadContext.cs
AlanRynne, line 1: This should follow the pattern established in Rhino, where the `AutocadContext` holds the Document as the comment suggests, and rely on picking the `Current` context in the context stack.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadReceiveBinding.cs
```diff
@@ -0,0 +1,231 @@
...
+  /// <summary>
+  /// Utility function to flatten a conversion result that might have nested lists of objects.
+  /// This happens, for example, in the case of multiple display value fallbacks for a given object.
+  /// </summary>
+  /// <param name="item"></param>
+  /// <returns></returns>
+  private List<object> FlattenToNativeConversionResult(object item)
+  {
+    List<object> convertedList = new();
+    void Flatten(object item)
```
AlanRynne, line 204: private local functions are somewhat controversial. I don't see why this needs to be a private function. It makes the logic a bit convoluted (and recursive too 😅)

We could flatten the recursive pattern into a simpler implementation using a `while` loop, which would also remove the need for the private function altogether.

Core/Core/Models/Extensions/BaseExtensions.cs
```diff
@@ -226,4 +226,70 @@ public static bool IsDisplayableObject(this Base speckleObject)
   {
     return TryGetParameters<Base>(obj);
   }
+
+  /// <summary>
+  /// A variation of the OG Traversal extension from Alan, but with tracking the object path as well.
+  /// </summary>
+  /// <param name="recursionBreaker"> Delegate condition to stop traverse.</param>
+  /// <returns>List of base objects with their collection path.</returns>
+  public static IEnumerable<(List<string>, Base)> TraverseWithPath(
```
AlanRynne, line 235: We would need some unit tests for this :)

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadReceiveBinding.cs
AlanRynne, line 1: We should look into splitting some of this methods into injected types to simplify the logic of the binding. 300 lines to start with seems like a lot :)

AlanRynne (changes requested):
Some changes, but mostly comments to consider with @BovineOx @oguzhankoral and @clairekuang in a call, as some come from my Acad ignorance.

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/IReceiveBinding.cs
oguzhankoral: Yeah, also moving send related ones.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadReceiveBinding.cs
oguzhankoral: Agree, added also into Core Utilities and unit test for new implementation.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
BovineOx: I'd rather not keep getting the active document from these static locations, it binds us to those locations and injecting it in an IHostApplicationObject<T> or similar feels like it could be a better solution, but we can POC: mark it and take it as

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
```diff
@@ -0,0 +1,113 @@
+using Autodesk.AutoCAD.DatabaseServices;
+using Autodesk.AutoCAD.EditorInput;
+using Autodesk.AutoCAD.LayerManager;
+
+namespace Speckle.Connectors.Autocad.HostApp;
+
+public class AutocadLayerManager
```
BovineOx, line 7: While this has a specific use the binding together of concrete types creates unnecessarily tight coupling, so I'd like to see an interface here, we can take it with POC: comment to that effect however

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
```diff
@@ -0,0 +1,113 @@
...
+  /// <summary>
+  /// Will create a layer with the provided name, or, if it finds an existing one, will "purge" all objects from it.
+  /// This ensures we're creating the new objects we've just received rather than overlaying them.
+  /// </summary>
+  /// <param name="layerName">Name to search layer for purge and create.</param>
+  public void CreateLayerOrPurge(string layerName)
+  {
+    Document doc = Application.DocumentManager.MdiActiveDocument;
+    doc.LockDocument();
+    using Transaction transaction = doc.TransactionManager.StartTransaction();
```
BovineOx, line 42: It would be nice to wrap this in some dispose pattern tbh. And obtain it from a factory perhaps. Though the semantics of dispose vs commit/rollback we should consider.

Either way the actions around creating, committing, etc..  would be nice to hide if at all possible...  

Again we can take with a POC: comment

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
```diff
@@ -0,0 +1,113 @@
...
+    transaction.Commit();
+  }
+
+  /// <summary>
+  /// Creates a layer filter for the just received model, grouped under a top level filter "Speckle". Note: manual close and open of the layer properties panel required (it's an acad thing).
+  /// This comes in handy to quickly access the layers created for this specific model.
+  /// </summary>
+  /// <param name="projectName"></param>
+  /// <param name="modelName"></param>
+  public void CreateLayerFilter(string projectName, string modelName)
```
BovineOx, line 78: On cursory glance we should break this into it's own service and/or factory. Let's add a POC: comment here plz

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/Extensions/EntityExtensions.cs
BovineOx: I'm not keen on this as an extension as it has a dependency on document which is tightly binding us to the static instance

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConverterToHost.cs
```diff
@@ -0,0 +1,54 @@
...
+using Speckle.Converters.Common;
+using Speckle.Converters.Common.Objects;
+using Speckle.Core.Models;
+
+namespace Speckle.Converters.Autocad;
+
+public class AutocadConverterToHost : ISpeckleConverterToHost
+{
+  private readonly IFactory<string, ISpeckleObjectToHostConversion> _toHost;
+  private readonly IHostToSpeckleUnitConverter<UnitsValue> _unitConverter;
```
BovineOx, line 13: Is this used?

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConverterToHost.cs
```diff
@@ -0,0 +1,54 @@
...
+  {
+    _toHost = toHost;
+    _unitConverter = unitConverter;
+  }
+
+  public object Convert(Base target)
+  {
+    if (target is not Base speckleObject)
+    {
+      throw new NotSupportedException(
```
BovineOx: We possible ought to have a POC: comment for deciding on actual exception type

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/DependencyInjection/AutofacAutocadConverterModule.cs
```diff
@@ -15,6 +15,7 @@ protected override void Load(ContainerBuilder builder)
     // should be InstancePerLifetimeScope
     // most things should be InstancePerLifetimeScope so we get one per operation
     builder.RegisterType<AutocadConverterToSpeckle>().As<ISpeckleConverterToSpeckle>().SingleInstance();
+    builder.RegisterType<AutocadConverterToHost>().As<ISpeckleConverterToHost>().SingleInstance();
```
BovineOx, line 18: I suspect the scope should be instance per lifetime

BovineOx (changes requested):
Much of it could be taken with some POC: comments in the spirit of getting this in. There maybe some low-hanging fruit we can address

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/AutocadConverterToHost.cs
clairekuang: No, will remove from this class and also the `AutocadConverterToSpeckle` class as well. Should probably also remove it from Rhino, as it isn't used there as well

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadContext.cs
```diff
@@ -3,4 +3,17 @@ namespace Speckle.Connectors.Autocad.HostApp;
 public class AutocadContext
 {
   // POC: we may want to inject this autocadcontext with the active doc?
```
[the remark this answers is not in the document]
AlanRynne: Agreed, Kateryna found a valid issue with the contexts when changing Documents in ArcGIS that is basically what you just described.

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/DependencyInjection/AutofacAutocadConverterModule.cs
AlanRynne: Most definitely, considering our last conversation around "unit of work" and scoped factories

AlanRynne:
FYI, this branch is missing some commits from `CNX-9175-Send-Filters-Operation` branch that Claire pushed. Not sure if they're critical but we should clarify this.

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/DependencyInjection/AutofacAutocadConverterModule.cs
oguzhankoral: It is just for `AutocadConverterToHost`? or both `AutocadConverterToHost` and `AutocadConverterToSpeckle`

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadReceiveBinding.cs
oguzhankoral: This is done with `IHostObjectBuilder` which iterates commit object and use converter to create each host app object. Now, ReceiveBinding doesn't depend on any converter.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
AlanRynne: POC comments added to address this later with some AutocadContext refactor.

AlanRynne (approved):
Super nice! In we go!

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
```diff
@@ -0,0 +1,60 @@
...
+    _hostObjectBuilder = hostObjectBuilder;
+  }
+
+  public async Task<IEnumerable<string>> Execute(
+    string accountId, // POC: all these string arguments exists in ModelCard but not sure to pass this dependency here, TBD!
+    string projectId,
+    string projectName,
+    string modelName,
+    string versionId,
+    CancellationTokenSource cts,
```
JR-Morgan, line 30: Question: Do we need to be passing the token source around?
the general pattern for cancellation is to keep a reference to the CancellationTokenSource in a single place (the object that "owns" the token source), and to pass only the `CancellationToken` into methods as the last parameter.

I'm curious if we are breaking this pattern intentionally?

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    Base? commitObject =
+      await Speckle.Core.Api.Operations
+        .Receive(version.referencedObject, cancellationToken: cts.Token, remoteTransport: transport)
+        .ConfigureAwait(false)
+      ?? throw new SpeckleException(
+        $"Failed to receive commit: {version.id} objects from server: {nameof(Operations)} returned null"
+      );
+
+    apiClient.Dispose();
```
JR-Morgan, line 54: This type of disposal is not safe. As if an exception occurs between creating the Client, and this line, then the client will not be disposed.

Instead we should be using the `using` keyword on line 40 (where we are constructing the client)

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
JR-Morgan: see https://learn.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads and https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource?view=net-8.0

PR #3255, opened 2024-04-02
fix(Revit): Dockable panel initialization not working when launching Dockable Revit form a file

Mehdi-Lucas:
## Description & motivation

The Dockable Pane wasn't initializing correctly when users opened Revit by double-clicking a file. Previously, the Speckle add-in would handle this by displaying an error message instructing the user to create a new file or open Revit first for a correct pane initialization. While this approach served as a workaround, it proved sometimes cumbersome for users, especially those with automated batch jobs that automatically opened files, as highlighted in a user's (@GianClaudioScarafini ) [comment](https://github.com/specklesystems/speckle-sharp/issues/1469#issuecomment-1812486081) in the linked issue #1469.

In my development of an add-in for my client, which also utilizes a DockablePane, I aimed for a more seamless experience and sought to avoid workarounds. I discovered that the issue arose because the DocumentOpening event was being triggered before the ApplicationInitialized event. Consequently, the DockablePane wasn't registered at that point, leading to its malfunction.

To resolve this, I implemented a solution where the pane registration occurs during both the DocumentOpening and ApplicationInitialized events. However, since two events can call the pane creation now, it's essential to check whether the pane has already been registered and created before proceeding with the registration process during these events. 
 
Fix #1469 
## Changes:

Revit Connector

## To-do before merge:

Please ensure that this modification does not impact the behavior of the connector in other aspects. While my tests suggest it's working, due to time constraints, I couldn't delve deeply into the intricacies of Speckle domain logics. It's prudent to verify that all changes still function as intended.

Particularly, scrutinize the bindings creation process, as I have made adjustments to accommodate their addition during the DocumentOpening events to ensure smooth UI creation. 

And let me know If some clarification is required or if there are any uncertainties.

## Validation of changes:

Manual testing by opening multiple files. Do not hesistate to test on more file/scenarios.
Here is the behavior with the Pane not being created when double clicking a file (before the change):
dockablePaneBefore

Here is the behavior with the Pane being correctly created when double clicking a file (after the change):
dockablePaneAfter

## Checklist:

- [x] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x] My commits are related to the pull request and do not amend unrelated code or documentation.
- [x] My code follows a similar style to existing code.
- [ ] I have added appropriate tests.
[...]

PR #3256, opened 2024-04-03
Fix(logging): profile union and set calls to analytics

teocomi:
The `profile-union` and `profile-set` calls cannot be combined and were failing silently; this fixes that.
I'm pointing to `main` but this can go in any time, also in the next release.


AlanRynne (approved):
Looks good! Let's combine this and the other analytics improvements by Claire in #3255 and make a pre-release to ensure both work right

clairekuang (approved):

PR #3257, opened 2024-04-03
Hotfix(metrics): CNX-9196 add object type event tracking for mixpanel in autocad, rhino, and revit

clairekuang:
## Description & motivation

Adds two new events for mixpanel tracking: `SendObjectReport` and `ReceiveObjectReport`.
Attaches a dictionary containing each Speckle type found in the sent/received commit and their count, and tracks this event separately from send and receive events.

Added for AutoCAD, Rhino, and Revit.

### DIFF
--- a/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
+++ b/ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
@@ -11,6 +11,7 @@ using DesktopUI2;
 using DesktopUI2.Models;
 using DesktopUI2.ViewModels;
 using Speckle.Core.Api;
+using Speckle.Core.Credentials;
 using Speckle.Core.Kits;
 using Speckle.Core.Logging;
 using Speckle.Core.Models;
@@ -110,6 +111,9 @@ public partial class ConnectorBindingsAutocad : ConnectorBindings
         progress.Report = new ProgressReport();
         var conversionProgressDict = new ConcurrentDictionary<string, int>();
         conversionProgressDict["Conversion"] = 0;
+
+        // track object types for mixpanel logging
+        Dictionary<string, int> loggingTypeCountDict = new();
 
         // create a commit prefix: used for layers and block definition names
         var commitPrefix = Formatting.CommitInfo(stream.name, state.BranchName, id);
@@ -217,18 +221,28 @@ public partial class ConnectorBindingsAutocad : ConnectorBindings
             return;
           }
 
-          // convert base (or base fallback values) and store in appobj converted prop
-          if (commitObj.Convertible)
+          if (StoredObjects.TryGetValue(commitObj.OriginalId, out Base commitBaseObj))
           {
-            if (StoredObjects.TryGetValue(commitObj.OriginalId, out Base obj))
+            // log received object type
+            if (loggingTypeCountDict.TryGetValue(commitBaseObj.speckle_type, out int value))
             {
-              converter.Report.Log(commitObj); // Log object so converter can access
-              commitObj.Converted = ConvertObject(obj, converter);
+              loggingTypeCountDict[commitBaseObj.speckle_type] = ++value;
             }
             else
             {
-              commitObj.Update(status: ApplicationObject.State.Failed, logItem: "Object not found in StoredObjects");
+              loggingTypeCountDict.Add(commitBaseObj.speckle_type, 1);
             }
+          }
+          else
+          {
+            commitObj.Update(status: ApplicationObject.State.Failed, logItem: "Object not found in StoredObjects");
+          }
+
+          // convert base (or base fallback values) and store in appobj converted prop
+          if (commitObj.Convertible)
+          {
+            converter.Report.Log(commitObj); // Log object so converter can access
+            commitObj.Converted = ConvertObject(commitBaseObj, converter);
           }
           else
           {
@@ -270,6 +284,14 @@ public partial class ConnectorBindingsAutocad : ConnectorBindings
           progress.Report.Log(commitObj);
         }
         progress.Report.Merge(converter.Report);
+
+        // track the object type counts as an event before we try to send
+        // this will tell us the composition of a commit the user is trying to send, even if it's not successfully sent
+        Analytics.TrackEvent(
+          AccountManager.GetDefaultAccount(),
+          Analytics.Events.ReceiveObjectReport,
+          loggingTypeCountDict.ToDictionary(o => o.Key, o => o.Value as object)
+        );
 
         // add applicationID xdata before bake
         if (!ApplicationIdManager.AddApplicationIdXDataToDoc(Doc, tr))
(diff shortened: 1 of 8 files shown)


AlanRynne (approved):
Codewise, changes look good. We should make a pre-release to tripple check the metrics before going live

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -217,20 +221,30 @@ string id
             return;
           }
 
-          // convert base (or base fallback values) and store in appobj converted prop
-          if (commitObj.Convertible)
+          if (StoredObjects.TryGetValue(commitObj.OriginalId, out Base commitBaseObj))
```
BovineOx, line 224: Is this just giving is things we successfully converted?

Core/Core/Logging/Analytics.cs
```diff
@@ -67,7 +67,17 @@ public enum Events
     /// <summary>
     /// Event triggered by the Mapping Tool
     /// </summary>
-    MappingsAction
+    MappingsAction,
+
+    /// <summary>
+    /// Event triggered on send, capturing the object type counts of the sent commit
+    /// </summary>
+    SendObjectReport,
```
BovineOx, line 75: Can we just make these without setup or do we need to setup on MixPanel side, presumably it's either automatic or we did the setup? Just asking? Also, do we know what part of our estate is release and what part is WIP?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -271,6 +285,14 @@ string id
         }
         progress.Report.Merge(converter.Report);
 
+        // track the object type counts as an event before we try to send
+        // this will tell us the composition of a commit the user is trying to send, even if it's not successfully sent
+        Analytics.TrackEvent(
+          AccountManager.GetDefaultAccount(),
```
teocomi, line 291: Using the default account could skew our reporting, as the person doing the operation could be someone else.

teocomi (changes requested):
I would not introduce two new events to track this; instead, we should attach this information to the existing `Send` and `Receive` events.

Also, I did not see any instance of `SendObjectReport` and `ReceiveObjectReport` in mixpanel. Has this code been tested? What's the payload like?

clairekuang:
> I would not introduce two new events to track this; instead, we should attach this information to the existing `Send` and `Receive` events.

Since we are tracking this event from the connector binding (not DUI2), wouldn't this create 2 send events for every send then? Originally I introduced a new event to avoid this issue, but perhaps I'm misunderstanding how the send event is tracked

clairekuang:
> I would not introduce two new events to track this; instead, we should attach this information to the existing `Send` and `Receive` events.

Since we are tracking this event from the connector binding (not DUI2), wouldn't this create 2 send events for every send then? Originally I introduced a new event to avoid this issue, but perhaps I'm misunderstanding how the send event is tracked

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
clairekuang: No, it's giving the top level Speckle objects in the commit that the user is receiving. Does not record successful conversion or bake.

clairekuang:
Just refactored the payload to be an array instead of a dictionary! Sample Mixpanel report below:
image

JR-Morgan (commented):
Were building up a fair bit of duplicate code.
 I think @alanrynne and @BovineOx were discussing consolidating the send + receive bindings into a shared project (so all connectors use the same) this would likely fix things here.
Not a blocker for this pr, but just extra fuel for us needing to re-consider where this lives.

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -271,6 +284,21 @@ string id
         }
         progress.Report.Merge(converter.Report);
 
+        // track the object type counts as an event before we try to receive
+        // this will tell us the composition of a commit the user is trying to receive, even if it's not successfully received
+        // we are capped at 255 properties for mixpanel events, so we need to check dict entries
+        var typeCountArray = typeCountDict
+          .ToArray()
```
JR-Morgan, line 291: To nitpick a little here. Is this initial `.ToArray()` needed, is it doing anything here?

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -271,6 +284,21 @@ string id
...
+          .Select(o => new { TypeName = o.Key, Count = o.Value })
+          .OrderBy(pair => pair.Count)
+          .Reverse()
+          .Take(250);
```
JR-Morgan, line 295: Perhaps we can encode this value as a constant inside the Analytics class, just so we avoid having magic numbers

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
```diff
@@ -217,20 +220,30 @@ string id
...
+            if (typeCountDict.TryGetValue(commitBaseObj.speckle_type, out int value))
             {
-              converter.Report.Log(commitObj); // Log object so converter can access
-              commitObj.Converted = ConvertObject(obj, converter);
+              typeCountDict[commitBaseObj.speckle_type] = ++value;
             }
             else
             {
-              commitObj.Update(status: ApplicationObject.State.Failed, logItem: "Object not found in StoredObjects");
+              typeCountDict.Add(commitBaseObj.speckle_type, 1);
```
teocomi, line 232: This can be simplified, no need for branching:
typeCountDict.TryGetValue(commitBaseObj.speckle_type, out int value); 
typeCountDict[commitBaseObj.speckle_type] = ++value;

ConnectorAutocadCivil/ConnectorAutocadCivil/UI/ConnectorBindingsAutocadCivil.Receive.cs
teocomi: SInce the limit in mixpanel is 255, and we might be adding some other default props, I would lower it to 230 or so to be safe, maybe even 200.

ConnectorRevit/ConnectorRevit/UI/ConnectorBindingsRevit.Send.cs
```diff
@@ -180,6 +196,21 @@ await APIContext
...
+      .ToArray()
+      .Select(o => new { TypeName = o.Key, Count = o.Value })
+      .OrderBy(pair => pair.Count)
+      .Reverse()
+      .Take(250);
+
+    Analytics.TrackEvent(
+      Analytics.Events.ConvertToSpeckle,
+      new Dictionary<string, object>() { { "typeCount", typeCountArray } }
+    );
```
teocomi, line 212: Wondering if this could go in a helper function, but up to you guys...

teocomi (approved):
Some minor comments

PR #3261, opened 2024-04-05
CNX-8397 improve the user experience of encountering dll conflicts and make the event more actionable

connorivy:
### DllConflictManager
- Detects conflicts between assemblies referenced by the connector and currently loaded assemblies in the app domain
- Handles warning the user of potential conflicts before any exception is thrown
- Handles notifying the user of dll conflicts after exception is throw. Specifically "MissingMethodException" and "TypeLoadException".
  - I think we probably need to add this exception handling in more places. I currently only have it for the loading of the connector when Revit starts up

### DllConflictManagmentOptions
- Holds a list of dll conflicts that the user has elected to ignore
- DllConflictManagmentOptionsLoader class will load these options from a json file in the same directory as the connector dll

### RevitExternalApplication
- Moved all types that are not either connector, System, or Revit types out of the "OnStartup" method.
  - This is because if there are conflicting dll types in this method, then an exception will be thrown before the method is even entered
  - By putting the other types in different methods, the type loading exception will be thrown on entering that method. Therefore we can alert the user about the exception

## Checklist:


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+  {
+    _optionsLoader = optionsLoader;
+  }
+
+  public void DetectConflicts()
+  {
+    Dictionary<string, Assembly> loadedAssembliesDict = new();
+    foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+    {
+      // TODO : what to do about multiple versions of loaded dlls?
```
BovineOx, line 26: We need a solution if we can come to something quick - let's not add a bunch of TODOs unnecessarily/CNX ticket would be better.

Can we make a stronger key that would be unique?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
BovineOx, line 1: I guess this is a Revit only version, but that might be fine for now...  I can't recall what we said in the scope..?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+    Assembly assembly,
+    Dictionary<string, Assembly> loadedAssemblies,
+    HashSet<string> visitedAssemblies
+  )
+  {
+    if (visitedAssemblies.Contains(assembly.GetName().Name))
+    {
+      return;
+    }
+    visitedAssemblies.Add(assembly.GetName().Name);
```
BovineOx, line 42: Line space between end of if clause

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+    LoadAssemblyAndDependencies(typeof(DllConflictManager).Assembly, loadedAssembliesDict, new HashSet<string>());
+  }
+
+  private void LoadAssemblyAndDependencies(
+    Assembly assembly,
+    Dictionary<string, Assembly> loadedAssemblies,
+    HashSet<string> visitedAssemblies
+  )
+  {
+    if (visitedAssemblies.Contains(assembly.GetName().Name))
```
BovineOx, line 38: So if we have this assembly by Name and a match, we are happy there's no conflict - Does the Name contain the verson always? Or is this something else.... ?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+
+  public void DetectConflicts()
+  {
+    Dictionary<string, Assembly> loadedAssembliesDict = new();
+    foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+    {
+      // TODO : what to do about multiple versions of loaded dlls?
+      loadedAssembliesDict[assembly.GetName().Name] = assembly;
+    }
+    LoadAssemblyAndDependencies(typeof(DllConflictManager).Assembly, loadedAssembliesDict, new HashSet<string>());
```
BovineOx, line 29: I wonder if GetCallingAssembly() is better? Especially if we wanted to move this somewhere more generic and call from each connector DLL

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+
+  private static Assembly? GetLoadedAssembly(AssemblyName assemblyName)
+  {
+    try
+    {
+      return Assembly.Load(assemblyName);
+    }
+    catch (FileNotFoundException)
+    {
+      // POC : add logging
```
BovineOx, line 83: hmmm, so it occurs to me we're doing this here in DUI3-DX but I think the intention was to do this for existing DUI2 - maybe this wasn't clear mind. POC: comments (let's have that spacing consistent so w can always find these POC: comments)

Do we know why we would get these exceptions? And why is it okay to swallow them?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+      {
+        continue;
+      }
+
+      if (loadedAssemblies.TryGetValue(assemblyName.Name, out Assembly? loadedAssembly))
+      {
+        if (loadedAssembly.GetName().Version != assemblyName.Version)
+        {
+          _assemblyConflicts[assemblyName.Name] = new(assemblyName, loadedAssembly);
+          continue; // if we already know there is a conflict here, no need to continue iterating dependencies
```
BovineOx, line 56: Maybe. Nothing like hitting one error at a time though, maybe worth knowing ALL the problems we can see?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
```diff
@@ -40,27 +41,40 @@ public RevitExternalApplication()
...
-        .Build();
-
-      // resolve root object
-      _revitPlugin = _container.Resolve<IRevitPlugin>();
-      _revitPlugin.Initialise();
+      conflictManager.DetectConflicts();
+
+      InitializePlugin(application, conflictManager);
+
+      conflictManager.WarnUserOfPossibleConflicts();
```
BovineOx, line 65: What happens if InitializePlugin() fails. Could InitializePlugin() not explode before being able to warn the user of the possible conflicts?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
```diff
@@ -69,6 +83,35 @@ public Result OnStartup(UIControlledApplication application)
...
+  {
+    // POC: not sure what this is doing...  could be messing up our Aliasing????
+    //AppDomain.CurrentDomain.AssemblyResolve += OnAssemblyResolve;
+    _container = new AutofacContainer(new StorageInfo());
+    _container.PreBuildEvent += _container_PreBuildEvent;
+
+    // init DI
+    _container
+      .LoadAutofacModules(_revitSettings.ModuleFolders)
+      .AddSingletonInstance(conflictManager) // conflict detector
```
BovineOx, line 105: Are we ever going to resovle this?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+      sb.AppendLine();
+      sb.AppendLine($"Conflicting dll folder: {assemblyConflictInfo.GetConflictingExternalAppName()}");
+
+      TaskDialog.Show("Conflict Report 🔥", sb.ToString());
+      return true;
+    }
+    return false;
+  }
+
+  private static bool TryParseTypeNameFromMissingMethodExceptionMessage(string message, out string? typeName)
```
BovineOx, line 119: Nice method name! Seriously!

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+      sb.AppendLine($"Dependency Name:  {assemblyConflictInfo.SpeckleDependencyAssemblyName.Name}");
+      sb.AppendLine($"Expected Version: {assemblyConflictInfo.SpeckleDependencyAssemblyName.Version}");
+      sb.AppendLine($"Actual Version:   {assemblyConflictInfo.ConflictingAssembly.GetName().Version}");
+      sb.AppendLine($"Conflicting version folder: {assemblyConflictInfo.GetConflictingExternalAppName()}");
+      sb.AppendLine();
+    }
+
+    using var dialog = new TaskDialog("Conflict Report 🔥");
+    dialog.MainInstruction =
+      "Speckle encountered an error loading the following dependencies. This is likely due to a different addin that uses a different version of the same dependency.";
```
BovineOx, line 228: At this point, is this definitely an error?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
```diff
@@ -0,0 +1,245 @@
...
+    dialog.MainContent = sb.ToString();
+    dialog.CommonButtons = TaskDialogCommonButtons.Ok;
+
+    _ = dialog.Show();
+
+    if (dialog.WasExtraCheckBoxChecked())
+    {
+      foreach (var conflictInfo in conflictInfoNotIgnored)
+      {
+        options.DllsToIgnore.Add(conflictInfo.SpeckleDependencyAssemblyName.Name);
```
BovineOx, line 240: I mean this is a bit different as we're ignoring the DLL. I feel there's some versioning info missing somewhere

BovineOx: I also don't follow how we're clearing this (or where atm) if the plugin fails to load ot the host app crashes (if we can detect that)

BovineOx (changes requested):
I think there are some changes and questions. Maybe review and make any changes and then book some time for us to discuss tomorrow before/after the SU?

I think there's some good work in here, just need to understand a bit more

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
AlanRynne: I think the warning should be added in a `finally` block to ensure that whatever happens we tell the user if there were conflicts.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
AlanRynne: Good catch, if we're only checking for DLL conflicts before the container is built, there is no need to register this class within the container 🤔

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
```diff
@@ -69,6 +83,35 @@ public Result OnStartup(UIControlledApplication application)
...
   }
 
+  /// <summary>
+  /// If we directly use ex.IsFatal() in the above method, and there is an outdated version of Core loaded,
+  /// the MissingMethodException will be throw prior to executing the OnStartup method. We would prefer if
+  /// the exception is thrown after the <see cref="DllConflictManager.DetectConflicts"/> method is executed
+  /// </summary>
+  /// <param name="ex"></param>
+  /// <returns></returns>
+  private static bool IsFatal(Exception ex) => ex.IsFatal();
```
AlanRynne, line 93: Not sure this is the best way to go about this. Maybe we should just decouple the conflict detection from any Speckle related stuff and have it run before anything else, but maybe I'm missing something here 🤔 

Right now, conflict resolution lives in a DLL that depends on Core, which inevitably leads to us having issues detecting conflicts in Core itself. I've done this type of things in the past on Grasshopper and they don't end up aging well 🥲

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManagmentOptionsLoader.cs
AlanRynne, line 1: So we're storing this JSON somewhere alongside the plugin? This location seems tricky as we ensure everything is removed from a previous install before installing a new version.

We should ensure either this file doesn't get deleted, or that it is copied into the %appdata% folder instead?

BovineOx:
@connorivy in line with some comments above let's do the following:
- Make changes for this issue branched of Dev
- Move all of the code out of Revit into a project Revit should consume - let's chat when you get online to confirm where this should be or if it should be a new project
- Consume in Revit for DUI2
- Consume in Revit for DUI3
- Related, we'll get the merge through main -> dev -> dui3/alpha

There will need to be multiple PRs for this

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
connorivy: We don't need to do this. Per the acceptance criteria, the plugin is warning users about potential conflicts in the "WarnUserOfPossibleConflicts" method. If intializePlugin fails due to a typeLoadException or MissingMethodException, then the exception will get handled in the catch block below by the conflictManger. 

The conflict manager will be able to tell the user that a conflict definitely did happen, as opposed to just warning that a conflict may happen.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
connorivy: Gotcha. Yeah this makes sense.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/RevitExternalApplication.cs
connorivy: Oh yeah I was resolving this but then I removed that code. I think we probably do want to resolve this eventually, but I can remove the registration in the meantime.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
connorivy: ah okay, I will make a new branch then

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
connorivy: Good point. It was an error loading the dependency that we needed, but it may not actually be an issue. I'll change the wording

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManager.cs
connorivy: we're not clearing the list anywhere. If the plugin initializes successfully, then we show the warning and skip the ignored conflicts. If the plugin throws an exception on initialization, we skip checking for ignored conflicts and just show the dialog

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/DllConflictManagment/DllConflictManagmentOptionsLoader.cs
connorivy: very good point

PR #3263, opened 2024-04-08
POC/WIP Implementation of Blocks (instances) for Rhino

didimitrie:
It's a shift from the previous implementation. What we're doing: 

On Send: 
- converting atomic elements, and keeping them on their respective layers 
- creating proxies for instances (treated similarly as atomic objects) and instance definitions (held separately in the root object) 
- keeping track of the maximum depth at which an instance/instance definition is found

On Receive: 
- Stage 1: converting all atomic objects to their respective layers
- Stage 2: creating (connector bound) all required instance definitions and instances based on a simple sort order by their maximum depth, starting with definitions always

Works well :) Note the code is a mess and i'd want to port it to the new DI architecture as soon as we've got basic geometry receive conversions in.

### DIFF
--- a/ConnectorRhino/ConnectorRhinoWebUI/Utils/Utils.cs
+++ b/ConnectorRhino/ConnectorRhinoWebUI/Utils/Utils.cs
@@ -20,37 +20,17 @@ public static class Utils
   public static readonly string RhinoAppName = HostApplications.Rhino.GetVersion(HostAppVersion.v7);
   public static readonly string AppName = "Rhino";
 #endif
-
-  // POC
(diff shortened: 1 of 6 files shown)


ConnectorRhino/ConnectorRhinoWebUI/Utils/Utils.cs
```diff
@@ -15,3 +21,119 @@ public static class Utils
   public static readonly string AppName = "Rhino";
 #endif
 }
+
+/// <summary>
+/// POC: hacking blocks
+/// </summary>
+public class RhinoInstanceUnpacker
```
didimitrie, line 28: Note: this should be deleted - forgot it in here by accident

PR #3266, opened 2024-04-08
CNX-9163 type mapping dialog not picking up loaded families

connorivy:
## Description & motivation

As some users have reported, the element type mapper in Revit is not picking up all of the types in the document, only all of the types for the category of incoming object. This works well most of the time, but sometimes incoming types are incorrectly categorized. In this case, a use will not be able to map the type of the incoming object to the desired document type.

For example, foundation slabs in Revit are sent to Revit as floors. On receive, the element type mapper will pull all of the floor types from the document. However, foundation slabs use foundation types, not floor types. For this and other edge cases, this PR attempts to solve the issue by pulling all types from the document instead of just the detected category types.

I was initially worried about the performance of querying the document for hundreds of types, but after some brief testing on heavy models, it doesn't seem to be a noticeable change.

## Changes:

The type mapping dialog will populate a dropdown with all of the types in the category that the type mapper THINKS the element is. However, if no types are found that match a user's query string, the type mapper will expand it's query to use all types in the document

## Validation of changes:

Before this PR
image

After this PR
image

## Checklist:


AlanRynne (approved):
It's not very clear to me what this is supposed to do 🤔

But then again... main changes are in DUI2 which we're no longer actively working on so I guess as long as it works this is fine.

PR #3273, opened 2024-04-09
CNX-8397-DLL-Conflict-Handling in existing DUI2 connectors

connorivy:
## Changes:

Added DllConlfictManagement project

### DllConflictManager
- Detects conflicts between assemblies referenced by the connector and currently loaded assemblies in the app domain
  - I think we probably need to add this exception handling in more places. I currently only have it for the loading of the connector when Revit starts up

### DllConflictUserNotifier
- Handles any exceptions that occur, reference them with the conflictManager, and report them to the user. This is an abstract class that is expected to be implemented by the host app to use the host app specific way of notifying a user.

### DllConflictManagmentOptions
- Holds a list of dll conflicts that the user has elected to ignore
- DllConflictManagmentOptionsLoader class will load these options from a json file in the same directory as the connector dll

### RevitExternalApplication
- Moved all types that are not either connector, System, or Revit types out of the "OnStartup" method.
  - This is because if there are conflicting dll types in this method, then an exception will be thrown before the method is even entered
  - By putting the other types in different methods, the type loading exception will be thrown on entering that method. Therefore we can alert the user about the exception

## To-do before merge:

Questions to be answered (probably by @AlanRynne)

- ~~Where in appdata should I keep options.json file? Can this be the same for all c# connectors that use this?~~
- ~~Currently I'm only trying to catch typeLoadException and MissingMethodException. Are there others that will be thrown when dll mismatch occurs?~~
- ~~I'm only catching the above exceptions in the plugin initialization. I'm thinking we also want to catch these for the operations. Is this needed / is it okay to do?~~
- ~~How to log while detecting plugin conflicts? Should I prioritize logger dependencies and then move to connector dependencies?~~
- ~~I'm referencing the System.Text.Json nuget in the new project because it isn't built into .net standard. Should I reference a really new version or a really old version of this?~~

## Checklist:


ConnectorCore/DllConflictManagement/DllConflictManagement.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <RootNamespace>Speckle.DllConflictManagement</RootNamespace>
+    <IsPackable>false</IsPackable>
```
AlanRynne, line 8: Added this to prevent this project from being released as a nuget

ConnectorCore/DllConflictManagement/DllConflictManagement.csproj
```diff
@@ -0,0 +1,19 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsDesktopBuild)' == false">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="4.6.0" />
```
AlanRynne, line 16: Note to @BovineOx and my future self. We may want to consider using Speckle.Newtonsoft considering many more people will use System.Text.Json in varying versions

AlanRynne (approved):

AlanRynne (changes requested):
I've been poking around this new feature in debug and I have a couple considerations:

- When a user sees this pop-up, it is unclear what they are expected to do about it.
- It is unclear what the consequences are of "not warning about this again"
- The checkbox to not warn will apply to each conflict individually, but users can't individually choose which ones to ignore
- From Matteo: We should have metrics on when does this pop-up get shown, how many conflicts we find or not, etc...

Further discussions may be needed.

teocomi (dismissed):
I was checking the data in MP:

- the `distinct_id` should either be `undefined` or yours, but it has some other value
- the `allDetectedConflicts` value should be a list of objects, but it's a list of strings
- `identifiedConflictingAssemblyName` and some other props you're adding are null, not sure if this is correct?

image

connorivy:
@teocomi,

the distinct id's that you're probably seeing are when I was using a fallback value that didn't come from the accounts (you should know that I pulled this fallback value from the existing Analytics class, so this alternate distinct_id may be being used already). I have changed this and have hard-coded an id of "undefined" for the "lite" version of the analytics. 

some system object types (System.Type, System.Reflection.AssemblyName, System.Reflection.Assembly) were throwing when attempting to be serialized so I just made a custom serialization to string. However, I have now implemented a data transfer object which has good serialization behavior, and I am sending that object instead of just a string.

identifiedConflictingAssemblyName can be null if a single conflict was not identified to be the culprit of a thrown exception. Is there a more clear way to handle this?

[Here](https://eu.mixpanel.com/project/2605593/view/3144341/app/profile#distinct_id=undefined) is an example of a couple events. The first is a "DllConflictsDetected" event and the second is a "DllConflictExceptionDialogShown" event

BovineOx:
> I've been poking around this new feature in debug and I have a couple considerations:
> 
> * When a user sees this pop-up, it is unclear what they are expected to do about it.
> * It is unclear what the consequences are of "not warning about this again"
> * The checkbox to not warn will apply to each conflict individually, but users can't individually choose which ones to ignore
> * From Matteo: We should have metrics on when does this pop-up get shown, how many conflicts we find or not, etc...
> 
> Further discussions may be needed.

For this we could have a blog post and a link to that.

teocomi:
> For this we could have a blog post and a link to that.

There's already a docs page about dll conflicts, if you want to link to it we can add some more explanation about the dialog...

AlanRynne (commented):
I don't see much blockers. Just one class that should be in it's own file, and a question about the autodesk specific method

AlanRynne (approved):
We're looking good! We should make a WIP to really test this in the wild (maybe ping some DPs and users we know have had conflicts in the past?)

PR #3274, opened 2024-04-10
perf(civil): CNX-9178 Civil: Optimised ToSpeckle conversion performance of Tin Surfaces

JR-Morgan:
Tested with the small surface test files, and no regressions (object ids perfectly match)

I'm not able to test any large surfaces, due to receive performance being awful still.

### DIFF
--- a/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
+++ b/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
@@ -731,7 +731,7 @@ public partial class ConverterAutocadCivil
 
     return speckleFeatureline;
   }
-  
+
   // surfaces
   public Mesh SurfaceToSpeckle(TinSurface surface)
   {
@@ -966,12 +966,10 @@ public partial class ConverterAutocadCivil
         {
           continue;
         }
-
-        // TODO: This foreach loop is unacceptably expensive... probably FindVertexAtXY is the culprit, we should avoid using it.
         
         var a1 = surface.FindVertexAtXY(edgeStart.X, edgeStart.Y);
         var a2 = surface.FindVertexAtXY(vertexToAdd.X, vertexToAdd.Y);
-        surface.AddLine(a1, a2); 
+        surface.AddLine(a1, a2);
         a1.Dispose();
         a2.Dispose();
       }
--- a/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
+++ b/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
@@ -745,14 +745,6 @@ public partial class ConverterAutocadCivil
     {
       try
       {
-#if CIVIL2023 || CIVIL2024 // skip any triangles that are hidden in the surface!
-        if (!triangle.IsVisible)
-        {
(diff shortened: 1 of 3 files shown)


Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -736,44 +736,48 @@ private Featureline FeaturelineToSpeckle(CivilDB.CorridorFeatureLine featureline
...
-      foreach (var vertex in triangleVertices)
+      try
       {
-        if (!vertices.Contains(vertex))
+#if CIVIL2023 || CIVIL2024 // skip any triangles that are hidden in the surface!
+        if (!triangle.IsVisible)
         {
-          faceIndices.Add(vertices.Count);
-          vertices.Add(vertex);
+          //TODO: why are we skipping trianlges like this, are we not excluding them already in our call to GetTriangles(false) ?
```
clairekuang, line 751: True - no need to have this codeblock then!

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
```diff
@@ -57,12 +57,9 @@ public Point PointToSpeckle(Point2d point, string units = null)
 
   public Point3d PointToNative(Point point)
   {
-    var _point = new Point3d(
-      ScaleToNative(point.x, point.units),
-      ScaleToNative(point.y, point.units),
-      ScaleToNative(point.z, point.units)
-    );
-    var intPt = ToInternalCoordinates(_point);
+    var scaleFactor = Units.GetConversionFactor(point.units, ModelUnits);
```
clairekuang, line 60: An optional improvement, but if we add `double? conversionFactor = null` as an optional parameter and calculate this at the mesh level, you may be able to squeeze out a bit more performance here as well (instead of running `GetConversionFactor` for every point)

clairekuang (changes requested):
Looks generally good to me! All I'd say for the hotfix is to address the `TODO`

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
JR-Morgan: Very true, I imagine we're using this function in quite a few places, so there's definitely more performance on the cards.

For tin surfaces tho, this wasn't in the hot path, so won't give us a tangeble performance improvement. And I wanted to keep this pr from bloating into other conversions.

For the dui3, we should make sure we don't carry this across, and try and perform fewer unit lookups (at least for the heavy objects like mesh, brep, and curves where we expect potentially hundreds of thousands of points.)

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
JR-Morgan: Removed!

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs
JR-Morgan: OOS

clairekuang (approved):

AlanRynne (approved):
LGTM 👀

PR #3286, opened 2024-04-16
CNX-9228 : minimal working revit receive operation that receives levels

connorivy:
## To-do before merge:

- Invoking common receive operation located in Speckle.Connectors.Utils.Operations with custom RevitHostObjectBuilder
  - Wrapped entire thing in RevitTask to execute in API context
  - Brought over the transactionManager (with slightly different method names) to provided transaction manipulation
  - Brought over the ErrorPreprocessingService (formerly called the ErrorEater) which can be used to either deal with Revit errors or alert the user
  - Brought over the "YieldToUIThread" method to pump the event loop and keep the UI refreshed. The previous one was based on the Avalonia Dispactcher, so this new one would need some testing to see if it is really working.
- RevitConverterToHost which is similar to other host app receive converters
- Brought over level conversion to Revit except left out a chunk of code dealing with receiving levels that have a "referenceOnly" property that I'm not fully understanding. Matteo is probably the best person to explain this.

## Checklist:


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToHost.cs
```diff
@@ -0,0 +1,44 @@
...
+    object result =
+      conversion.Convert(target)
+      ?? throw new SpeckleConversionException($"Conversion of object with type {target.GetType()} returned null");
+
+    return result;
+  }
+
+  private ISpeckleObjectToHostConversion? GetToHostConversion(Type? targetType)
+  {
+    if (targetType is null || targetType == typeof(object))
```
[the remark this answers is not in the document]
BovineOx: Ahh right you're walking back until you hit object. So we are looking at the super type.

DUI3-DX/Converters/Revit/Speckle.Converters.Revit2023.DependencyInjection/AutofacRevitConverterModule.cs
```diff
@@ -17,11 +17,15 @@ protected override void Load(ContainerBuilder builder)
   {
     // most things should be InstancePerLifetimeScope so we get one per operation
     builder.RegisterType<RevitConverterToSpeckle>().As<ISpeckleConverterToSpeckle>();
```
BovineOx, line 19: These should use the InstancePerLifetimeScope()

DUI3-DX/Converters/Revit/Speckle.Converters.Revit2023.DependencyInjection/AutofacRevitConverterModule.cs
```diff
@@ -17,11 +17,15 @@ protected override void Load(ContainerBuilder builder)
...
     // factory for conversions
     builder
       .RegisterType<Factory<string, IHostObjectToSpeckleConversion>>()
       .As<IFactory<string, IHostObjectToSpeckleConversion>>();
+    builder
+      .RegisterType<Factory<string, ISpeckleObjectToHostConversion>>()
+      .As<IFactory<string, ISpeckleObjectToHostConversion>>();
 
     // POC: do we need ToSpeckleScalingService as is, do we need to interface it out?
     builder.RegisterType<ToSpeckleScalingService>().AsSelf().InstancePerLifetimeScope();
```
BovineOx, line 31: These should all be interfaced out

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ErrorPreprocessingService.cs
```diff
@@ -0,0 +1,92 @@
+using Autodesk.Revit.DB;
+using Speckle.Core.Logging;
+
+namespace Speckle.Converters.RevitShared.Services;
+
+public class ErrorPreprocessingService : IFailuresPreprocessor
```
BovineOx, line 6: This seems completely out of scope tbh, we need to design and POC the reporting and this far we've largely been ignoring it. I think this should come out for now - was this a copy and paste?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ToHostScalingService.cs
```diff
@@ -0,0 +1,31 @@
...
+  {
+    if (string.IsNullOrEmpty(units))
+    {
+      return value;
+    }
+
+    return UnitUtils.ConvertToInternalUnits(value, UnitsToNative(units));
+  }
+
+  private static ForgeTypeId UnitsToNative(string units)
```
BovineOx, line 18: Isn't this being done elsewhere?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/TransactionManagementService.cs
```diff
@@ -0,0 +1,249 @@
...
+
+namespace Speckle.Converters.RevitShared.Services;
+
+/// <summary>
+/// Is responsible for all functionality regarding subtransactions, transactions, and transaction groups.
+/// This includes starting, pausing, committing, and rolling back transactions
+/// </summary>
+public sealed class TransactionManagementService : ITransactionManagementService
+{
+  private readonly Lazy<RevitConversionContextStack> _lazyContextStack;
```
BovineOx, line 13: Any reason for making this lazy? Is this essentially a singleton with the context of a conversion?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/TransactionManagementService.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+      var status = _transaction.Commit();
+      HandleFailedCommit(status);
+      _transaction.Dispose();
+      return status;
+    }
+    return TransactionStatus.Uninitialized;
+  }
+
+  private void HandleFailedCommit(TransactionStatus status)
```
BovineOx, line 103: This feels like it maybe doing too much and is using SpeckleLog directly.

BovineOx (changes requested):
We should probably be a bit picky about which bits we pull across from the old converters/connectors as we're getting into areas we want to make good/better and we will need more alignment

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ToHostScalingService.cs
connorivy: I thought the same thing but I don't think so. In the contextStack, we have an IHostToSpeckleUnitConverter<THostUnit> which goes from hostUnit to string, but not string to HostUnit. Maybe we should add this to that interface instead?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/TransactionManagementService.cs
connorivy: this was needed because it was being instantiated in a dependency of the binding (right when Revit was opened before there was an active document) which was causing an error because the context stack is instantiated using the active document. However, this is no longer needed after moving the uow out to the binding.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/TransactionManagementService.cs
connorivy: you are right. Removed the error handling and added POC comments

AlanRynne:
Moving this to `draft` as we'll get receive in later on

connorivy:
closing in favor of https://github.com/specklesystems/speckle-sharp/pull/3542

PR #3287, opened 2024-04-17
Jedds fixes

JR-Morgan:


DUI3-DX/Converters/Autocad/Speckle.Converters.Autocad2023.DependencyInjection/Speckle.Converters.Autocad2023.DependencyInjection.csproj
```diff
@@ -2,6 +2,7 @@
 
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
+        <PlatformTarget>x64</PlatformTarget>
```
oguzhankoral, line 5: Was it needed?

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
```diff
@@ -37,21 +37,20 @@ public ReceiveOperation(IHostObjectBuilder hostObjectBuilder)
       ?? throw new SpeckleAccountManagerException();
 
     // 3 - Get commit object from server
-    Client apiClient = new(account);
+    using Client apiClient = new(account);
     ServerTransport transport = new(account, projectId);
     Commit? version =
-      await apiClient.CommitGet(projectId, versionId, cancellationToken).ConfigureAwait(false)
+      await apiClient.CommitGet(projectId, versionId, cancellationToken).ConfigureAwait(true)
```
oguzhankoral, line 43: this is a common purpose function, not sure this will be the case with other connectors?

DUI3-DX/Converters/Autocad/Speckle.Converters.Autocad2023.DependencyInjection/Speckle.Converters.Autocad2023.DependencyInjection.csproj
JR-Morgan: Because the autocad api is x64 only. If you ommit a `PlatformTarget` it defaults to `AnyCPU`. this will give you a msbuild warning saying there is a mismatch u
between processor architecture. (because you can't be compatible with any CPU if you depend on x64 dependencies)

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
JR-Morgan: Good point.
`Operations.Receive` doesn't need to be done on the main thread.
Perhaps we should be syncing to main thread much further down the call sack (inside the `hostObjectBuilder`)

JR-Morgan: Changed to using ISyncToMainThread

oguzhankoral (approved):

PR #3289, opened 2024-04-17
simplify parameter conversion to Speckle

connorivy:
## Description & motivation

simplify parameter conversion to Speckle

## Changes:

- remove caching service that was moved over from the old connector
- fix various bugs in the parameter sending process

## Checklist:


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/ParameterConversionToSpeckle.cs
```diff
@@ -1,86 +1,40 @@
...
-      isShared = IsShared,
-      isReadOnly = IsReadOnly,
-      name = Name,
-      units = UnitsSymbol ?? "None",
-      value = value
+      applicationInternalName = internalName,
+      applicationUnit = unitTypeId?.ToUniqueString() ?? "None",
+      isShared = target.IsShared,
+      isReadOnly = target.IsReadOnly,
+      name = definition.Name,
```
BovineOx, line 35: we're sure neither can be null?

BovineOx (commented):
A couple of minor comments, so if all's good I can approve tomorrow

AlanRynne (approved):
Other than the nullable comment from Ian this looks good. 🚀

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/ParameterConversionToSpeckle.cs
connorivy: I'm honestly not sure, this is just how we are doing it in the existing connector.  I don't think I've run into any null values for parameter definition, but that certainly doesn't mean it can't happen. I don't think the Revit api documentation has accurate nullability information. How would you suggest I handle situations such as these?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterValueExtractor.cs
```diff
@@ -20,6 +20,11 @@ public ParameterValueExtractor(ToSpeckleScalingService scalingService)
 
   public object? GetValue(Parameter parameter)
```
BovineOx, line 21: I could be missing something, but I presume we cannot be more specific than object and there's no way to remove the nullable?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/LevelConversionToSpeckle.cs
```diff
@@ -5,31 +5,23 @@ namespace Speckle.Converters.RevitShared.ToSpeckle;
 
 public class LevelConversionToSpeckle : IRawConversion<DB.Level, SOBR.RevitLevel>
 {
-  private readonly CachingService _cachingService;
   private readonly ToSpeckleScalingService _scalingService;
 
-  public LevelConversionToSpeckle(CachingService cachingService, ToSpeckleScalingService scalingService)
+  public LevelConversionToSpeckle(ToSpeckleScalingService scalingService)
```
BovineOx, line 10: Maybe we should invert this? ScalingServiceToSpeckle @AlanRynne ?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/ParameterConversionToSpeckle.cs
BovineOx: I think if it's working atm then let's keep it as is, we will at worst get an NRE at a higher level - if we start seeing these, we can look at make changes

BovineOx (approved):
Looks generally okay, just a couple of questions before merging...

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/LevelConversionToSpeckle.cs
AlanRynne: I would be inclined to call it ScalingServiceToSpeckle as it sounds more like proper english 😄

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterValueExtractor.cs
connorivy: Yeah, we are already being more specific. 

If you look at this method, you'll see that it is basically just a switch on the parameterType which then dispatches the appropriate, strongly typed, method. These are public methods, so if you know that a parameter is going to be a double, then you can call GetValueAsDouble().

I don't think we need to worry about the nullability. We've had issues in the past with sending default values in the past (i.e. sending unset parameters with a value of 0 instead of a value of null)

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/LevelConversionToSpeckle.cs
connorivy: okay, renamed

connorivy:
@BovineOx @AlanRynne I renamed the scalingService. You can merge if everything looks good noew

PR #3292, opened 2024-04-19
CNX-9216 Receive - Rhino Geometry Conversions

AlanRynne:
https://spockle.atlassian.net/browse/CNX-9216

Just opening this to sort out some final details and have CI checks

- [x] Missing some XML comments (coming up soon!)


DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/RhinoHostObjectBuilder.cs
```diff
@@ -40,7 +41,7 @@ CancellationToken cancellationToken
 
     var objectsToConvert = rootObject
       .TraverseWithPath(obj => obj is not Collection)
-      .Where(obj => obj.Item2 is not Collection);
+      .Where(obj => obj.Item2 is not Collection && obj.Item2 is not DisplayStyle && obj.Item2 is not RenderMaterial);
```
BovineOx, line 44: Might it be wise to wrap these rules, they feel a but like implementation? But, I didn't get to have my traversal chat with @JR-Morgan so likely I am just lacking some context.

BovineOx (approved):
This was my 2 cents...

I think that for raw conversions we should always end this with RawConversion, the interface is IRawConversion, so instead of SpeckleSpiralRawToHostConversion I think it should be SpeckleSpiralToHostRawConversion

AlanRynne:
I'm merging this in if CI passes, leaving the naming discussion (and refactor) for when we decide what to do. Should be a 10 min thing regardless

PR #3295, opened 2024-04-19
CNX-9289 apply unit of work for revit send binding

connorivy:
## Changes:

- inject uow factory instead of singleton sendOperation in the sendBinding
- change sendOperation ioc container registration to scoped
- remove uow factory from send operation and replace it with scoped implementaion of RootObjectBuilder

## Checklist:

### DIFF
--- a/DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
+++ b/DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
@@ -15,6 +15,7 @@ using Speckle.Converters.RevitShared.Helpers;
 using Speckle.Connectors.Revit.Operations.Send;
 using Speckle.Connectors.DUI.Models.Card;
 using Speckle.Connectors.DUI.Bindings;
+using Speckle.Autofac.DependencyInjection;
 
 namespace Speckle.Connectors.Revit.Bindings;
 
@@ -26,24 +27,21 @@ internal class SendBinding : RevitBaseBinding, ICancelable, ISendBinding
   // POC: does it need injecting?
   private HashSet<string> ChangedObjectIds { get; set; } = new();
 
-  // POC: update SendOperation to be interfaced out and could be shared implementation
-  // could be abstract implementation
-  // In the context of the SEND operation, we're only ever expecting ONE conversion
-  private readonly SendOperation _sendOperation;
   private readonly IRevitIdleManager _idleManager;
+  private readonly IUnitOfWorkFactory _unitOfWorkFactory;
 
   public SendBinding(
     IRevitIdleManager idleManager,
     RevitContext revitContext,
     RevitDocumentStore store,
     IBridge bridge,
-    SendOperation sendOperation
+    IUnitOfWorkFactory unitOfWorkFactory
   )
     : base("sendBinding", store, bridge, revitContext)
   {
     _idleManager = idleManager;
+    _unitOfWorkFactory = unitOfWorkFactory;
     Commands = new SendBindingUICommands(bridge);
-    _sendOperation = sendOperation;
 
     // TODO expiry events
     // TODO filters need refresh events
@@ -71,8 +69,6 @@ internal class SendBinding : RevitBaseBinding, ICancelable, ISendBinding
 
   private async Task HandleSend(string modelCardId)
   {
-    // POC: should THIS be where we begin the UoW
-
     // POC: probably the CTS SHOULD be injected as InstancePerLifetimeScope and then
     // it can be injected where needed instead of passing it around like a bomb :D
     CancellationTokenSource cts = CancellationManager.InitCancellationTokenSource(modelCardId);
@@ -82,7 +78,9 @@ internal class SendBinding : RevitBaseBinding, ICancelable, ISendBinding
       throw new InvalidOperationException("No publish model card was found.");
     }
 
-    string versionId = await _sendOperation
+    using IUnitOfWork<SendOperation> sendOperation = _unitOfWorkFactory.Resolve<SendOperation>();
+
+    string versionId = await sendOperation.Service
       .Execute(
         modelCard.SendFilter,
         modelCard.AccountId,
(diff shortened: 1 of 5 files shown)


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/AutofacUIModule.cs
```diff
@@ -78,9 +78,6 @@ protected override void Load(ContainerBuilder builder)
     // register
     builder.RegisterType<RevitDocumentStore>().SingleInstance();
```
AlanRynne, line 79: If there can only ever be one document store, we should justify in a comment why is that to prevent someone changing it without context.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/SendOperation.cs
```diff
@@ -39,19 +37,14 @@ public SendOperation(IUnitOfWorkFactory unitOfWorkFactory, IRootObjectSender roo
...
-      // everything is being passed in from the caller? It feels like the caller should be instantiating the UoW
-      commitObject = rootObjectBuilder.Service.Build(
-        new SendSelection(sendFilter.GetObjectIds()),
-        onOperationProgressed,
-        ct
-      );
-    }
+    // POC: have changed this as I don't understand the injecting of the ISendFilter when we can just use it here
+    // it begs the question whether ISendFilter should just be injected into the roo object builder and whether this function needs it at all?
+    // this class is now so thing I wonder if it should exist at all?
```
AlanRynne, line 42: I somewhat agree, but this will be moved into a shared project where you won't even have to implement it yourself.

It does kinda feel like the filter is a "RootObjectBuilder" thing, but the separation is also nice.

As it stands send operation does:
- Find the object IDs to send based on selection
- Build the root object based on those IDs
- Send the root object to speckle

I'm sure the pattern could be streamlined in a way that makes this slightly less awkward 🙌🏼

DUI3-DX/Converters/Revit/Speckle.Converters.Revit2023.DependencyInjection/AutofacRevitConverterModule.cs
```diff
@@ -28,8 +29,7 @@ protected override void Load(ContainerBuilder builder)
     builder.RegisterType<ToSpeckleScalingService>().AsSelf().InstancePerLifetimeScope();
     builder.RegisterType<RevitConversionContextStack>().AsSelf().InstancePerLifetimeScope();
 
-    // POC: check with CI speckler but this AsImplementedInterfaces() seems wrong or non-specific here
-    builder.RegisterType<RevitToSpeckleUnitConverter>().AsImplementedInterfaces().SingleInstance();
+    builder.RegisterType<RevitToSpeckleUnitConverter>().As<IHostToSpeckleUnitConverter<ForgeTypeId>>().SingleInstance();
```
AlanRynne, line 32: Are we certain this class will never need the `context` for anything?

I would suggest changing this as `InstancePerLifetimeScope`, this allows us to heavily rely on the "unit of work" to deliver a completely clean set of classes, without re-use.

This may result in some extra instance initialisations but it's just 1 per operation (send/receive) so the performance hit should be neglible.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
```diff
@@ -82,7 +78,9 @@ private async Task HandleSend(string modelCardId)
       throw new InvalidOperationException("No publish model card was found.");
     }
 
-    string versionId = await _sendOperation
+    using IUnitOfWork<SendOperation> sendOperation = _unitOfWorkFactory.Resolve<SendOperation>();
```
AlanRynne, line 81: Unless we have a powerful reason not to, we should declare the unit of work as the first thing in the `HandleSend` method.

This will make it very clear that anything inside this is considered it's own bubble. Having it a couple lines down makes it less obvious.

AlanRynne (changes requested):
Just some minor comments, may not be blockers but I'd like to start unifying our understanding of these things.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/SendOperation.cs
BovineOx: There may be a better shape to this, but if we get to unit testing this stuff, the fact it is thin will make it very easy to test

DUI3-DX/Converters/Revit/Speckle.Converters.Revit2023.DependencyInjection/AutofacRevitConverterModule.cs
BovineOx: I think generally when used with a UoW we should probably go with InstancePerLifetimeScope because if we start adding any logging or complexity, when that work is done, it will naturally be within the same ILogger, which should be decorated with the CorrelationID and therefore easy to see all logs for a given operaton.

BovineOx (changes requested):

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
connorivy: I'm not understanding the benefits of this. "This will make it very clear that anything inside this (referring to the Method, I think?) is considered it's own bubble. Having it a couple lines down makes it less obvious". 

The unit of work _IS_ its own bubble. By bubble I am referring to a separate IOC container scope. This is somewhat enforced by the IDisposable implementation. The stuff above is validation logic that isn't part of the same scope. The CancellationTokenSource object's lifetime is completely different to the lifetime of the SendOperation, and dragging it down into the disposable block makes it seem like it has the same lifetime which isn't true. It's a pretty trivial difference, but I don't see any benefit to creating this object and its scope until we need it.

AlanRynne (approved):

PR #3296, opened 2024-04-19
CNX-9223 revit beam

connorivy:
## Description & motivation

This is a new pattern for handling host objects that can be converted into several different types of Speckle objects. 

I can't think of a Revit case where this pattern wouldn't work, but it would be good to hear from others connector owners to consider cases in different host apps.

#### IConditionalToSpeckleConverter
- converter object that can sometimes convert a certain host object to a Base object
- Has a canconvert() method that tests if the target method meets the criteria required for the converter to convert to target

#### IToSpeckleConverter
-  Has a ConvertToSpeckle() method that just returns a base object. This is needed if there are multiple conditional converters that can handle the same host object, but return different Speckle objects
- I previously made a different pr that had a similar interface, and @AlanRynne removed it. It is definitely worth asking why he removed it and if we shouldn't add it back.

#### DI
- more generic way of registering types with the container. Can be extended to register types with different lifetimes if needed.

## Checklist:


DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/RawConversionRegisterer.cs
```diff
@@ -40,4 +41,27 @@ private static void RegisterRawConversionsForType(ContainerBuilder containerBuil
...
+        : interfaceTypes.Where(it => it == interfaceType);
+
+      foreach (var iterfaceTypeMatch in interfaceTypes)
+      {
+        containerBuilder.RegisterType(type).As(iterfaceTypeMatch).InstancePerLifetimeScope();
+      }
+    }
+
+    return containerBuilder;
+  }
```
AlanRynne, line 66: Neat, my RegisterRawConversion functions ran through all assemblies, which at the time I already considered somewhat unsafe. Loading from specified assemblies is so much better and gives us somewhat control over what we want to load up.

AlanRynne (changes requested):
Some comments regarding the new interfaces and pattern, I think we could tweak it a bit to better align with IRawConversion pattern.

I also have some concerns about loosing the execution order of the original Beam method (which had a bunch of if statements). This may not be an issue for this conversion (haven't really checked) but it could happen that to `CanConvert` return true for the same object, and with the injected IEnumerable we loose the ability to force one to be checked before the other.

This wasn't an issue before bc the order was manually crafted, but looking at some of this conditions are relying on the previous one running and returning null, or not running at all? Both could be the case.

Objects/Converters/ConverterRevit/ConverterRevitTests/ConverterRevitTestsShared/Generated/REVIT2023/TestGenerator/TestGenerator.Generator/GeneratedTests.g.cs
AlanRynne, line 1: I'm assuming this file was deleted by accident?

AlanRynne (approved):
All good with the latest changes, just check if that generated file needs to be there and we're good to go.

I'm approving this so you can merge it while I'm sleeping 👍🏻

PR #3298, opened 2024-04-22
Brings CreateNewVersionInProject up to parity with SpeckePy

secretlyagoblin:
Per the issue logged in the Automate Bug reports in February:

https://speckle.community/t/automate-bug-reports/9136/20

The functionality in this PR wasn't hooked up - you can compare it to the same lines in Specklepy: https://github.com/specklesystems/specklepy/blob/6469b6f757f00bb93041a835c04da71ceff1d3d6/src/speckle_automate/automation_context.py#L162 

So the Resulting Models subsection in for automations isn't populating in C#

### DIFF
--- a/Automate/Speckle.Automate.Sdk/AutomationContext.cs
+++ b/Automate/Speckle.Automate.Sdk/AutomationContext.cs
@@ -134,6 +134,12 @@ public class AutomationContext
         }
       )
       .ConfigureAwait(false);
+
+    if (!string.IsNullOrEmpty(versionId))
+    {
+      AutomationResult.ResultVersions.Add(versionId);
+    }
+
     return versionId;
   }
 


AlanRynne:
Oh! Nice catch @secretlyagoblin 🙌🏼 🙇🏼‍♂️ We'll get this in ASAP, pending a chat with the team.

AlanRynne (approved):
Everything looks correct for what I can gather.

BovineOx (approved):

JR-Morgan (approved):
LGTM,
I have a question for the team r.e. the nullability of the CommitCreate return.
I think we've left it null oblivious so this PR is correct to do the null check.

This is something we should clarify when we if we ever create a new `VersionCreate` endpoint

PR #3304, opened 2024-04-23
CNX-9221 DUI3-DX DirectShape conversion to Speckle 

AlanRynne:
CNX-9221

- Adds DirecShape conversion to speckle
- Fixes issue with parameter extractor assuming all elements have a valid typeId.
- Deleted the existing mesh converter that took a list as input.
- New `MeshConversionToSpeckle` to just be a 1 to 1 conversion.
- Added Raw MaterialToSpeckle conversion and injected it where needed.
- Deleted `MeshBuildHelper` and `MeshDataTriangulator` now replaced by a more specific raw conversion
- Deleted `SolidsConversionToSpeckle` in favor of single solid conversion `SolidConversionToSpeckle` (note the `s` 😅)
- New `IRawConversion<Dictionary<DB.ElementId, List<DB.Mesh>>, List<SOG.Mesh>>` (super verbose I know...) which takes in a Dictionary holding meshes grouped by their materialId and returns a list of Speckle meshes (1 per material)
- Brought over some Category Extensions

## UPDATE:

Wanting to unify some of the duplicated logic I had created, while also adding a specific Solid conversion, lead me to investigate our `MeshBuildHelper` and `MeshDataTriangulator`.

The purpose of the 2, in combination, was to create a single speckle mesh for each material encountered in a list of solids. This is a performance optimisation, as a Solid could have 1000s of faces with the same material. The result is 1 mesh for each material the solid used.

The problem is that it was unclear what was the intended use of these 2 classes, and specially `MeshBuildHelper` was performing some very obscure caching that was, in it's core, just a way to group meshes by material.

### My proposed change

In order to better express the intent of this, I've created a new conversion for this purpose. It takes in a Dictionary<ElementId, List<DB.Mesh>>, which is a map of unique material IDs used in a solid, and the list of meshes that were using each material.

The output of this converter is a list of Speckle meshes (1 per material).

The action of assembling this dictionary is trivial and not speckle-specific, so I left it inside the Solid converter as a utility function.

> Admittedly, the type is quite bonkers... but is also very specific.
> `IRawConversion<Dictionary<DB.ElementId, List<DB.Mesh>>, List<SOG.Mesh>>`

The functionality that `MeshDataTriangulator` was doing is now part of this converter, as it is the only place it was needed. I would be happy to move it out if need be, but the call inputs of the original class where a bit miss-leading.

### Caveats:

- Previously, we were converting all solids into meshes by material, ending up with meshes from different solids becoming 1 single material mesh. I don't think this was exactly what we intended.
    - My solution now only optimises meshes by material on each solid. I think this is a way better choice when receiving in other software such as Rhino/ACAD.
    - Since the converter now accepts a dictionary of meshes, if we want the previous behaviour it can easily be done without major changes by computing the dictionary out of many solids instead.


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
```diff
@@ -0,0 +1,74 @@
...
+    _parameterObjectAssigner = parameterObjectAssigner;
+  }
+
+  public override SOBR.DirectShape RawConvert(DB.DirectShape target)
+  {
+    var category = target.Category.GetBuiltInCategory().GetSchemaBuilderCategoryFromBuiltIn();
+    var element = target.get_Geometry(new DB.Options());
+
+    var geometries = new List<Base>(); // POC: This could be an list of meshes.
+    foreach (var geometryObject in element)
```
connorivy, line 38: As you know, in the existing connector, we are converting a list of revit meshes into a list of speckle meshes. I am speculating the reason is to that our meshes have different requirements than Revit meshes. 

In Revit, a single object is made up of several solids (as illustrated by this loop you've written), and as you can see by your code on line 47, solids have different faces with different materials.

Perhaps the viewer only supports one material per mesh? I'm guessing this because the existing mesh conversion groups by material and creates one mesh per material.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
```diff
@@ -0,0 +1,74 @@
...
+            // Override mesh material (which in this case should be null) with face material if it exists.
+            if (doc.GetElement(face.MaterialElementId) is DB.Material faceMaterial)
+            {
+              var speckleMaterial = _materialConverter.RawConvert(faceMaterial);
+              speckleMesh["renderMaterial"] = speckleMaterial; // POC: Mesh not having a typed renderMaterial is madness
+            }
+
+            geometries.Add(speckleMesh);
+          }
+          break;
```
connorivy, line 61: This code has 5 levels of indentation which seems like a bit much. Also, perhaps this solid -> mesh code would be useful elsewhere. Maybe we should have a typedConverter handle the solids like we have one handling the meshes above?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+
+    SOBR.DirectShape result = new(target.Name, category, geometries) { displayValue = geometries };
+
+    // POC: Parameter extractor exlodes with DS, returns a null elementType somewhere.
```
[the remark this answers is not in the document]
AlanRynne: They did, i left it as reference for you :) Will remove in the end.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
AlanRynne: I think this was an optimisation we did on Revit side to potentially generate less meshes, as you could end up with a solid of 1000s of tiny faces.

For what I gather, it ends up spitting out as many meshes as materials, but does so in the most convoluted way ever 😅

It may be that this performance optimisation was needed, but I don't see why we would keep it as-is. A rewrite of that logic is necessary to keep things slightly de-tangled.

It could be that our `Solid->SpeckleMesh` conversion requires this optimisation (In which case I'd argue it would be a `Solid->List<SpeckleMesh>` conversion instead, but we should still have a `Mesh->Mesh` conversion that does just the bare minimum.

WDYT?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
AlanRynne: I agree, solid to mesh conversion should be injected :) I'll do that.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
connorivy: does the mesh object not suffer from the same issue? In the direct shape converter that you've written, you're looping through a list of geometry objects and converting each of them to a single mesh. Is it not possible that this will also lead to us have many small meshes?

AlanRynne: It does! I've now created a `SolidToSpeckleConversion` that does this, but only for each solid (see reasoning on the PR description I just updated)

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/DisplayValueExtractor.cs
JR-Morgan, line 1: Just a thought, not a comment about this PR intrinsicly. 
But it would be really nice if we separated all our converter projects into `ToHost` and `ToSpeckle` namespaces.
I should know just by looking at the namespace what a class like this is used for.

connorivy:
@AlanRynne 

It looks like you did fix the one to many issue in the "DisplayValueExtractor", but DirectShape conversion is not using that class. Why is that? The direct shape conversion implements it's own displayValue extraction which seems to be mapping the meshes 1 to 1 (which I don't think is what we want)

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/DisplayValueExtractor.cs
AlanRynne: I've done this in Rhino and I have been thinking of doing this too everywhere.

Maybe an item to do at the same time we rename the classes to the new naming standard.

AlanRynne:
@connorivy I left this on purpose but to be honest, the more I think about it I'm convinced my reasoning was not valid, and it would in fact be much better to just inject the displayValue extractor and unify all the logic.

I was thinking of DirectShapes as being their own thing but in reality there is no practical difference in the way we do things now between both (since we don't support solids on send).

Changing it right now :D

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
```diff
@@ -11,42 +11,26 @@ namespace Speckle.Converters.Revit2023.ToSpeckle;
...
-      {
-        case DB.Mesh mesh:
-          geometries.Add(_meshConverter.RawConvert(mesh));
-          break;
-        case DB.Solid solid:
-          geometries.AddRange(_solidConverter.RawConvert(solid));
-          break;
-      }
-    }
+    // POC: Making the analogy that the DisplayValue is the same as the Geometries is only valid while we don't support Solids on send.
```
connorivy, line 32: I'm not sure what this comment means. We are supporting solids on send, right? We are using an IRawConversion<DB.Solid, List<SOG.Mesh>> during this method. I'm wondering if were are referring to different solids.

connorivy (approved):
Looks great!

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/DirectShapeConversionToSpeckle.cs
AlanRynne: What I meant is that we don't support solids as their native representation, rather as a lossy list of meshes. DisplayValues should only have simple geometry (lines, polylines, meshes); so if/when we support solids, this conversion will need to be updated as the list of geometries will no longer be the same as the `displayValue` list.

PR #3305, opened 2024-04-23
Arcgis send conversions

KatKatKateryna:
## Checklist:


KatKatKateryna:
Main changes:
- Send Conversions for Point, Polyline, Polygon and Multipatch Vector layers; Standalone tables; Rasters; Pointclouds
- QueuedTask moved to overall layer conversion, top level. So it doesn't need to be called for every small operation
- better way of getting layer Field values
- injecting GeometryUtils interface
- edit some class definitions: make properties non-nullable and add constructors

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/GlobalUsings.cs
```diff
@@ -1,3 +1,4 @@
 global using SOG = Objects.Geometry;
+global using SGIS = Objects.GIS;
```
AlanRynne, line 2: Not sure if I would prefer doing `SOG.GIS.XXX` rather than making a new shorthand for it.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/GeometryUtils.cs
AlanRynne, line 1: I suspect these could just become extension methods instead of an injected utility class.

None of this have any state, but not sure what's @BovineOx opinion on this.

This class seems to have multiple responsibilities and be bound to keep growing. I would propose creating `xxxExtension.cs` classes (color extensions, mesh extensions, polygon extensions...)

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/GeometryUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+
+  public int CIMColorToInt(CIMColor color)
+  {
+    return (255 << 24)
+      | ((int)Math.Round(color.Values[0]) << 16)
+      | ((int)Math.Round(color.Values[1]) << 8)
+      | (int)Math.Round(color.Values[2]);
+  }
+
+  public bool IsClockwisePolygon(List<SOG.Point> points)
```
AlanRynne, line 33: This signature is kinda odd, a list of point's in speckle doesn't necessarily represent a Polygon. Is there no higher class in speckle that you can use? 

I'm assuming these points will become a polyline, in which case why not have the polyline as input instead? This could allow us to make this an extension method of that class.

I'm also not fully certain why we need to be doing this operations. Why do we care? :)

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisFeatureToSpeckleConverter.cs
```diff
@@ -2,41 +2,99 @@
...
+            int ptCount = boundaryPts.Count;
+            List<int> faces = new() { ptCount };
+            faces.AddRange(Enumerable.Range(0, ptCount).ToList());
+
+            displayVal.Add(
+              new SOG.Mesh(boundaryPts.SelectMany(x => new List<double> { x.x, x.y, x.z }).ToList(), faces)
+            );
+          }
+        }
+        // POC: how to deal with multiploygons, where not all would have displayValue? Will they not be all received in other apps?
```
AlanRynne, line 88: IMO, the meshing hack for polygons shouldn't be something we rely upon. Instead, we should request the viewer should be able to use fills in shapes, not just line colors.

This would also be applicable to other objects in other host apps

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisFeatureToSpeckleConverter.cs
```diff
@@ -2,41 +2,99 @@
...
+      }
+    }
+
+    try
+    {
+      var shape = (ACG.Geometry)target["Shape"];
+      var speckleShapes = _geometryConverter.RawConvert(shape).ToList();
+
+      // re-shape the GisFeature:
+      // if shapes is a list of Meshes, set them as DisplayValue instead of geometry
```
AlanRynne, line 55: I think we should move away from these weird edge cases where the same function can return radically different things with different meaning.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisFeatureToSpeckleConverter.cs
```diff
@@ -2,41 +2,99 @@
...
       }
-      i++;
+
+      // otherwise set shapes as Geometries
+      return new GisFeature(speckleShapes, attributes);
+    }
+    catch (Exception ex) when (ex is ArgumentOutOfRangeException or GeodatabaseFieldException) // if no geometry
+    {
+      return new GisFeature(attributes);
     }
```
AlanRynne, line 98: I'm leaning towards not accepting this. If this exception is trying to catch the "no geometry" scenario, I'm sure there's a better way to do it than letting it explode.

This also doesn't let us know if the problem was that the ArcGIS feature had no geometry, if we failed to convert it, or if we failed to mesh the geometry.

These are 3 separate edge cases that ideally should be dealt with separately (and try/catched with whatever exceptions we can recover from.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisRasterToSpeckleConverter.cs
AlanRynne, line 1: This needs to be broken down into smaller methods, maybe even inject parts of it.

My opinion is that the RawConvert logic should be as simple to understand as possible :) The complexity can be hidden in private methods with clear naming.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/MultipatchFeatureToSpeckleConverter.cs
AlanRynne, line 1: Same as above, this conversion logic is too long to just be 1 single method, it needs splitting up.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/EnvelopBoxToSpeckleConverter.cs
```diff
@@ -0,0 +1,56 @@
...
+      target.YMin,
+      target.ZMin,
+      _contextStack.Current.Document.SpatialReference
+    ).ToGeometry();
+    MapPoint pointMax = new MapPointBuilderEx(
+      target.XMax,
+      target.YMax,
+      target.ZMax,
+      _contextStack.Current.Document.SpatialReference
+    ).ToGeometry();
```
AlanRynne, line 36: this can be refactored into a private method and used twice

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/IGeometryUtils.cs
AlanRynne, line 1: If we make these into extensions methods (pending confirmation) we won't need this interface.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/PointcloudLayerToSpeckleConverter.cs
AlanRynne, line 1: Too long and nested implementation, at least it should be broken down into smaller private methods.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/VectorLayerToSpeckleConverter.cs
```diff
@@ -24,33 +30,60 @@ public Base Convert(object target)
...
+      spekleGeometryType = "Polyline";
+    }
+    else if (speckleLayer.nativeGeomType.ToLower().Contains("polygon"))
+    {
+      spekleGeometryType = "Polygon";
+    }
+    else if (speckleLayer.nativeGeomType.ToLower().Contains("multipatch"))
+    {
+      spekleGeometryType = "Multipatch";
+    }
```
AlanRynne, line 84: This could be also refactored out into a private method. It also seems like this comparisons are quite weak: anything that contains `point` will match. Could we make it more specific or is this by design?

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/VectorLayerToSpeckleConverter.cs
```diff
@@ -61,12 +94,35 @@ public VectorLayer RawConvert(FeatureLayer target)
...
+          }
+          element.attributes = elementAttributes;
+
+          // differentiate between Ring Multipatches and Mesh Multipatches
+          if (
+            spekleGeometryType == "Multipatch"
+            && element.geometry != null
+            && element.displayValue != null
+            && element.geometry.Count > 0
+            && element.displayValue.Count == 0
```
AlanRynne, line 113: 🤔 Why are we doing so many checks on Send using speckle objects instead of interrogating the GIS API?

I would think that the GIS feature converter would do this if it is needed, or am I missing something?

AlanRynne (changes requested):
Dropped some comments, I'm not sure they're actual blockers for this PR to go in, but maybe add POC comments on the things that need breaking down into smaller pieces.

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Operations/Send/RootObjectBuilder.cs
```diff
@@ -58,47 +59,50 @@ public RootObjectBuilder(IUnitOfWorkFactory unitOfWorkFactory)
 
     Collection rootObjectCollection = new(); //TODO: Collections
 
-    foreach (string uri in uriList)
+    QueuedTask.Run(() =>
```
JR-Morgan, line 62: Is this `QueuedTask` required. I see we have one already in the `SendOperation`

If and when we were to move `SendOperation` to a shared class, this is where any syncing to main threads will occur. I'd be curious if we may still require more syncing going on in the converters...

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/GlobalUsings.cs
JR-Morgan: But its not `Speckle.Objects.Gis`, its `Speckle.Gis`
Maybe we should have named it `SO.G` and `SO.GIS`
(just a thought provoking suggestion)

JR-Morgan (changes requested):

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/GlobalUsings.cs
AlanRynne: Oh you are so right... 🙇🏼‍♂️🤦🏼‍♂️ Let's leave it be then.......

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/GeometryUtils.cs
KatKatKateryna: Indeed, Polyline can be a better input 👌 The method is called in different scenarios:
1. Multipolygons are written as a list of ArcGIS Polylines, and the only way to understand what is what (outer vs. inner ring) is by orientation. 
2. When we have a simple outer-ring polygon, it will have a 1-faced Mesh created for the displayValue. And in this case, the order of the points is important to make the mesh facing upwards.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisFeatureToSpeckleConverter.cs
KatKatKateryna: My problem here, is how do I assign displayValues to the GisFeature, if the best place to generate it, is within "Geometry converter" of GisFeature. And then, distribute returned geometry primitives into "geometry" (Point, Polyline, GisPolygon) and "displayValue" (Meshes) attributes. Maybe there is a better way?

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisFeatureToSpeckleConverter.cs
KatKatKateryna: These exceptions are catching a specific case when the field "Shape" in inaccessible, but fair point, changed it to IF statement 👍

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/EnvelopBoxToSpeckleConverter.cs
KatKatKateryna: It's min vs max, I could create 2 methods to get those coordinates and unpack them Python style (*list) inside MapPointBuilderEx method, but didn't find a good way to do it on C#

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/VectorLayerToSpeckleConverter.cs
KatKatKateryna: Splitting into a separate method👍  but the simplification is intended for interop between different GIS software. Each of the classes can be either Simple- or Multi-  (and some only simple or only multi), but Speckle writes everything as Multitype by default.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/VectorLayerToSpeckleConverter.cs
KatKatKateryna: killed this part. Indeed, too much verifications, but it still needs to be figured out. Multipatches are basically Meshes, but certain type of Multipatches have outer and inner rings, like polygons, so we cannot consistently convert all types multipatches to meshes. Needs some rethinking.

KatKatKateryna: And instead, if we turn them all into Polygons, then ArcGIS will be confused on receiving that, because it does not allow 3-dimentional polygons

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Operations/Send/RootObjectBuilder.cs
KatKatKateryna: Indeed! I suppose, we remove it from here and leave a POC comment in SendOperation? To check whether move it down the hierarchy

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/GisFeatureToSpeckleConverter.cs
```diff
@@ -2,41 +2,99 @@
...
     {
       string name = field.Name;
+      if (name != "Shape") // ignore the field with geometry itself
+      {
+        try
+        {
+          var value = target[name];
+          attributes[name] = value; // can be null
+        }
+        catch (ArgumentException)
```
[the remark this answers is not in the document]
KatKatKateryna: I just noticed that certain types of fields fail to read their value, but I don't know if the list of failing types is exhaustive.

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Layers/VectorLayerToSpeckleConverter.cs
AlanRynne: I don't think we can say that a `GIS Multipatch` is equivalent to a `Mesh`... They are drawn with 2d curves that represent closed areas, meshes are points grouped by faces.

The analogy between the 2 is computationally expensive and, if we cannot guarantee a consistent result, I would argue we should not be making them meshes at all.

KatKatKateryna: It has nothing to do with gis-gis. On the contrary, it tries to represent gis geometry in a cad friendly way. Also, the multipatches are not all curves with closed areas as I described above, just 1 type (Ring type), all the others are exactly like Speckle meshes: https://pro.arcgis.com/en/pro-app/latest/sdk/api-reference/topic27403.html
Btw, this part of code is already gone, I added instead GisMultipatch class (for multipatches that act like meshes) and GisPolygon3d (for multipatches that are enclosed areas, which is basically a polygon but not necessarily on a horizontal surface).

Objects/Objects/GIS/GisPolygonGeometry3d.cs
```diff
@@ -0,0 +1,19 @@
...
+public class GisPolygonGeometry3d : Base
+{
+  public string units { get; set; }
+  public Polyline boundary { get; set; }
+  public List<Polyline> voids { get; set; }
+
+  public GisPolygonGeometry3d()
+  {
+    voids = new List<Polyline>();
+    units = Units.Meters;
```
JR-Morgan, line 17: I see what you're doing here @KatKatKateryna,
And sorry, I think I may have given you some bad advice by saying its fine to defaulting values in constructor. 

For empty lists, I have no issue. 

But I don't think we should EVER default Units, its just too dangerous and never desirable, only opening us up to the possibility that we receive data in the wrong units.

I would rather us send an null here, and let it fail on the receiving end.

JR-Morgan: Kat's (quite correctly) pointed out that we are doing exactly this in our mesh class, and that this is our current way to do things. 
We'll come up with a better structure as part of https://spockle.atlassian.net/browse/CNX-9352

JR-Morgan (approved):

AlanRynne (approved):

PR #3307, opened 2024-04-23
CNX-9320 resolve duplication of webview xaml control

JR-Morgan:
1. Created a new shared project (`Speckle.Connectors.DUI.WebView`) referenced by all WebView connectors (ArcGIS, Autocad, Rhino)
2. Moved the duplicated xaml file that existed in each of the three projects.
3. Made some changes: (for the aforementioned connectors only)
    - Replaced use of `Lazy<IBinding>` with `IBinding` - Tested, Works!)
    - `ExecuteScriptAsyncMethod` will now throw `InvalidOperationException` instead of `SpeckleException`
    - The shared XAML is using`DockPanel`, This is a change for Autocad only, which was using `Grid` instead
    

@AlanRynne I've also removed that exception handler that Rhino was using to prevent exceptions during binding resolution.
Its no longer needed if we remove `Lazy` since the binding resolution will happen immediately when resolving the `RhinoPlugin` and be handled here:
image

I hope that's satisfactory!


AlanRynne (approved):
No complaints on my side, It's a pretty straightforward refactor.  ♥️

BovineOx (approved):

oguzhankoral (approved):

PR #3309, opened 2024-04-23
CNX-9291 Implement Unit of Work for ArcGIS Send and Receive bindings

KatKatKateryna:
## Checklist:

### DIFF
--- a/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
+++ b/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
@@ -59,7 +59,7 @@ public class AutofacArcGISModule : Module
     // POC: Register here also RootObjectBuilder as IRootObjectBuilder
 
     // binding dependencies
-    builder.RegisterType<CancellationManager>().InstancePerLifetimeScope();
+    builder.RegisterType<CancellationManager>().InstancePerDependency();
 
     // register send filters
     builder.RegisterType<ArcGISSelectionFilter>().As<ISendFilter>().InstancePerLifetimeScope();
--- a/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
+++ b/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
@@ -45,26 +45,29 @@ public class AutofacArcGISModule : Module
     builder.RegisterType<ArcGISSelectionBinding>().As<IBinding>().SingleInstance();
     builder.RegisterType<ArcGISSendBinding>().As<IBinding>().SingleInstance();
     builder.RegisterType<ArcGISReceiveBinding>().As<IBinding>().SingleInstance();
-    builder.RegisterType<ArcGISToSpeckleUnitConverter>().As<IHostToSpeckleUnitConverter<Unit>>().SingleInstance();
+    builder
+      .RegisterType<ArcGISToSpeckleUnitConverter>()
+      .As<IHostToSpeckleUnitConverter<Unit>>()
+      .InstancePerLifetimeScope();
 
     // Operations
     builder.RegisterType<ReceiveOperation>().AsSelf().InstancePerLifetimeScope();
-    builder.RegisterType<SyncToCurrentThread>().As<ISyncToMainThread>().SingleInstance();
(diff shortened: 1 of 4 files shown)


DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
```diff
@@ -45,26 +45,29 @@ protected override void Load(ContainerBuilder builder)
...
+    builder.RegisterType<SyncToCurrentThread>().As<ISyncToMainThread>().InstancePerLifetimeScope();
 
     // Object Builders
-    builder.RegisterType<HostObjectBuilder>().As<IHostObjectBuilder>().InstancePerDependency();
+    builder.RegisterType<HostObjectBuilder>().As<IHostObjectBuilder>().InstancePerLifetimeScope();
     // POC: Register here also RootObjectBuilder as IRootObjectBuilder
 
     // binding dependencies
-    builder.RegisterType<CancellationManager>().InstancePerDependency();
+    builder.RegisterType<CancellationManager>().InstancePerLifetimeScope();
```
AlanRynne, line 62: I'm looking at this and I'm leaning towards admitting I made a mistake changing this one.

I'm assuming there should only be 1 cancellation manager for DUI3? @oguzhankoral could you confirm? If so it should be `SingleInstance`.

Not a blocker for this PR if everything is working, just trying to come up with the right logic.

AlanRynne (approved):
Approving, but I have a question for @oguzhankoral or @didimitrie regarding DUI3 CancellationManager

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
AlanRynne: Ok I just double checked on Rhino and this is still `InstancePerDependency` on my side. I vividly remember changing it but I must have changed it back before merging.

Let's put this back as `InstancePerDependency` just to be sure we're all doing the same things (pending bringing this into a common module)

KatKatKateryna: What would be the consensus on the scope in this case?

AlanRynne: Just revert it to `InstancePerDependency` for now

BovineOx: Wouldn't this be SingleInstance() in this case? We should have a plan tbh to inject the actual CancellationToken so we can inject it instead of passing it about tbh

BovineOx: But, shouldn't such a manager be the same manager for the lifetime of the application?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
```diff
@@ -45,26 +45,29 @@ protected override void Load(ContainerBuilder builder)
...
 
     // register send operation and dependencies
-    builder.RegisterType<SendOperation>().SingleInstance();
-    builder.RegisterType<RootObjectBuilder>().SingleInstance();
-    builder.RegisterType<RootObjectSender>().As<IRootObjectSender>().SingleInstance();
+    builder.RegisterType<SendOperation>().InstancePerLifetimeScope();
+    builder.RegisterType<RootObjectBuilder>().InstancePerLifetimeScope();
+    builder.RegisterType<RootObjectSender>().As<IRootObjectSender>().InstancePerLifetimeScope();
 
     //POC: how tf does this work?
```
BovineOx, line 72: :rofl

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/DependencyInjection/AutofacArcGISModule.cs
AlanRynne: Well, I would think so but all other connectors have it as `InstancePerDependency`.

Since most of this registration is going to be moved into a shared place for all connectors (as they all register the exact same things, and should do so in the same way) we can tackle that then.

I think i have a bullet point In one of the tickets related to this.

PR #3312, opened 2024-04-25
Fix (autocad): CNX-9346 autocad document states are broken

oguzhankoral:
The previous assumption was that always will have a null doc whenever Bindings are created. But whenever we had a document opened already before bindings were created, we were missing track on the first doc and that's why it was failing to follow selections and feedback to UI with empty state if we're switching to the main screen in AutoCAD.

Another PR on UI side is -> https://github.com/specklesystems/speckle-server/pull/2231

Before
acad_XZBmXXlu4W


AlanRynne (approved):
LGTM! 👀

PR #3315, opened 2024-04-26
CNX-9336 CNX-9218 Unify traversal and Rhino fallback conversion

AlanRynne:
CNX-9336 CNX-9218

New Universal traversal function that is NOT hostApp dependent, meaning it will return the same results regardless of the application is running and that (for what we've gathered) it returns only the objects we're interested in converting.

We will explain better how this works in the demo today.

Additionally adds fallback conversion for any object that is not supported directly but has a `displayValue`. Since this is a one-to-many conversion, the fallback conversion result is grouped after baking to ensure they become 1 single unit in the model.

### Caveats
Rhino used to traverse all properties of a `RevitWall` (or any other unsupported object) as if it were a Base object. This is no logner the case, traversal will only traverse the `elements` property of any subClass of base, and will stop at any dynamic Base displayable objects.
This means that we were previously getting things in Rhino like `Levels` for free, but now we stop at the `Wall`, the level is never returned to the flattened list as it doesn't follow any predictable pattern (ish)


AlanRynne:
CI failing for external reasons, fix is coming into dui3/alpha first

AlanRynne:
Since @JR-Morgan is off I'm taking over the changes required to get this in.

AlanRynne:
Closing this in favor of #3327 3327

PR #3319, opened 2024-04-29
CNX-9358 Fix write to file with DocumentToBeDeactivated event

oguzhankoral:
This is how it is working now.

acad_dcKWzYTM7Y


clairekuang:
Did a quick test with send/receive card swapping and adding/removing between different documents (both parented and windowed), looks good to me :)

clairekuang (approved):

PR #3321, opened 2024-04-29
CNX-9292 apply unit of work for auto cad

oguzhankoral:
UoWs added for receive and send. And UoWs in builders replaced with `ISpeckleConverterToSpeckle _converter` and `ISpeckleConverterToHost _converter` dependency.

### DIFF
--- a/DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/AutofacAutocadModule.cs
+++ b/DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/AutofacAutocadModule.cs
@@ -50,7 +50,7 @@ public class AutofacAutocadModule : Module
     // Operations
     builder.RegisterType<SendOperation<(DBObject obj, string applicationId)>>().InstancePerLifetimeScope();
     builder.RegisterType<ReceiveOperation>().SingleInstance();
-    builder.RegisterType<SyncToUIThread>().As<ISyncToMainThread>().SingleInstance();
+    builder.RegisterType<SyncToUIThread>().As<ISyncToMainThread>().SingleInstance().AutoActivate(); // Auto activation makes SynchronizationContext available right after building. Otherwise we were missing it in UoWs.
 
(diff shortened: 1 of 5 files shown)


DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/AutofacAutocadModule.cs
```diff
@@ -50,7 +50,7 @@ protected override void Load(ContainerBuilder builder)
     // Operations
     builder.RegisterType<SendOperation<(DBObject obj, string applicationId)>>().InstancePerLifetimeScope();
     builder.RegisterType<ReceiveOperation>().SingleInstance();
-    builder.RegisterType<SyncToUIThread>().As<ISyncToMainThread>().SingleInstance();
+    builder.RegisterType<SyncToUIThread>().As<ISyncToMainThread>().SingleInstance().AutoActivate(); // Auto activation makes SynchronizationContext available right after building. Otherwise we were missing it in UoWs.
```
AlanRynne, line 53: Does this need propagating to Rhino/Revit?

AlanRynne (approved):
Just one question regarding the AutoActivate 🙌🏼

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/AutofacAutocadModule.cs
oguzhankoral: I believe it is more of a MainThread thing, Rhino should be fine without AutoActive, but not sure about Revit. And @connorivy was saying he did something Revit specific? If `ISyncToMainThread` interface work with Revit too, we should align it I believe.

AlanRynne: Ok no blocker for this then! In we go!

PR #3327, opened 2024-04-30
CNX-9219 CNX-9336 CNX-9218 Traversal unification + ACAD/Rhino Fallback conversion

JR-Morgan:
CNX-9219 CNX-9336 CNX-9218

Traversal:
- Unified Traversal: `GraphTraversal` independent of host app 
- Implemented said traversal in Rhino & Autocad

Fall-back:
- Implemented a unified `ISpeckleConverterToHost`
- Used right now by Autocad + Rhino (With some minor tweaks)

TODO:
 - [x] Remove the Unit tests for traversal
 - [x] Test Autocad
 - [x] Test Rhino
 - [x] Add XML docs to new stuff
 - [x] Inject `GraphTraversal` in rhino


AlanRynne (approved):
I may be biased but I approve of these changes 🤣

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/HostObjectBuilder.cs
```diff
@@ -60,20 +51,28 @@ CancellationToken cancellationToken
...
-          string layerFullName = _autocadLayerManager.LayerFullName(baseLayerPrefix, string.Join("-", path));
+          string layerFullName = GetLayerPath(tc, baseLayerPrefix);
 
           if (uniqueLayerNames.Add(layerFullName))
           {
             _autocadLayerManager.CreateLayerOrPurge(layerFullName);
           }
 
-          object converted = _converter.Convert(obj);
+          //POC: this transaction used to be called in the converter, We've moved it here to unify converter implementation
```
oguzhankoral, line 67: We will reconsider it on Beta phase to refactor AutocadContext with Document and Transactions. Should be safe for now to have it here.

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ToHostConverterWithFallback.cs
oguzhankoral, line 1: I'm a bit confused 🤔 why we have these conversion files into `DependencyInjection` project? There is nothing related to registration or injection

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ToHostConverterWithFallback.cs
```diff
@@ -0,0 +1,70 @@
...
+  /// <returns>The converted host object.
+  /// Fallbacks to display value if a direct conversion is not possible.</returns>
+  /// <remarks>
+  /// The conversion is done in the following order of preference:
+  /// 1. Direct conversion using the <see cref="Speckle.Converters.Common.DependencyInjection.ToHost.ToHostConverterWithoutFallback.TryConvert(Base, out object?)"/> method.
+  /// 2. Fallback to display value using the <see cref="Speckle.Core.Models.Extensions.BaseExtensions.TryGetDisplayValue{T}"/> method, if a direct conversion is not possible.
+  ///
+  /// If the direct conversion is not available and there is no displayValue, a <see cref="System.NotSupportedException"/> is thrown.
+  /// </remarks>
+  /// <exception cref="System.NotSupportedException">Thrown when no conversion is found for <paramref name="target"/>.</exception>
```
oguzhankoral, line 32: Seems like also throws `InvalidOperationException` via private `FallbackToDisplayValue` function

oguzhankoral (changes requested):
Looks good, only change request on maybe the place of the new ToHost converter files. If there is a reason for this I'm ok with it

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ToHostConverterWithFallback.cs
AlanRynne: It does, but that is not the normal execution flow, and devs should not prepare against it (as they cannot recover from it)

If a connector doesn't support fallbacks it shouldn't use the fallback default converter 😁

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ToHostConverterWithFallback.cs
AlanRynne: OH... you, me, Jedd... we're all a bit confused haha We actually discussed this for a good half an hour and tested some alternatives.
We decided to leave it here because of our dependency tree being ever so slightly messy and the name of this project is actually miss-leading (it doesn't do dependency injection either)

TLDR: We needed IFactory in this converter, and adding it to `Speckle.Converters.Common` (it's more logical place to live) would require adding a reference to `Speckle.Autofac` into that (which seemed like breaking a boundary)

There is a unification ticket where a lot of this will be addressed.

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ToHostConverterWithFallback.cs
JR-Morgan: It does, but we don't want callers to handle the `InvalidOperationException`, if a connector registeres his `ToHostConverterWithFallback` but forgets to register a DisplayValue converter...

I'm ok about being a bit selective with which exceptional paths are documented, as we want to craft a clear DX for users of this function, and this failure case is not one.

Something we would want to fail loudly. If a connector doesn't need fall-back, we have the `ToHostconvererWithoutFallback`

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ToHostConverterWithFallback.cs
JR-Morgan: We need `IFactory`, which is defined in `Speckle.Autofac`.

In this case `Speckle.Converters.Common.DependencyInjection` references `Speckle.Autofac`, but not `Autofac`
We did some tidying up of the dependencies to keep `AutoFac` dependency specific to connectors, and the converter modules.

JR-Morgan: We could look at alternatives, maybe we should move `IFactory` somewhere else, or use a different delegate...

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/HostObjectBuilder.cs
JR-Morgan: Thanks for investigating, 
Shall I remove this POC comment?

oguzhankoral: You can edit it by saying: Transaction logic will be moved into AutocadContext.. soemthing like this

PR #3329, opened 2024-04-30
feat(autocad): CNX 9327 autocad polycurve receive separate converters

clairekuang:
## Description & motivation

Completes `AutocadPolycurve` receiving with the following changes:
- Changed `plane` prop on the `AutocadPolycurve` class to `normal` and `elevation`, due to complications on receive (plane retrieved on the original polyline did not accurately capture the normal and elevation properties).
- polyline, polyline2d, and polyline3d are now received as the original type, with original editability

NOTE:
Due to limitations of Autocad's API, it is extremely difficult to guarantee that the `Polycurve` segments will be sequential and correctly oriented (eg, instead of a contiguous segment endpoint to next segment startpoint orientation, a segment may be reversed so that it's endpoint coincides with the previous segment's endpoint).
This results in incorrect rendering of the polycurve in the viewer, and a fix should be discussed with @AlexandruPopovici : 
[https://app.speckle.systems/projects/b53a53697a/models/66aca735ab](https://app.speckle.systems/projects/b53a53697a/models/66aca735ab
)

## Changes:

- Objects
- Autocad Polycurve converters

## To-do before merge:

- [ ] Test Receiving in Rhino for regressions compared to 2.18.3
- [ ] Sync with @AlexandruPopovici on polycurve display issues for segments that are not necessarily in order or in the correct start/end point orientation

## Screenshots:

image


DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/AutocadPolycurveToHostPolylineRawConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+    IConversionContextStack<Document, ADB.UnitsValue> contextStack
+  )
+  {
+    _vectorConverter = vectorConverter;
+    _contextStack = contextStack;
+  }
+
+  public ADB.Polyline RawConvert(SOG.Autocad.AutocadPolycurve target)
+  {
+    if (target.normal is SOG.Vector normal && target.elevation is double elevation)
```
oguzhankoral, line 25: I would throw first with single if block instead if-else. Not a big deal but it is nicer to do them similar as we throw first in conversions

oguzhankoral (approved):
Tested, works super!

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/Extensions/ListExtensions.cs
```diff
@@ -14,4 +14,44 @@ public static SOG.Polyline ConvertToSpecklePolyline(this List<double> pointList,
...
+
+  public static List<AG.Point3d> ConvertToPoint3d(this List<double> pointList, double conversionFactor = 1)
+  {
+    // throw if list is malformed
+    if (pointList.Count % 3 != 0)
+    {
+      throw new System.ArgumentException("Point list of xyz values is malformed.");
+    }
+
+    List<AG.Point3d> points3d = new();
```
JR-Morgan, line 43: I would prefer we presize these lists if they will be of a known size, and have potential to be really really large.
This is for performance reasons (since the list never needs to be dynamically resized)

```cs
List<Ag.Point3d point3d = new(pointList.Count * 3);
```

I'll try and do a round of profiling after the alpha is shipped, and we can identify and fix bottlenecks then if you'd prefer to merge this PR with out this change

oguzhankoral: Good point Jedd, doing it.

oguzhankoral: But I think it is `pointList.Count / 3` not multiple

JR-Morgan (approved):

oguzhankoral:
@clairekuang Realized an issue but this might not be related directly with this PR. When I receive it first time it works but when I open new document and try to receive it again we do not receive all of the polylines
image

oguzhankoral (approved):
Created follow up ticket that I mentioned on comment for this and merging this.
https://spockle.atlassian.net/browse/CNX-9382

DUI3-DX/Converters/Autocad/Speckle.Converters.AutocadShared/Extensions/ListExtensions.cs
JR-Morgan: Yep you're absolutely right, good catch

PR #3332, opened 2024-05-01
fix(revit): exclude element params which are not visible

teocomi:
https://speckle.community/t/transferring-problem/10462

Debugging this community-reported bug, I noticed we're sending to Spekcle also any hidden element parameter; in this case, there are hundreds of huge parameters attached to some elements (maybe images or something else?) added, most likely via an adding.

I think we should ignore these props.

Ideally this PR would be released as hotfix sometime soon.

image

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
@@ -333,6 +333,13 @@ public partial class ConverterRevit
         continue;
       }
 
+      //ignore parameters that are not visible by the end users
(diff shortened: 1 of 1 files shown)


BovineOx (approved):

AlanRynne:
Are we targeting `main` for this, i.e. as a hotfix? Is it worth it if we're already about to release 2.19?

There's a lot of dancing around to keep all our long-running branches up to date ATM so if we could target `dev` for this I'd appreciate it :)

teocomi:
I'll let @bimgeek prioritize

connorivy (approved):

AlanRynne (approved):
Retargeting to `dev` for 2.19 release

PR #3336, opened 2024-05-01
Cnx 9231 revit roof

connorivy:
## Description & motivation

Adds the three different types of roof that our current connector supports. RoofBase, ExtrusionRoof, and FootPRintRoof. 

In our current connector, the footprint roof employs some decently complex logic to extract the slope if it were set by a slope arrow (the slope arrow is not directly accessible via the api). However, the current logic is only supporting the Revit <-> Revit use case. I spent some time trying to make that logic work more generally, but realized that is going to take a bit of time and testing so I just left it as a POC comment for now. All other functionality was moved over.

## Checklist:


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/WallConversionToSpeckle.cs
```diff
@@ -69,10 +69,11 @@ public override SOBR.RevitWall RawConvert(DB.Wall target)
...
     // POC : what to do if these parameters are unset (instead of assigning default)
     speckleWall.height =
-      _parameterValueExtractor.GetValueAsDouble(target, DB.BuiltInParameter.WALL_USER_HEIGHT_PARAM) ?? 0;
+      _parameterValueExtractor.GetValueAsDoubleOrNull(target, DB.BuiltInParameter.WALL_USER_HEIGHT_PARAM) ?? 0;
     speckleWall.baseOffset =
-      _parameterValueExtractor.GetValueAsDouble(target, DB.BuiltInParameter.WALL_BASE_OFFSET) ?? 0;
-    speckleWall.topOffset = _parameterValueExtractor.GetValueAsDouble(target, DB.BuiltInParameter.WALL_TOP_OFFSET) ?? 0;
+      _parameterValueExtractor.GetValueAsDoubleOrNull(target, DB.BuiltInParameter.WALL_BASE_OFFSET) ?? 0;
+    speckleWall.topOffset =
+      _parameterValueExtractor.GetValueAsDoubleOrNull(target, DB.BuiltInParameter.WALL_TOP_OFFSET) ?? 0;
```
AlanRynne, line 76: this seems odd, is a wall of height 0 even valid?

Wouldn't this be better as a `TryGet` pattern where you get both the result and the object? `bool TryGetValueAsDouble(target, ...WALL_TOP_OFFSET , out var theDouble)`

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/HostedElementConversionToSpeckle.cs
```diff
@@ -22,6 +23,22 @@ ISpeckleConverterToSpeckle converter
...
+      {
+        elements.AddRange(hostedObjects);
+      }
+      else
+      {
+        @base.SetDetachedProp("elements", hostedObjects);
+      }
+    }
+  }
+
```
AlanRynne, line 41: I'm starting to dislike this "assign pattern". it's basically a huge side-effect.

It also adds to an existing list of hosted elements, which implies we may be setting the hosted elements of a given object more than once or in parts?

This also makes it more difficult to distinguish between:
- Getting the hosted elements (Revit exclusive operation)
- Converting the hosted elements (Converter operation)
- Assigning the hosted elements (Operation done in the parent element)

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/FootPrintRootConversionToSpeckle.cs
```diff
@@ -0,0 +1,81 @@
...
+    var profiles = _modelCurveArrArrayConverter.RawConvert(target.GetProfiles());
+    speckleFootprintRoof.outline = profiles.FirstOrDefault();
+    speckleFootprintRoof.voids = profiles.Skip(1).ToList<ICurve>();
+
+    var elementType = (ElementType)target.Document.GetElement(target.GetTypeId());
+    speckleFootprintRoof.type = elementType.Name;
+    speckleFootprintRoof.family = elementType.FamilyName;
+
+    // POC: we are starting to see logic that is happening in all converters. We should definitely consider some
+    // conversion pipeline behavior. Would probably require adding interfaces into objects kit
```
AlanRynne, line 74: We could pull this off initially with some basic inheritance, but I agree, we are going to end up needing all the way up to IBase (ISpeckleObject ♥️)

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/FootPrintRootConversionToSpeckle.cs
```diff
@@ -0,0 +1,81 @@
...
+      {
+        level = _levelConverter.RawConvert(baseLevel),
+        cutOffLevel = topLevel is not null ? _levelConverter.RawConvert(topLevel) : null,
+        slope = slope
+      };
+
+    // POC: again with the incorrect assumption that the first profile is the floor and subsequent profiles
+    // are voids
+    // POC: in current connector, we are doing serious gymnastics to get the slope of the floor as defined by
+    // slope arrow. The way we are doing it relys on dynamic props and only works for Revit <-> Revit
```
AlanRynne, line 64: @BovineOx the floor class in Speckle doesn't seem to reflect this reality, hence why we're having to do this.

Should we consider the creation of a new `floor` class that has `profiles` instead of making the asumption that there will only ever be 1 outline and multiple voids?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterValueExtractor.cs
```diff
@@ -98,14 +107,20 @@ or _
...
 
-  public T GetValueAsDocumentObject<T>(Element element, BuiltInParameter builtInParameter)
+  public T? GetValueAsDocumentObjectOrNull<T>(Element element, BuiltInParameter builtInParameter)
     where T : class
   {
     ElementId? elementId = GetValueAsElementId(element, builtInParameter);
     Element paramElement = element.Document.GetElement(elementId);
-    return paramElement as T
+    return paramElement as T;
+  }
```
AlanRynne, line 116: We should consider making this a `TryGet` pattern instead, as this doesn't make the distinction between a value that was in fact null, and a value that failed to cast to `T` but was not null.

```csharp
public bool TryGetValueAsDocumentObject<T>(
    Element element, 
    BuiltInParameter builtInParameter, 
    out T value
)
```

AlanRynne (changes requested):
Some comments, main change I'd make is in the `orNull` parameterValueExtractor functions.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/FootPrintRootConversionToSpeckle.cs
BovineOx: If the assumption is weak, we should probably upgrade it, for the Alpha we can take it as is if this is a migration of current patterns but open a ticket and drop the link here

BovineOx (changes requested):
Apologies - did 90% of this last night and forgot to submit some how

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/RoofBaseConversionToSpeckle.cs
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public override RevitRoof RawConvert(RoofBase target)
+  {
+    RevitRoof revitRoof = new();
+    var elementType = (ElementType)target.Document.GetElement(target.GetTypeId());
+    revitRoof.type = elementType.Name;
+    revitRoof.family = elementType.FamilyName;
+
+    _parameterObjectAssigner.AssignParametersToBase(target, revitRoof);
```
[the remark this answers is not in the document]
AlanRynne: I would say an Abstract `Element` converter with the shared logic that forces implementation of the specifics for each class, but keeps things like `applicationId`, `level`, `parameters` and surely others, unified in a single place.

We should still have the services injected, but the implementation should be in a single place.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/RoofBaseConversionToSpeckle.cs
```diff
@@ -0,0 +1,39 @@
...
+  {
+    _displayValueExtractor = displayValueExtractor;
+    _hostedElementConverter = hostedElementConverter;
+    _parameterObjectAssigner = parameterObjectAssigner;
+  }
+
+  public override RevitRoof RawConvert(RoofBase target)
+  {
+    RevitRoof revitRoof = new();
+    var elementType = (ElementType)target.Document.GetElement(target.GetTypeId());
```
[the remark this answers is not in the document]
AlanRynne: I've just made this change in the Floor PR too.

My understanding is that requesting an element in the document using it's `TypeId` will always return a `DB.ElementType` so this cast is safe. If it explodes by any means it would be an unexpected result.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/HostedElementConversionToSpeckle.cs
```diff
@@ -14,52 +12,41 @@ public class HostedElementConversionToSpeckle
...
-  {
-    var convertedHostedElements = new List<Base>();
-    foreach (var elemId in hostedElementIds)
-    {
-      var element = host.Document.GetElement(elemId);
-      if (_convertedObjectsCache.ContainsBaseConvertedFromId(element.UniqueId))
+      catch (SpeckleConversionException)
       {
+        // POC: logging
         continue;
```
AlanRynne, line 46: SWALLOWING EXCEPTION ⚠️ This is the main reason hosted elements need to be handled on the Connector side instead of within the converter.

In order to successfully convert the wall we need the host element conversion to not-explode, which leads to us defending of exceptions we shouldn't be.

This hosted element should be considered a "converted element" at the top level, and the new traversal assumes this. 

To be honest I would rather fail a Wall conversion entirely rather than tell users "your model sent OK" and then be missing half of it's doors/windows/etc...

AlanRynne (changes requested):
More of a philosophical question around exceptions

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/FootPrintRootConversionToSpeckle.cs
```diff
@@ -0,0 +1,81 @@
...
+    var baseLevel = _parameterValueExtractor.GetValueAsDocumentObject<DB.Level>(
+      target,
+      DB.BuiltInParameter.ROOF_BASE_LEVEL_PARAM
+    );
+    var topLevel = _parameterValueExtractor.GetValueAsDocumentObjectOrNull<DB.Level>(
+      target,
+      DB.BuiltInParameter.ROOF_UPTO_LEVEL_PARAM
+    );
+
+    //NOTE: can be null if the sides have different slopes
```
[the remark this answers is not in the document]
AlanRynne: I'm not sure this is something we can recover from? I.e. if the roof has more than 1 slope we don't really support this use-case.

BovineOx (approved):

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/HostedElementConversionToSpeckle.cs
connorivy: You're right that this function was doing too much. I fixed this to be more SOLID last night

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/HostedElementConversionToSpeckle.cs
connorivy: In my mind, the behavior that we ultimately want is for the wall to convert without the children, but have quality reporting to the user about the failure. If we agree on that, then the behavior that we implement for the alpha doesn't really matter so much. I would be fine with throwing and stopping the entire wall conversion. If we disagree on that, then that is when we have the philosophical conversation

AlanRynne (approved):

PR #3337, opened 2024-05-01
Fix(Revit) : add possible length units

connorivy:
## Description & motivation

Add weird units in Revit. Don't ask me the difference between Meters and MetersCentimeters because I have no clue.

### DIFF
--- a/DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/RevitToSpeckleUnitConverter.cs
+++ b/DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/RevitToSpeckleUnitConverter.cs
@@ -13,8 +13,10 @@ public sealed class RevitToSpeckleUnitConverter : IHostToSpeckleUnitConverter<DB
     _unitMapping[DB.UnitTypeId.Millimeters] = Units.Millimeters;
     _unitMapping[DB.UnitTypeId.Centimeters] = Units.Centimeters;
     _unitMapping[DB.UnitTypeId.Meters] = Units.Meters;
+    _unitMapping[DB.UnitTypeId.MetersCentimeters] = Units.Meters;
     _unitMapping[DB.UnitTypeId.Inches] = Units.Inches;
     _unitMapping[DB.UnitTypeId.Feet] = Units.Feet;
+    _unitMapping[DB.UnitTypeId.FeetFractionalInches] = Units.Feet;
   }
 
   // POC: maybe just convert, it's not a Try method


AlanRynne:
Was this something we were supporting before?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/RevitToSpeckleUnitConverter.cs
```diff
@@ -13,8 +13,10 @@ public RevitToSpeckleUnitConverter()
     _unitMapping[DB.UnitTypeId.Millimeters] = Units.Millimeters;
     _unitMapping[DB.UnitTypeId.Centimeters] = Units.Centimeters;
     _unitMapping[DB.UnitTypeId.Meters] = Units.Meters;
+    _unitMapping[DB.UnitTypeId.MetersCentimeters] = Units.Meters;
     _unitMapping[DB.UnitTypeId.Inches] = Units.Inches;
     _unitMapping[DB.UnitTypeId.Feet] = Units.Feet;
+    _unitMapping[DB.UnitTypeId.FeetFractionalInches] = Units.Feet;
```
AlanRynne, line 19: Interestingly we used to have 2 different unit conversions:

- `UnitsToSpeckle(DisplayUnitType)` did in fact cover `MetersCentemeters`, `FeetFractionalInches`, and also `DecimalFeet` but the type is not `DB.ForgeTypeId`, it's `DisplayUnitType` https://github.com/specklesystems/speckle-sharp/blob/b67e46c0dc3eee8572f4cf9c9cdef339949f536b/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/Units.cs#L94-L118
- `UnitsToSpeckle(string typeId)` which seems to be the somewhat equivalent of this did not have these edge-cases included https://github.com/specklesystems/speckle-sharp/blob/b67e46c0dc3eee8572f4cf9c9cdef339949f536b/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/Units.cs#L94-L118

Have we unified the way we do units? The `ForgeTypeId` seems to be  2021 and above so we may be good with adding these.

We're still missing the `DecimalFeet` and `DecimalInches`and or `FractionalInches` from the original implementation of DisplayUnit input if we want to be 100% covered.

AlanRynne (changes requested):
I think we may want to add some that are missing from the original implementation so we're aligned (or at least we'd want to check)

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/RevitToSpeckleUnitConverter.cs
connorivy: We were still missing the fractional inches so I added that one. It seems like decimal feet and inches don't have an equivalent in the new ForgeTypeId

PR #3338, opened 2024-05-02
Cnx 9229 revit topography

BovineOx:
Added in Topography TopLevel conversion


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/BaseConversionToSpeckle.cs
```diff
@@ -5,14 +5,19 @@
 namespace Speckle.Converters.RevitShared.ToSpeckle;
 
 // POC: maybe but could be generic abstract Converters.Common?
+// or maybe it's not actually doing very much now and can come out
```
AlanRynne, line 8: We've managed to unify the ToHost converter so it could be possible to unify how the top-level converter works for everyone.

Would need a bit of investigation around the current implementations.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/BaseConversionToSpeckle.cs
```diff
@@ -5,14 +5,19 @@
...
     IRawConversion<THost, TSpeckle>
   where TSpeckle : Base
 {
   public Base Convert(object target)
   {
     var result = RawConvert((THost)target);
+
+    // POC: unless I am going bonkers, we've constrained TSpeckle to Base
+    // so it should always BE base?
     if (result is not Base @base)
```
AlanRynne, line 21: Oh yeah... not sure how we'll ever get here 🤔

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/WallConversionToSpeckle.cs
```diff
@@ -77,11 +77,13 @@ public override SOBR.RevitWall RawConvert(DB.Wall target)
...
     speckleWall.flipped = target.Flipped;
 
+    // POC: shouldn't we just do this in the RevitConverterToSpeckle ?
     speckleWall.displayValue = _displayValueExtractor.GetDisplayValue(target);
 
     AssignVoids(target, speckleWall);
     AssignHostedElements(target, speckleWall);
 
+    // POC: shouldn't we just do this in the RevitConverterToSpeckle ?
     _parameterObjectAssigner.AssignParametersToBase(target, speckleWall);
```
AlanRynne, line 87: Not a blocker but, I would advocate for a parent converter class (or several as needed) that performs this consistently for every `Element` in Revit (they all have parameters, levels, location, should have displayValue, should have applicationId...)

This would reduce the amount of duplicated registration needed on each individual converter, just for all to do the same things (parameterAssigner, levelConverter...)

Of course every operation needs to be individually injectable in case of re-use, but we should have a consistent way of building up "just the Topography logic" without the extra boilerplate.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/TopographyConversionToSpeckle.cs
```diff
@@ -0,0 +1,54 @@
...
+  public override SOBR.RevitTopography RawConvert(DBA.TopographySurface target)
+  {
+    var speckleTopo = new SOBR.RevitTopography();
+    speckleTopo.displayValue = _displayValueExtractor.GetDisplayValue(target);
+
+    // POC: shouldn't we just do this in the RevitConverterToSpeckle ?
+    _parameterObjectAssigner.AssignParametersToBase(target, speckleTopo);
+
+    return speckleTopo;
+  }
```
AlanRynne, line 53: RevitTopography has an `elementId` property, where we not setting that before?

If not, let's keep it that way and we could flag it as Obsolete in the near future.

AlanRynne (changes requested):
Just one blocker regarding `RevitTopography` having an `elementId` property that we're not currently setting. If this was the same way in the old connector, I would not consider this a blocker.

Some other comments for posterity.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/BaseConversionToSpeckle.cs
BovineOx: We have the POC comment but the whole class seems a bit bonkers now

AlanRynne (approved):

PR #3341, opened 2024-05-02
DUI3: Highlight logic simplication

didimitrie:
Brings back implementation from dui3/dev, basically + fixes for empty model cards (ie, nothing to select case)


AlanRynne (approved):
Less code seems better to me 👍🏻 not going to complain 🤐🥳

PR #3342, opened 2024-05-02
Fixes Revit empty sends + adds error handling logic in `HandleSend`

didimitrie:
There is a semi implemented "SpeckleTopLevelExceptionHandler" that wraps the call of this HandleSend, but it currently is not doing anything - resulting in all errors from here bubbling up to the bridge, which is not kosher.

Unsure what the intentions are around the SpeckleTopLevelExceptionHandler going forward.


AlanRynne (approved):
Nice! This should actually have been here the whole time. 🙇🏼‍♂️

PR #3344, opened 2024-05-02
CNX-9397 CNX-9401 rhino does not receive autocad polylines

JR-Morgan:
I've moved @connorivy's inheritance search behaviour from the Revit `ToSpeckle` converter, into its own service `RecursiveConverterResolver<TConverter>`
and made this generic to both ToSpeckle and ToHost conversions.

Registering this allows us to resolve converters for super classes if the concrete class has no conversion (e.g. `AutocadPolycurve`)

---

Feedback welcome, this change has highlighted the `...ToHostWithFallback` class is now very very simple... perhaps too simple, it literally just gets a converter from the converter resolver, and calls convert on it... not something you'd usually want a service for. I fear we may be loosing some good code locality with such a simple behaviour spread out over many files.

I think we could do a roundup of some of our constructs post alpha. We have several very similarly named classes with not super intuitively defined responsibilities, with scattered locality, and awkward long names


AlanRynne (approved):
Looks good! Will give this a test throughout the morning and get it merged if I find no issues

DUI3-DX/Sdk/Speckle.Converters.Common/RecursiveConveterResolver.cs
```diff
@@ -0,0 +1,27 @@
...
+
+  public TConverter? GetConversionForType(Type objectType)
+  {
+    if (objectType.BaseType == null)
+    {
+      //We've reached the top of the inheritance tree
+      return null;
+    }
+
+    return _factory.ResolveInstance(objectType.Name) ?? GetConversionForType(objectType.BaseType);
```
BovineOx, line 25: I see this is a layer on top of the factory. Seems fine. Is this pattern in Revit at all yet? I guess this is on the receive side mind

BovineOx (approved):
Minor comments but I think I get it

DUI3-DX/Sdk/Speckle.Converters.Common/RecursiveConveterResolver.cs
JR-Morgan: This pattern was originally in Revit for ToSpeckle.
This PR moves it to a shared place, and can be used for both send and receive.

Right now, just used in Revit for Send, and Rhino & Autocad for Receive

connorivy (approved):
Looks good to me. The only comment I would make is that it may be helpful to control the type at which the recursion breaks. For example, if no converter is found while receiving until we get all the way up to Base, we probably want to notify the user that we didn't find a valid converter and are converting it more generically. This is not a blocker, just something that I think we may need in the future. I'm happy to merge this now and then add that functionality if we ever actually do need it.

PR #3348, opened 2024-05-03
CNX-9365 model cards not saved to document rvt

oguzhankoral:
Fixing multiple problems.

- First and most important one was wrong function that passed into every `BrowserBridge`. We must use alwasys Dispacthed version of `ExecuteScriptAsync`
- We do not necessarily need to run `Document` related operations under `RevitTask`. We just run it sync since we do not have anything with `UIDocument`
-  Previously there were wrong references injected into `RevitBaseBinding`. Now we create `RevitDocumentStore` as `DocumentModelStore`
- Document switch not even handled before I believe. Now we read/write model cards on doc switch by doing some magical checks on `ViewChanged` event handler..

Related to:
- https://spockle.atlassian.net/browse/CNX-9365
- https://spockle.atlassian.net/browse/CNX-9379
- https://spockle.atlassian.net/browse/CNX-9335


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitDocumentStore.cs
```diff
@@ -19,74 +21,98 @@ internal class RevitDocumentStore : DocumentModelStore
...
   {
     _revitContext = revitContext;
+    _documentModelStorageSchema = documentModelStorageSchema;
+    _idStorageSchema = idStorageSchema;
 
     UIApplication uiApplication = _revitContext.UIApplication;
+    uiApplication.ApplicationClosing += (_, _) => WriteToFile(); // POC: Not sure why we would need it since we have save and clos events
+    uiApplication.Application.DocumentSaving += (_, _) => WriteToFile();
+    uiApplication.Application.DocumentSavingAs += (_, _) => WriteToFile();
+    uiApplication.Application.DocumentSynchronizingWithCentral += (_, _) => WriteToFile(); // POC: Not sure why we have it
```
oguzhankoral, line 43: @didimitrie I don't why we need this, probably you added this one onto dev branch and I just left it there since doesn't have any side effect on aynthing (YET)

oguzhankoral: do we need it now? should I remove?

AlanRynne (approved):
Tested on Debug and works 🚀🥳

BovineOx (approved):
Seems sensible but I'd probably be tempted to interface some other pieces out but not a blocker for right now

PR #3351, opened 2024-05-03
sync to ui thread

JR-Morgan:
Previously, we were never capturing the exception within the sync to UI thread
This was causing some fatal crashes on cancellation

This PR fixes this, by ensuring the exception is captured as part of the task completion source


didimitrie (approved):

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bridge/SyncToUIThread.cs
```diff
@@ -19,8 +20,15 @@ public Task<T> RunOnThread<T>(Func<T> func)
...
     _bridge.RunOnMainThread(() =>
     {
-      T result = func.Invoke();
-      tcs.SetResult(result);
+      try
+      {
+        T result = func.Invoke();
+        tcs.SetResult(result);
+      }
+      catch (Exception ex)
```
didimitrie, line 28: Blocker warning/ci error here, but imho valid place to overrule it, as it's doing the right thing (as far as I understood it!)

BovineOx: seems legit

BovineOx (approved):

PR #3352, opened 2024-05-03
RVT: Commit structure hotfix

didimitrie:
Stores elements as
Root (collection)
- Document (collection)
  - Level (collection)
    - Category (collection)
      - Obj

Note: this is an edited paste of the old POC from the dev branch

image


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
```diff
@@ -42,26 +42,87 @@ RevitConversionContextStack contextStack
       throw new InvalidOperationException("No objects were found. Please update your send filter!");
     }
 
-    Base commitObject = new();
+    var doc = _contextStack.Current.Document.Document; // POC: Document.Document is funny
+    var rootObject = new Collection() { name = doc.PathName.Split('\\').Reverse().First().Split('.').First() };
```
BovineOx, line 46: readability I think Last() is clearer and given split returns an array it is also faster, unless I am misreading what this is doing

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
```diff
@@ -42,26 +42,87 @@ RevitConversionContextStack contextStack
...
+  /// <param name="cache"></param>
+  /// <param name="root"></param>
+  /// <returns></returns>
+  private Collection GetAndCreateObjectHostCollection(
+    string[] path,
+    Dictionary<string, Collection> cache,
+    Collection root
+  )
+  {
+    string fullPathName = string.Join("", path);
```
BovineOx, line 99: I think you want string.concat() I was going to be old school and say string.Empty is better but it's actually no different these days as "" is given over as shorthand for string.Empty

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
```diff
@@ -42,26 +42,87 @@ RevitConversionContextStack contextStack
...
+    if (cache.TryGetValue(fullPathName, out Collection value))
```
BovineOx, line 100: So the key is the logical path to this element and this should always be unique? I ask for completeness as it doesn't feel like it _could_ be colliding?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
```diff
@@ -42,26 +42,87 @@ RevitConversionContextStack contextStack
...
+    foreach (Element revitElement in objects)
     {
       ct.ThrowIfCancellationRequested();
-      if (_convertedObjectsCache.ContainsBaseConvertedFromId(obj.UniqueId))
+
+      var cat = revitElement.Category.Name;
+      var path = new[] { doc.GetElement(revitElement.LevelId) is not Level level ? "No level" : level.Name, cat };
+      var collection = GetAndCreateObjectHostCollection(path, collectionCache, rootObject);
+
+      if (_convertedObjectsCache.ContainsBaseConvertedFromId(revitElement.UniqueId))
```
BovineOx, line 62: Have we confirmed the lifetime so that different operations, should it be possible for them to be concurrent, won't collide their cache? Or is does the key include the file or cloud location?

BovineOx (changes requested):
Couple of minor requests perhaps and a few qs but otherwise seems fine

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
didimitrie: It could and it should, hence why we're getting it out of the cache below. Basically this whole function allows me to decouple collection generation and nesting from the conversion loop above and keep things "simpler"

in the conversion loop for each element we can just say gimme my collection host on whatever pattern we want to (level + category right now) and then we add the conversion result in the provided collection host. 

I feel like this is not a very eloquent explanation, but the net result is we're bundling objects that should be together in the right collection :D

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
didimitrie: Confirmed by voice, and this will allow me to do some extra cleanup now!

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
```diff
@@ -42,26 +51,76 @@ RevitConversionContextStack contextStack
...
+        collection.elements.Add(convertedElement);
       }
       catch (SpeckleConversionException ex)
       {
         // POC: logging
       }
+
+      countProgress++;
+      onOperationProgressed?.Invoke("Converting", (double)countProgress / objects.Count);
+      Thread.Sleep(1000);
```
oguzhankoral, line 83: Forgetten?

oguzhankoral (changes requested):

BovineOx (approved):

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RootObjectBuilder.cs
didimitrie: FML, yes!

oguzhankoral (approved):

PR #3354, opened 2024-05-03
CNX-9366 [RVT] Object Tracking not working

didimitrie:
centralises id operations across on element ids. note, it would be better do to do it the other way around, but the change events in revit give us plain ids rather than unique ids.

The underlying issue was that we were using element ids (coming from change events in revit) and for the rest we were using the unique ids of elements. 

What I am not fully aware of is whether relying on the element ids rather than the unique ids will cause future problems down the line regarding selecting objects from linked models. This though currently seems a far off problem.

## Checklist:


connorivy (approved):
Tested and works well. I even tried to trick the change tracker by modifying the document via the api instead of the frontend, and it still detected the changes.

If we ever run into conflicts with linked documents, we could always switch to composite keys using the document id or something like that.

PR #3355, opened 2024-05-03
Fix issue with leaving rhino in an undrawing state if an exception (including cancellation) occurs during conversion

JR-Morgan:
Pretty self explanatory this PR, building off https://github.com/specklesystems/speckle-sharp/pull/3353

---

However, while investigating this problem, it raises the question of how we should be handling cancellation and the order of baking in Rhino

In DUI2, did we not convert everything first, then bake?

This was good, because we could cancel anytime up to the point where we bake the first object... then no more cancellation as were too far into the receive.
Do we not want to bring a similar behaviour to DUI3...

Right now, when you cancel we give up, and leave however many objects are already baked just there loose in the users document. To most users cancel means I've changed my mind, I don't want to receive this... And if we can't honour the users request to cancel, we do what we do in DUI2 and say "Receive completed, it was too late to cancel"

### DIFF
--- a/DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/DisableRedrawScope.cs
+++ b/DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/DisableRedrawScope.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using Rhino.DocObjects.Tables;
+
+namespace Speckle.Connectors.Rhino7.Operations.Receive;
+
+/// <summary>
+/// Helper class to disable <see cref="ViewTable.RedrawEnabled"/> within a scope
+/// </summary>
+public sealed class DisableRedrawScope : IDisposable
+{
+  private readonly ViewTable _viewTable;
+  private readonly bool _returnToStatus;
+
+  public DisableRedrawScope(ViewTable viewTable, bool returnToStatus = true)
+  {
+    _viewTable = viewTable;
+    _returnToStatus = returnToStatus;
+
+    _viewTable.RedrawEnabled = false;
+  }
+
+  public void Dispose()
+  {
+    _viewTable.RedrawEnabled = _returnToStatus;
+  }
+}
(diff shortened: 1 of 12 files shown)


didimitrie:
I'd allow for cancellation any time within reason. Bakes can take time too. As well, we should throw if cancellation is requested from places we know we can wind back from - so if we're deep in a transaction, maybe we shouldn't (thinking of revit now).

To your question: i believe for now we can live with an impartial bake. In the future, i'd tear down gracefully whatever's been baked up to the point of cancellation. 

PS: In rhino, it's as simple as tearing down the root layer (until we get blocks poc in)

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/DisableRedrawScope.cs
```diff
@@ -0,0 +1,26 @@
...
+    _viewTable = viewTable;
+    _returnToStatus = returnToStatus;
+
+    _viewTable.RedrawEnabled = false;
+  }
+
+  public void Dispose()
+  {
+    _viewTable.RedrawEnabled = _returnToStatus;
+  }
```
AlanRynne, line 25: Do we need return to status if this class explicitly only Disables the Redraw scope? (i.e. it won't ever switch it "on", making the need of returning to the prior status somewhat redundant?)

AlanRynne (approved):
I like this contraption. Makes it very difficult (if not impossible) to forget to re-enable redraw after needing to disable it ♥️

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/DisableRedrawScope.cs
JR-Morgan: No, I was thinking that, if we start using this in a nested way, you'd actually want to disable, then restore to what ever the status was before the scoped block, that would prevent us from ever accidentally enabling on dispose when we should actually stay disabled.

I thought to kepp it simple for this pr, do that parameter is always true

If you think what I describe is not useful, we can simply remove it from this class

AlanRynne: No this is fine. I was just wondering since you weren't using it currently; but I agree it is a good idea to have it.

PR #3357, opened 2024-05-03
CNX-9408 - Fix(Revit): silence exception throw by duplicate conversions

connorivy:
## Description & motivation

Sometimes multiple objects in revit can host the same sub element. In our current connector, we are just converting these items multiple times. However, in this connector an exception is being thrown. For now, I am just silencing this exception until we can address the issue more completely https://spockle.atlassian.net/browse/CNX-9402.

I also made a change to make the "TryGetValueAsDocumentObject" safer

## Checklist:

### DIFF
--- a/DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
+++ b/DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
@@ -45,7 +45,17 @@ public class RevitConverterToSpeckle : ISpeckleConverterToSpeckle
       // POC: is this the right place?
       result.applicationId = element.UniqueId;
 
-      _convertedObjectsCache.AddConvertedBase(element.UniqueId, result);
+      try
+      {
+        _convertedObjectsCache.AddConvertedBase(element.UniqueId, result);
+      }
+      catch (ArgumentException)
+      {
+        // POC: object converted multiple times
+        // we are doing this all the time in our current converter, and the serializer is fixing it for us.
+        // so for now, I am just silencing this exception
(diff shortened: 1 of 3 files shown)


JR-Morgan (approved):

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitConverterToSpeckle.cs
```diff
@@ -45,7 +45,17 @@ public Base Convert(object target)
...
+      {
+        _convertedObjectsCache.AddConvertedBase(element.UniqueId, result);
+      }
+      catch (ArgumentException)
+      {
+        // POC: object converted multiple times
+        // we are doing this all the time in our current converter, and the serializer is fixing it for us.
+        // so for now, I am just silencing this exception
+        // https://spockle.atlassian.net/browse/CNX-9402
+      }
```
AlanRynne, line 58: Why aren't we just coding the cache to not explode?

Something like `HasConvertedBase()` or refactor `AddConvertedBase()` to return true if it succeeds to add the object to the cache?

We're also using `Add` in the cache super naively, assuming an object will never be set beforehand, but this PR's existence means that this assumption is not correct.

I would argue this is not the right fix, and instead the cache should be made less easy to explode (it is a cache after all, which should be a utility thing that doesn't really need to exist to perform the main logic correctly)

AlanRynne (changes requested):
Valid solution, but I think it's being done in the wrong place (see comment). Happy to discuss more.

AlanRynne (approved):
Discussed with Ian and Jedd. We're merging this in and adding a ticket to re-visit hosted element conversion and caching.

PR #3364, opened 2024-05-06
CNX-9410 Revit caching not working

JR-Morgan:
Revit change tracking was never fully implemented

1. Updated Revit to use the new shared `RootObjectSender` (and deleted the old implementation in the Revit project)
2. Copied the object caching code from Rhino
3. Updated the bindings to use the new method signatures (send info)


oguzhankoral (approved):
Draft version LGTM

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SendBinding.cs
```diff
@@ -83,20 +92,40 @@ private async Task HandleSend(string modelCardId)
...
           (status, progress) => OnSendOperationProgress(modelCardId, status, progress),
           cts.Token
         )
         .ConfigureAwait(false);
 
-      Commands.SetModelCreatedVersionId(modelCardId, versionId);
+      // Store the converted references in memory for future send operations, overwriting the existing values for the given application id.
+      foreach (var kvp in sendResult.convertedReferences)
+      {
+        _convertedObjectReferences[kvp.Key + modelCard.ProjectId] = kvp.Value;
```
connorivy, line 122: we may want to see if we can get some logging showing the size of this dictionary since we are always adding items and never removing any.

PR #3365, opened 2024-05-06
fix: Save to file even when no models exist (covers edge-case where last model was removed)

AlanRynne:
Saving to file should be done even when no models exist.


oguzhankoral (approved):
That's my bad before, ty for catch!

PR #3371, opened 2024-05-07
DO NOT MERGE! Github Actions build + annotations test

AlanRynne:
This is just a PR to test a quick Github actions build pipeline I created last night.

we shouldn't actually merge this, it's just a POC. I need the PR to verify PR annotations are coming through and check if the result provides better DX for the team.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -25,6 +25,8 @@ jobs:
       #   run: dotnet tool restore
       # - name: Code formatting check
       #   run: dotnet csharpier --check .
+      - name: Restore projects
+        run: dotnet restore DUI3-DX.slnf
       - name: Build
         run: msbuild DUI3-DX.slnf /p:Configuration=Release /p:IsDesktopBuild=false
       - name: Upload artifacts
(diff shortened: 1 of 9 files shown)


.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v2
+      # - name: Restore tools
+      #   run: dotnet tool restore
+      # - name: Code formatting check
+      #   run: dotnet csharpier --check .
+      - name: Restore projects
+        run: dotnet restore DUI3-DX.slnf
+      - name: Build
+        run: msbuild DUI3-DX.slnf /p:Configuration=Release /p:IsDesktopBuild=false
```
adamhathcock, line 31: not sure if this helps but using bullseye and C# instead of msbuild is what I do: https://github.com/adamralph/bullseye

AlanRynne: Nice! I've been eyeing some build systems for a while but didn't know about this one. Adding it to my "must look into" list!

adamhathcock: I like that the build is the same language as the project and not XML.

I used to use (and contribute) to Cake but thought it was overblown.  Found bullseye and used it ever since dotnet got simple and fast

AlanRynne:
Closing to reduce noise, @adamhathcock has his branch ongoing and we'll see what we do with this once we reach somewhere meaningful.

PR #3378, opened 2024-05-07
fix(autocad): polyline3d splines had incorrect displayvalue

clairekuang:
Fixes the polyline3d spline display value.
image


oguzhankoral (approved):
LGTM

PR #3380, opened 2024-05-08
Document.Document & IRevitConversionContextStack

BovineOx:
Fixed Document.Document and also changed to always reference stack to always use IRevitConversionContextStack

We were using the UIDocument in the context and NO conversions should care about this document. I've changed this to inject Document instead of UIDocument so we don't have the somewhat silly Document.Document everywhere.

Additionally we added the IRevitConversionContextStack interface some time back to allow conversions to reference this interface instead of the concrete Revit type. I've updated all references and registrations now so they are correct and we now MUST reference IRevitConversionContextStack.


oguzhankoral (commented):
@BovineOx tested with Sample Revit House and it worked as expected

BovineOx:
> @BovineOx tested with Sample Revit House and it worked as expected

thanks @oguzhankoral

AlanRynne (approved):
LGTM 🙌🏼 🚀

PR #3387, opened 2024-05-09
use Microsoft.Build.Tasks.Core.dll

adamhathcock:
Stackoverflow says: https://stackoverflow.com/questions/46741633/jetbrains-rider-the-task-factory-codetaskfactory-could-not-be-loaded-from-th

image


AlanRynne:
Closing this as the issue was resolved by telling the IDE to build with `msbuild` instead of `dotnet`.

The change **is valid** though 🙌🏼 But it's affecting projects of the old DUI2 we're already moving away from, so I wouldn't be too precious of this right now.

As I understand this (made some test with this on ArcGIS last night) just changing the `dll` to use the `Core` one is not enough, because the version `dotnet` ships is different than the one from `msbuild` and requires the use of `RoslynCodeTaskFactory` instead of `CodeTaskFactory` class.

We'll do a full sweep of this when we get there, but most likely these older projects would be excluded from it.

PR #3390, opened 2024-05-09
Fix(dui3): host application name of revit

oguzhankoral:


AlanRynne (approved):
Great catch! Also agree on the extra POC, these things need a home.

PR #3394, opened 2024-05-10
CNX-9199 implement opening revit and dui3 when double clicking a file

oguzhankoral:
I have couple of findings on double-click file open issue,

### Problems
1. Docable pane didn't register successfully if we try to do registration after `RevitTask.Initialize(uiApplication)`
2. Cef binding registirations didn't catch if we do it before `RevitTask.Initialize(uiApplication)` (causes black screen of UI)
3. There are no Revit event that we can trigger `ReadFromFile` and `OnDocumentChanged` functions after double-click open

image

### Solutions
1. Creating and registration of ribbons, buttons and panes done as first pass
2. Cef binding registrations done in `PostApplicationInit` function
3. `ReadFromFile` and `OnDocumentChanged` called in the constructor on `RevitDocumentStore`. Even if I don't like it, I couldn't find any other event or way for now

### Tests
- [x] Double-click file opens the pane
- [x] After double click open we can read the saved model cards
- [x] Non-double-click file opens work as before (switching docs, opening new docs etc..)

Comments welcome!

### DIFF
--- a/DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitDocumentStore.cs
+++ b/DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitDocumentStore.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using Autodesk.Revit.DB;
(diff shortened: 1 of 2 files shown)


connorivy (approved):
Seems to work perfectly

PR #3405, opened 2024-05-13
suppress warning dialog for potential dll conflicts

connorivy:
## Description & motivation

We've decided that the warning message for dll conflicts may not be necessary because we can rely on only notifying the user when a conflict happens such that an exception is thrown and the Speckle connector cannot be used.


AlanRynne (approved):
No objections. So we will trigger the conflict resolution pop-up if an exception reaches the top-level exception handler now?

PR #3407, opened 2024-05-13
Cnx 9487 persist model card state to host file on any change of model card state

didimitrie:
Saves state (model cards) to host app file on add/remove/edit, rather than on document save and co. events. This creates a bit more predictable behaviour as swapping through document tabs (revit, acad) no longer trips up the state. 

Implementation notes: 

Made the `Models` prop an observable collection so we get access to the famous `CollectionChanged` event, to which we reliably subscribe with a handler that, if the host app says "save everytime", we `WriteToFile` everytime the collection is changed.  Removed any onSave handlers that we had in the past as they were no longer relevant - we basically always have an up to date state. 

Tested in Acad, Rhino and Revit. ArcGIS was missing an actual document model store so it's only scaffolded there. I have tested this in Rhino with a 100k object selection sender, and no performance issues, but didn't do the same in Revit/Acad. Assuming that writing to the host file is fast there as well; otherwise we could move the actual writing to an on application idle sub. 

Other unrelated changes: 
- In revit, the write to file was not wrapped in a proper transaction, so i did that
- Removed the inheritance of DocumentModelStore from DiscriminatedObject as it was not needed, we serialse the models prop only


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitDocumentStore.cs
```diff
@@ -66,40 +69,39 @@ private void OnViewActivated(object sender, ViewActivatedEventArgs e)
...
-    ReadFromFile();
-    OnDocumentChanged();
+    _idleManager.SubscribeToIdle(() =>
+    {
+      ReadFromFile();
+      OnDocumentChanged();
+    });
   }
 
   private void WriteToFileWithDoc(Document doc)
```
oguzhankoral, line 80: Seems like this function is useless anymore. It was there before bc we were saving on doc switch with previous doc

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
```diff
@@ -35,8 +35,8 @@ protected override void Load(ContainerBuilder builder)
     RegisterLoggerFactory(builder);
 
     // Register instances initialised by Rhino
-    builder.RegisterInstance<PlugIn>(SpeckleConnectorsRhino7Plugin.Instance);
-    builder.RegisterInstance<Command>(SpeckleConnectorsRhino7Command.Instance);
+    builder.RegisterInstance<PlugIn>(SpeckleConnectorsRhino7Plugin.Instance).SingleInstance();
```
oguzhankoral, line 38: Did it fix something else?

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/AutofacAutocadModule.cs
```diff
@@ -43,7 +43,11 @@ protected override void Load(ContainerBuilder builder)
     builder.RegisterType<AutocadPlugin>().As<IAutocadPlugin>().SingleInstance();
     builder.RegisterType<TransactionContext>().InstancePerDependency();
     builder.RegisterInstance<AutocadDocumentManager>(new AutocadDocumentManager()); // TODO: Dependent to TransactionContext, can be moved to AutocadContext
-    builder.RegisterType<AutocadDocumentStore>().As<DocumentModelStore>().SingleInstance();
+    builder
+      .RegisterType<AutocadDocumentStore>()
+      .As<DocumentModelStore>()
+      .SingleInstance()
+      .WithParameter("writeToFileOnChange", true);
```
oguzhankoral, line 50: I don't have a badge to judge this :)

oguzhankoral (commented):
Few comments/questions. Looks good to me code wise, but will test revit and acad then approve

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
didimitrie: actually no, i should revert this change. i was looking into this https://spockle.atlassian.net/jira/software/c/projects/CNX/boards/5?selectedIssue=CNX-9509

oguzhankoral (approved):

PR #3411, opened 2024-05-13
Chore(Revit): remove batch uploader from running in the background

connorivy:
## Description & motivation

Batch uploader was never fully viable and should not be running in the connector because it doesn't do anything


BovineOx (approved):
@JR-Morgan should have a sanity check but looks legit to me - we ought to have a ticket on the board though

JR-Morgan (approved):

AlanRynne (dismissed):
If this fully disables BatchUploader, we should also pull the references from all the revit csproj files.

JR-Morgan (changes requested):
Good catch Alan

AlanRynne:
Tested this locally and nothing breaks, managed to send/receive just fine without the BatchUploader as expected.

JR-Morgan (approved):

PR #3412, opened 2024-05-13
CNX-9499 Save model cards into metadata and read/write on map init/switch

oguzhankoral:
The way we can store metadata is writing it as XML. I do read/write metadata with `SpeckleModelCards` element into root metadata XML.

Tested and seems work fine but would be nice double-triple tests.

### DIFF
--- a/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
+++ b/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
@@ -61,13 +61,32 @@ public class ArcGISDocumentStore : DocumentModelStore
 
   private void WriteToFileWithMap(Map map)
   {
-    string serializedModels = Serialize();
-
-    XDocument xmlDocument = new(new XElement("metadata", new XElement("SpeckleModelCards", serializedModels)));
-
     QueuedTask.Run(() =>
     {
-      map.SetMetadata(xmlDocument.ToString());
+      // Read existing metadata - To prevent messing existing metadata. 🤞 Hope other add-in developers will do same :D
+      var existingMetadata = map.GetMetadata();
+
+      // Parse existing metadata
+      XDocument existingXmlDocument = !string.IsNullOrEmpty(existingMetadata)
+        ? XDocument.Parse(existingMetadata)
+        : new XDocument(new XElement("metadata"));
+
+      string serializedModels = Serialize();
+
+      XElement xmlModelCards = new("SpeckleModelCards", serializedModels);
+
+      // Check if SpeckleModelCards element already exists at root and update it
+      var speckleModelCardsElement = existingXmlDocument.Root?.Element("SpeckleModelCards");
+      if (speckleModelCardsElement != null)
+      {
+        speckleModelCardsElement.ReplaceWith(xmlModelCards);
+      }
+      else
+      {
+        existingXmlDocument.Root?.Add(xmlModelCards);
+      }
+
+      map.SetMetadata(existingXmlDocument.ToString());
     });
   }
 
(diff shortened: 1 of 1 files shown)


AlanRynne (commented):
Code-wise looks good. I don't have ArcGIS to test so I'll leave the approval to someone who can ;)

KatKatKateryna:
Looks good! Just a question: do other connectors also save model cards onClose event? Meaning if user messed something up and didn't intent to save the file?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
```diff
@@ -9,22 +14,81 @@ public class ArcGISDocumentStore : DocumentModelStore
...
+  private void WriteToFileWithMap(Map map)
   {
-    // Implement the logic to save it to file
+    string serializedModels = Serialize();
+
+    XDocument xmlDocument = new(new XElement("metadata", new XElement("SpeckleModelCards", serializedModels)));
+
+    QueuedTask.Run(() =>
+    {
+      map.SetMetadata(xmlDocument.ToString());
```
KatKatKateryna, line 70: I have noticed this is overwriting existing file metadata that might look like this: 
image

Maybe we can append new tag to the existing metadata?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
```diff
@@ -9,22 +14,81 @@ public class ArcGISDocumentStore : DocumentModelStore
...
   }
 
+  private Task OnProjectClosing(ProjectClosingEventArgs arg)
+  {
+    if (MapView.Active is null)
+    {
+      return Task.CompletedTask;
+    }
+
+    WriteToFile();
```
KatKatKateryna, line 29: Just a question: do other connectors also save model cards onClose event? Meaning if user messed something up and didn't intent to save the file?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
```diff
@@ -9,22 +14,81 @@ public class ArcGISDocumentStore : DocumentModelStore
...
+      var root = XDocument.Parse(metadata).Root;
+      var element = root?.Element("SpeckleModelCards");
+      if (element is null)
+      {
+        Models = new List<ModelCard>();
+        return;
+      }
+
+      string modelsString = element.Value;
+      Models = Deserialize(modelsString);
```
KatKatKateryna, line 91: Opening file without model cards results in an old model card stuck there with an error
image

KatKatKateryna (changes requested):
Looks good!

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
oguzhankoral: Good point, let me fix it!

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
oguzhankoral: We save this info even on doc switch. But this is smth we will reconsider later by just saving model cards right after they created instead on save or on close

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ArcGisDocumentStore.cs
oguzhankoral: Hmm, I would like to check with you

KatKatKateryna (approved):

PR #3416, opened 2024-05-14
add revit 2025 support to the .sln in order for installer to work

connorivy:
## Description & motivation

These projects were already added to All.sln and the corrosponding Revit solution filter, but the installer still assumes we are using the old .sln file, so we need to add the Revit 2025 projects to that file as well


AlanRynne:
All tests are passing now. We still have a ref to xUnitRevit projects, which are not part of the mono-repo.

You must have it checked out in the right place. You were also missing the `TestGenerator` project in the Revit solution.

I must acknowledge this is a PITA 🙇🏼‍♂️

Lets hope it's not for long 🤞🏻

AlanRynne (approved):
Approved, but this should go into main ;) Swapping the target branch now

JR-Morgan (approved):

PR #3417, opened 2024-05-15
CNX-9433 Remove duplication on the Connector Autofac modules again

adamhathcock:
CNX-9433

replacement for https://github.com/specklesystems/speckle-sharp/pull/3384 to have file moves


adamhathcock:
Tested with Rhino, Revit and AutoCAD to see if I could send and receive something.  After initial fixes, things worked

DUI3-DX/Sdk/Speckle.Autofac/DependencyInjection/SpeckleContainerBuilder.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    public ModuleAdapter(ISpeckleModule speckleModule)
+    {
+      _speckleModule = speckleModule;
+    }
+
+    protected override void Load(ContainerBuilder builder) => _speckleModule.Load(CreateInstanceForLoading(builder));
+  }
+
+  public SpeckleContainerBuilder AddModule(ISpeckleModule module)
```
JR-Morgan, line 104: What was the purpose of creating all these `AddModule`, `AddSingleton`, `AddTransient` wrappers?
Why not simply have callers call `ContainerBuilder.RegisterInstance`? 

Seems like we'll have to duplicate a lot of autofac functions and I'm not sure I understand the purpose of this?

DUI3-DX/Sdk/Speckle.Autofac/EnumerationExtensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Speckle.Autofac;
+
+public static class EnumerationExtensions
+{
+  public static void ForEach<T>(this IEnumerable<T> items, Action<T> action)
```
JR-Morgan, line 5: Doesn't linq already have effectively this?
https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.foreach?view=netstandard-2.0

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/HostApp/RhinoDocumentStore.cs
```diff
@@ -9,8 +9,8 @@ public class RhinoDocumentStore : DocumentModelStore
   private const string SPECKLE_KEY = "Speckle_DUI3";
   public override bool IsDocumentInit { get; set; } = true; // Note: because of rhino implementation details regarding expiry checking of sender cards.
 
-  public RhinoDocumentStore(JsonSerializerSettings jsonSerializerSettings, bool writeToFileOnChange)
-    : base(jsonSerializerSettings, writeToFileOnChange)
+  public RhinoDocumentStore(JsonSerializerSettings jsonSerializerSettings)
+    : base(jsonSerializerSettings, true)
```
JR-Morgan, line 13: What was the original intention with this bool? @AlanRynne

DUI3-DX/Sdk/Speckle.Autofac/EnumerationExtensions.cs
JR-Morgan: Ah, its not linq sorry, just `List<T>`
Out of interest, do you really prefer this over just using a regular `ForEach` loop? is this not just slower (due to closure captures)?

JR-Morgan (approved):

PR #3418, opened 2024-05-15
CNX-9308 CNX-9632 Moving to Github Actions for DUI3

adamhathcock:
CNX-9308 CNX-9632

Uses new actions for CI builds and Main builds.
Separate repo will contain logic and installers for Main artifact output with versions


AlanRynne:
Moving this out of draft. I think we're ready to get this in and polish any rough edges as we go 🚀

AlanRynne (approved):

PR #3421, opened 2024-05-15
CNX-9509 RHN opening up plugin panel again results in error

oguzhankoral:
It was painful to spot the problem and solution. Because there is a different behavior on closing and reopening Panels in Rhino.

###  First behavior -> Close via list. If it is closed/disabled via list
On re-open --> It creates `SpeckleRhinoPanelHost` object and expects UIElement under `RhinoWindows.Controls.WpfElementHost` is not identical as before or empty. That was the problem before because we register `DUI3ControlWebView` as `SingleInstance` on Autofac. To fix this I've subscribed `Panels.Closed` event to Disconnect previous UIElement from `WpfHostElement`.

### Second behavior -> Close via ❌ button on detached panel.
On re-open --> It doesn't try to create `SpeckleRhinoPanelHost`

To distinguish these two behaviors, we have static `Panels.IsPanelVisible` function. According to this flag either I disconnect UIElement or not.

Hopefully it is clear enough, otherwise you know my Discord address..

### DIFF
--- a/DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
+++ b/DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
@@ -63,7 +63,6 @@ public class AutofacRhinoModule : Module
 
     // binding dependencies
     builder.RegisterType<CancellationManager>().InstancePerDependency();
-    builder.RegisterType<ServerTransport>().As<ITransport>().InstancePerDependency();
 
     // register send filters
     builder.RegisterType<RhinoSelectionFilter>().As<ISendFilter>().InstancePerDependency();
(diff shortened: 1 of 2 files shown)


DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
```diff
@@ -63,7 +63,6 @@ protected override void Load(ContainerBuilder builder)
 
     // binding dependencies
     builder.RegisterType<CancellationManager>().InstancePerDependency();
-    builder.RegisterType<ServerTransport>().As<ITransport>().InstancePerDependency();
```
didimitrie, line 66: this looks weird as part of this pr, but probably is related to something else...

didimitrie (approved):
Tested and it works. Thanks for the deep digging! Only one unrelated remark/comment

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/AutofacRhinoModule.cs
didimitrie: cools, merge away then!

PR #3422, opened 2024-05-15
CNX 9442 Incorrect Export of 'Room Type' Parameter from Revit to Speckle

connorivy:
## Description & motivation

Previously we were sending elementId base parameters as just the elementId which is pretty much useless because we are not doing anything with these on receive. A user asked us to send the name instead which both @JR-Morgan and I think is an improvement over the existing functionality.

This change is causing us to add a lot of document queries that we weren't previously doing. Profiling the performance model that we have, this functionality causes the conversion process to go from 28ish seconds to 29ish seconds which seems acceptable (admittedly these tests were done in debug mode).

## Screenshots:

Before (left) and after (right) for wall parameters
image

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ParameterExtensions.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ParameterExtensions.cs
@@ -49,7 +49,9 @@ internal static class ParameterExtensions
       case StorageType.String:
         return parameter.AsString();
       case StorageType.ElementId:
-        return parameter.AsElementId().ToString();
+        ElementId id = parameter.AsElementId();
+        Element el = parameter.Element.Document.GetElement(id);
+        return el?.Name ?? id.ToString();
       default:
         return null;
     }


Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ParameterExtensions.cs
```diff
@@ -49,7 +49,9 @@ internal static class ParameterExtensions
       case StorageType.String:
         return parameter.AsString();
       case StorageType.ElementId:
-        return parameter.AsElementId().ToString();
+        ElementId id = parameter.AsElementId();
+        Element el = parameter.Element.Document.GetElement(id);
```
JR-Morgan, line 53: I assume that `parameter.Element` isn't actually an `Element` type (and therefore the same element we're trying to get by `GetElementId`?)

JR-Morgan (changes requested):
Shall we retarget this PR into `Main` for 2.19?

Objects/Converters/ConverterRevit/ConverterRevitShared/Extensions/ParameterExtensions.cs
connorivy: It is of "Element" type. According to the documentation, the Element prop is the element that the parameter belongs to. So, if we're converting wall parameters, then the element prop of all the parameters will be same wall object, so we do need to still query the document for the element that the parameter points to.

JR-Morgan: Gotcha, perfect

JR-Morgan (approved):

PR #3423, opened 2024-05-16
Cnx 9502 agis object tracking change detection

KatKatKateryna:
Currently the change detection only works for the change of the map elements, their properties and names, but NOT the change in the data source (actual features /data rows in the table). This requires dynamic subscription to every new layer on the map, and it is tricky to extract the layer id from the feature-changed event.

### DIFF
--- a/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/ArcGISSendBinding.cs
+++ b/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/ArcGISSendBinding.cs
@@ -11,6 +11,8 @@ using Speckle.Connectors.DUI.Models.Card;
 using Speckle.Connectors.DUI.Models.Card.SendFilter;
 using Speckle.Connectors.DUI.Settings;
 using Speckle.Connectors.Utils;
+using ArcGIS.Desktop.Mapping.Events;
+using ArcGIS.Desktop.Mapping;
 
 namespace Speckle.Connectors.ArcGIS.Bindings;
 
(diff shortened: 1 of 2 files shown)


KatKatKateryna:
extra step added on RunExpirationCheck to update ModelCard ids, otherwise it could have tried to send non-existing layers and pass silently through no-layers-to-send check.

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/ArcGISSendBinding.cs
```diff
@@ -112,17 +163,28 @@ public async Task Send(string modelCardId)
   /// <summary>
   /// Checks if any sender model cards contain any of the changed objects. If so, also updates the changed objects hashset for each model card - this last part is important for on send change detection.
   /// </summary>
-  private void RunExpirationChecks()
+  private void RunExpirationChecks(bool idsDeleted)
```
oguzhankoral, line 166: This seems different from others, can you explain in the comments why it was needed? Maybe this is something we need for others too? It is important to know when we start to refactor this function to have unified functionality as in https://spockle.atlassian.net/browse/CNX-9576

oguzhankoral (approved):
Tested and works great so far 🚀Just single comment request, then should be good to go!

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/ArcGISSendBinding.cs
KatKatKateryna: not sure if other connectors are affected in the same way, but basically this is needed to update the list of selected objectIDs saved in the model card. Otherwise, it can happen that all objects have been deleted but modelCard still have the list of Ids and prevents "No objects selected" warning on Send. Just sends an empty commit.

PR #3424, opened 2024-05-17
Error feedback on ReceiveOperation

adamhathcock:
Also initialize logging (which logs to file)

### DIFF
--- a/DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
+++ b/DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
@@ -41,12 +41,23 @@ public sealed class ReceiveOperation
 
     cancellationToken.ThrowIfCancellationRequested();
 
+    try
+    {
     // 4 - Convert objects
-    return await _syncToMainThread
-      .RunOnThread(() =>
-      {
-        return _hostObjectBuilder.Build(commitObject, projectName, modelName, onOperationProgressed, cancellationToken);
-      })
-      .ConfigureAwait(false);
+      var x = await _syncToMainThread
+        .RunOnThread(() =>
+        {
+          return _hostObjectBuilder.Build(commitObject, projectName, modelName, onOperationProgressed, cancellationToken);
+        })
+        .ConfigureAwait(false);
+
+      return x ?? new List<string>();
+
+    }
+    catch (Exception e)
+    {
+      Console.WriteLine(e);
+      throw;
+    }
   }
 }
--- a/DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
+++ b/DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
@@ -1,3 +1,4 @@
+using Microsoft.Extensions.Logging;
 using Speckle.Connectors.Utils.Builders;
 using Speckle.Core.Api;
 using Speckle.Core.Credentials;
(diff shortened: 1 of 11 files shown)


DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
```diff
@@ -41,12 +48,28 @@ public ReceiveOperation(IHostObjectBuilder hostObjectBuilder, ISyncToMainThread
...
+            modelName,
+            onOperationProgressed,
+            cancellationToken
+          );
+        })
+        .ConfigureAwait(false);
+
+      return x ?? new List<string>();
+    }
+    catch (Exception e)
```
BovineOx, line 69: At best this should use our when !IsFatal() patten which @JR-Morgan can tell you about. Catching everything here I am not sure about, it's definitely sensible but we probably want a top level handler in the bridge and my question is, would they be different. We probably already had something bogus here. @JR-Morgan needs to chip in on this because it will bleed into reporting and exception handling.

Naturally the same applies to the receive. At this least the IsFatal should go here

BovineOx (changes requested):
One minor change from my PoV, I'd also like to see tickets around these things, even if it becomes a sub-task on an existing ticket, purely because we're punting for SOC2 and change management is a big deal there, also if the branch can be made from Jira so it is tied to that ticket, even if the subtask. No drama, is all refining in the right direction.

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
adamhathcock: Happy to do that.  I was more concerned about the nullability because I was getting an exception swallowed and a null value back (that shouldn't be null)

adamhathcock: Also might tie into the logging work @JR-Morgan was looking at as this logger is separate from the one used while debugging it seems.

BovineOx: Yeah we just have a policy of not swallowing everything, if it's Fatal we have to begin to warn the customer, hey really bad sh#t happened so you might want to not continue editing your critical path Revit construction file :D

BovineOx: I did have some questions about why the return type was nullable from a lot of this, if that were possible or whether we should be saying it isn't but I'll let @JR-Morgan weigh-in on this.

JR-Morgan: Just to clarify then, 
are we saying that sometimes, `_hostObjectBuilderBuild` is returning `null`?

If so, it shouldn't be! and I don't think we should be trying to handle it here. If Its marked as `nonnullable` then its not on the caller to double check this...

DUI3-DX/Sdk/Speckle.Connectors.Utils/Operations/SendOperation.cs
```diff
@@ -20,19 +20,24 @@ ISyncToMainThread syncToMainThread
...
-  public async Task<(string rootObjId, Dictionary<string, ObjectReference> convertedReferences)> Execute(
+  public async Task<(string rootObjId, Dictionary<string, ObjectReference> convertedReferences)?> Execute(
     IReadOnlyList<T> objects,
     SendInfo sendInfo,
     Action<string, double?>? onOperationProgressed = null,
     CancellationToken ct = default
   )
   {
-    Base commitObject = await _syncToMainThread
+    Base? commitObject = await _syncToMainThread
```
JR-Morgan, line 30: I don't like these changes.
I don't think the builder should be returning null
If it is,  then the problem is with the builder, and we should fix this.

We should be using exceptions to communicate errors, A null return here gives us no context on what went wrong. We don't want silent failures.

adamhathcock: I think it's an artifact of Task usage that if the task throws an exception, the value ends up being null.

I need to double check things but I was definitely getting this when using the Rhino plugin and a conversion failed

JR-Morgan (changes requested):
I question if we want some of these changes.
Allowing `null` returns for things is a bit of an anti-pattern that I strongly feel we should avoid.
We should be trusting our nullability, throwing when we fail, and generally avoiding passing nulls around.

Lets discuss this more @adamhathcock

adamhathcock:
I agree, I was concerned about the behavior.  We should discuss what I saw and possibly how to handle it

PR #3428, opened 2024-05-18
CNX-9445 arc gis precision check for acrs circles curves ellipse

KatKatKateryna:
1. Clean up line-type geometry conversions
2. Use Map's Spatial Referrence precision to "densify" arcs and curves on Send
3. Receive Circles natively, without segmentation
4. Add support for US feet Unit 
5. Rely on FactoryCode of ArcGIS Units instead of Name (more units available through Code)


KatKatKateryna:
Do not merge yet, precision is assumed in meters but provided in native Spatial Reference units. Needs to be adjusted

KatKatKateryna:
ready for review

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Geometry/SegmentCollectionToSpeckleConverter.cs
```diff
@@ -34,27 +28,39 @@ public SOG.Polyline RawConvert(ACG.ReadOnlySegmentCollection target)
...
-        case ACG.SegmentType.Bezier:
-          var segmentBezier = (ACG.CubicBezierSegment)segment;
-          points.AddRange(_bezierConverter.RawConvert(segmentBezier).GetPoints());
-          break;
-        case ACG.SegmentType.EllipticArc:
-          var segmentElliptic = (ACG.EllipticArcSegment)segment;
-          points.AddRange(_arcConverter.RawConvert(segmentElliptic).GetPoints());
+          points = AddPtsToPolylinePts(
+            points,
+            new List<SOG.Point>()
```
AlanRynne, line 37: `AddPtsToPolylinePts` doesn't need to return the list. List are passed by reference, so the list that the method is modifying is the same as the one in the `points` variable`

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Features/PolylineFeatureToSpeckleConverter.cs
```diff
@@ -1,21 +1,46 @@
...
   {
     // https://pro.arcgis.com/en/pro-app/latest/sdk/api-reference/topic8480.html
     List<SOG.Polyline> polylineList = new();
-    foreach (var segmentCollection in target.Parts)
+    ACG.Polyline polylineToConvert = target;
+
+    // segmentize the polylines with curves using precision value of the Map's Spatial Reference
+    if (target.HasCurves is true)
+    {
+      double tolerance = _contextStack.Current.Document.SpatialReference.XYTolerance;
```
AlanRynne, line 30: Very nice that we can get the tolerance from the doc ♥️

AlanRynne (approved):
Just one minor comment about a method implementation. Nothing to change but would be good to have a chat about this as you get deeper into the C# world 🙂

PR #3429, opened 2024-05-19
CNX-9446 create enums for arc gis custom types

KatKatKateryna:
- Create clear expected types for LayerGeometryType and FieldTypes
image

- sort received geometry by type more selectively (as fallback conversions have added possibility of mixing different converted geometry types under the same parent)


AlanRynne (approved):
This is so much cleaner ♥️

Whenever we support serialising `Enum`s by their string representation we could make this even slimmer.

PR #3430, opened 2024-05-19
dynamic subscriptions to layers and tables

KatKatKateryna:
Follors the PR https://github.com/specklesystems/speckle-sharp/pull/3423 and adds dynamic subscriptions to the data source changes of layers and tables


oguzhankoral (approved):
There was an issue on registration. Probably it was missed today on @adamhathcock's PR. I've fixed it.

Tested the data tables on change detection and works great. Codewise, send binding is bubbling but we have a ticket to simplify it on every connector https://spockle.atlassian.net/browse/CNX-9576

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/ArcGISSendBinding.cs
```diff
@@ -57,6 +62,111 @@ private void SubscribeToArcGISEvents()
...
+
+  private void SubscribeToAnyDataSourceChange(Table layerTable)
+  {
+    RowCreatedEvent.Subscribe(
+      (args) =>
+      {
+        OnRowChanged(args);
+      },
+      layerTable
+    );
```
oguzhankoral, line 127: Hmm, what if we have saved layer tables on the document? Should we subscribe each table at the initialization?

KatKatKateryna: Yes, it's called on the ActiveMapView change event, which means whenever we open a map all the saved layers will be subscribed

oguzhankoral: Super!

KatKatKateryna: LayersAdded & TablesAdded events are catching up when we add layers to the document. Or did you mean something else?

KatKatKateryna: Ah yes, MapViewChanged subscribes to every layer existing in the Map which we have opened/ switched to. But definitely worth testing for edge cases

oguzhankoral: Then we are good for now, when you are back, we can investigate edge cases.

PR #3432, opened 2024-05-20
CNX-9500 Implement highlighting sent elements

oguzhankoral:


DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
```diff
@@ -49,5 +53,100 @@ public BasicConnectorBinding(DocumentModelStore store, ArcGISSettings settings,
...
+    }
+  }
+
+  private void SelectMapMembers(List<MapMember> mapMembers)
+  {
+    foreach (var member in mapMembers)
+    {
+      if (member is FeatureLayer featureLayer)
+      {
+        using (RowCursor rowCursor = featureLayer.Search())
```
JR-Morgan, line 134: Could we use the `inline` using syntax here?
i.e. 
```cs
using RowCursor rowCursor = featureLayer.Search();
```
It might help make this function easier to read

oguzhankoral: agree, will fix it!

JR-Morgan (approved):

PR #3437, opened 2024-05-23
`ConversionResult` in `HostObjectBuilder`

JR-Morgan:
Still WIP

### DIFF
--- a/Core/Core/Models/ConversionResult.cs
+++ b/Core/Core/Models/ConversionResult.cs
@@ -28,5 +28,7 @@ public sealed class ConversionResult
   }
 
   public override string ToString() =>
-    IsSuccessful ? $"Successfully converted {Target} to {Result}" : $"Failed to convert {Target}: {Error!.Message}";
+    IsSuccessful
+      ? $"Successfully converted {Target} to {Result}"
+      : $"Failed to convert {Target}: {Error!.GetType()}: {Error!.Message}";
 }
--- a/Core/Core/Models/ConversionResult.cs
+++ b/Core/Core/Models/ConversionResult.cs
@@ -1,0 +1,32 @@
+﻿using System;
+
+namespace Speckle.Core.Models;
+
+public sealed class ConversionResult
+{
+  public string? ResultId { get; }
+  public object? Result { get; }
+  public Exception? Error { get; }
+  public object Target { get; }
+
+  //[MemberNotNullWhen(true, nameof(Result))]
+  //[MemberNotNullWhen(true, nameof(ResultId))]
+  //[MemberNotNullWhen(false, nameof(Error))]
+  public bool IsSuccessful => Result is not null;
+
+  public ConversionResult(object target, object result, string resultId)
+  {
+    Target = target;
+    Result = result;
+    ResultId = resultId;
+  }
+
+  public ConversionResult(object target, Exception error)
+  {
+    Target = target;
+    Error = error;
+  }
+
+  public override string ToString() =>
+    IsSuccessful ? $"Successfully converted {Target} to {Result}" : $"Failed to convert {Target}: {Error!.Message}";
+}
--- a/DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
+++ b/DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
@@ -1,4 +1,3 @@
-using System.Diagnostics;
 using Autodesk.AutoCAD.DatabaseServices;
 using Speckle.Connectors.Autocad.HostApp;
 using Speckle.Connectors.Autocad.HostApp.Extensions;
@@ -12,22 +11,22 @@ namespace Speckle.Connectors.Autocad.Operations.Receive;
 
 public class AutocadHostObjectBuilder : IHostObjectBuilder
 {
-  private readonly IRootToHostConverter _converter;
   private readonly AutocadLayerManager _autocadLayerManager;
+  private readonly IRootToHostConverter _converter;
   private readonly GraphTraversal _traversalFunction;
 
   public AutocadHostObjectBuilder(
     IRootToHostConverter converter,
-    AutocadLayerManager autocadLayerManager,
-    GraphTraversal traversalFunction
+    GraphTraversal traversalFunction,
+    AutocadLayerManager autocadLayerManager
   )
   {
     _converter = converter;
-    _autocadLayerManager = autocadLayerManager;
     _traversalFunction = traversalFunction;
+    _autocadLayerManager = autocadLayerManager;
   }
 
-  public IEnumerable<string> Build(
+  public IReadOnlyList<ConversionResult> Build(
     Base rootObject,
     string projectName,
     string modelName,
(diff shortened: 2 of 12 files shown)


Core/Core/Models/ConversionResult.cs
```diff
@@ -0,0 +1,34 @@
...
+
+  //[MemberNotNullWhen(true, nameof(Result))]
+  //[MemberNotNullWhen(true, nameof(ResultId))]
+  //[MemberNotNullWhen(false, nameof(Error))]
+  public bool IsSuccessful => Result is not null;
+
+  public ConversionResult(object target, object result, string resultId)
+  {
+    Target = target;
+    Result = result;
```
oguzhankoral, line 20: This object might know whether the conversion is successful with fallback displayValue or not. I mean, it is "successful" with "but". TBD

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
```diff
@@ -40,64 +39,65 @@ CancellationToken cancellationToken
...
+    foreach (var tc in _traversalFunction.TraverseWithProgress(rootObject, onOperationProgressed, cancellationToken))
     {
-      foreach (TraversalContext tc in traversalGraph)
+      try
+      {
+        var convertedObjects = ConvertObject(tc, baseLayerPrefix, uniqueLayerNames);
+
+        results.AddRange(convertedObjects.Select(e => new ConversionResult(tc, e, e.Handle.Value.ToString())));
+      }
+      catch (Exception ex) when (!ex.IsFatal())
```
oguzhankoral, line 56: hmm, just have general Qs: 
- Will we continue conversion if some bad error happens that might affect the state of the app?
- Will we have a revert operation mechanism if smth bad happens?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Bindings/RhinoReceiveBinding.cs
```diff
@@ -51,7 +52,7 @@ public async Task Receive(string modelCardId)
       }
 
       // Receive host objects
-      IEnumerable<string> receivedObjectIds = await unitOfWork.Service
+      IReadOnlyList<ConversionResult> conversionResults = await unitOfWork.Service
```
oguzhankoral, line 55: Probably, we will have a different structure on send and receive since we also return `ObjectReference`s on send, so maybe it would be nice to name class explicitly as `ReceiveConversionResult`

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Bindings/RhinoReceiveBinding.cs
```diff
@@ -63,8 +64,7 @@ public async Task Receive(string modelCardId)
         )
         .ConfigureAwait(false);
 
-      // POC: Here we can't set receive result if ReceiveOperation throws an error.
-      Commands.SetModelReceiveResult(modelCardId, receivedObjectIds.ToList());
+      Commands.SetModelReceiveResult(modelCardId, conversionResults);
```
oguzhankoral, line 67: This is where I can start to think interop

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/RhinoHostObjectBuilder.cs
```diff
@@ -38,25 +38,18 @@ CancellationToken cancellationToken
     var baseLayerName = $"Project {projectName}: Model {modelName}";
 
     var objectsToConvert = _traverseFunction
-      .Traverse(rootObject)
-      .Where(obj => obj.Current is not Collection)
-      .Select(ctx => (GetLayerPath(ctx), ctx.Current))
-      .ToArray();
+      .TraverseWithProgress(rootObject, onOperationProgressed, cancellationToken)
```
oguzhankoral, line 41: Do we report progress on Traversing to UI?

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bindings/ReceiveBindingUICommands.cs
```diff
@@ -11,13 +12,16 @@ public class ReceiveBindingUICommands : BasicConnectorBindingCommands
...
 
-  public void SetModelReceiveResult(string modelCardId, List<string> bakedObjectIds)
+  public void SetModelReceiveResult(string modelCardId, IReadOnlyList<ConversionResult> conversionResults)
   {
     ReceiverModelCardResult res =
       new()
       {
         ModelCardId = modelCardId,
-        ReceiveResult = new ReceiveResult() { BakedObjectIds = bakedObjectIds }
+        ReceiveResult = new ReceiveResult()
```
oguzhankoral, line 21: Need some changes on UI side to send proper result

DUI3-DX/Sdk/Speckle.Connectors.Utils/Builders/TraversalExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+  {
+    var traversalGraph = traversalFunction.Traverse(rootObject).ToArray();
+    int count = 0;
+    foreach (var tc in traversalGraph)
+    {
+      cancellationToken.ThrowIfCancellationRequested();
+
+      yield return tc;
+
+      onOperationProgressed?.Invoke("Converting", (double)++count / traversalGraph.Length);
```
oguzhankoral, line 23: Should we really report to UI progress when traversing? If so I believe it shouldn't be named with "Converting"?

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ConverterWithFallback.cs
```diff
@@ -54,24 +52,13 @@ public object Convert(Base target)
       return FallbackToDisplayValue(displayValue);
```
oguzhankoral, line 52: Need some thoughts on it -> How can we report fallback conversion, or should we?

Core/Core/Models/ConversionResult.cs
JR-Morgan: Potentially simply the result type will be enough for us to communicate this if we wish.
"Brep" was converted as "Mesh" communicates exactly the same thing as "with dispalyValues", but imo will be more intuitive to the user.

If we wanted more than this, I could see some utility in knowing what converter was used to convert the object. 
But this would require some more restructuring of our "...WithFallBack" service, because right now, the converter is abstracted and the connector doesn't see it.

I think we could do something nice, where the "...WithFallBack" simply returns the converter (without calling it), instead of right now, where it both finds and immediately calls it

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
JR-Morgan: Right now, I've gone for "continue conversion of the next object" for all non-fatal exceptions.
This was the behaviour of DUI2 and was (I seem to remember)  agreed with Dim.

I don't think we ever saw any problems with this approach.

Although, when we do logging, we will be logging "expected" exceptions (like `SpeckleException` subtypes) differently from unexpected (e.g. nullreference) 

We do have the flexibility here to pick and choose if we wanted to pivot to something else.
But I think this will strike a good balance for now.

We'll take a look at transactions/reverting for cancellation, probably in a separate PR.

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Receive/RhinoHostObjectBuilder.cs
JR-Morgan: The actually traversal step, does not get reported with progress because its super quick and only happens once.

 `TraverseWithProgress`, since it returns a lazily enumerated `IEnumerable` reports the actual conversion progress (as the enumeration of objects occurs)

This was simply my hope to be DRY with the progress reporting, but I'm not tied to it if we find a better way.

DUI3-DX/Sdk/Speckle.Connectors.Utils/Builders/TraversalExtensions.cs
JR-Morgan: See other comment, this isn't traversal progress (because traversal is super quick)
This is the conversion progress (as the `IEnumerable` is being enumerated)

Let me know if you want me to explain this better over voice call.

DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ConverterWithFallback.cs
JR-Morgan: See other comment above, Perhaps we don't need to, since we will know the original and result type and that's the main thing users will care about.

But if we wanted to go further, and know the converter that was used, I have an idea, but I think we should wait till we can get some feedback from the team.

JR-Morgan:
Changes will be included in https://github.com/specklesystems/speckle-sharp/pull/3448

PR #3441, opened 2024-05-27
Chore (dui3): error message from inner exception on sync functions

oguzhankoral:
Otherwise error message is rendered wrong on UI.
@JR-Morgan happy to consider if you have alternative.

Before
image

After
image

### DIFF
--- a/DUI3-DX/DUI3/Speckle.Connectors.DUI/Bridge/BrowserBridge.cs
+++ b/DUI3-DX/DUI3/Speckle.Connectors.DUI/Bridge/BrowserBridge.cs
@@ -255,7 +255,12 @@ public class BrowserBridge : IBridge
   /// </summary>
   private void ReportUnhandledError(string requestId, Exception e)
   {
-    var errorDetails = new { e.Message, Error = e.ToString() };
+    var message = e.Message;
+    if (e is TargetInvocationException tie) // Exception on SYNC function calls. Message should be passed from inner exception since it is wrapped.
+    {
+      message = tie.InnerException?.Message;
+    }
+    var errorDetails = new { Message = message, Error = e.ToString() };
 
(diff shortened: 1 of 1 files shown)


DUI3-DX/DUI3/Speckle.Connectors.DUI/Bridge/BrowserBridge.cs
```diff
@@ -255,7 +255,12 @@ private void ExecuteMethod(string methodName, string requestId, string args)
   /// </summary>
   private void ReportUnhandledError(string requestId, Exception e)
   {
-    var errorDetails = new { e.Message, Error = e.ToString() };
+    var message = e.Message;
+    if (e is TargetInvocationException tie) // Exception on SYNC function calls. Message should be passed from inner exception since it is wrapped.
+    {
+      message = tie.InnerException?.Message;
+    }
+    var errorDetails = new { Message = message, Error = e.ToString() };
```
AlanRynne, line 263: If I may add a small suggestion. Calling `e.ToString()` includes the callstack of the exception, which is useful for us as devs but it becomes noise to the end user.

We have an extension method `ToFormattedString` that prints all exception messages deeply nested into a given exception, and I think that could result in clearer information to the user when they press "see details"

AlanRynne (approved):
Approved! Just have a small suggestion on the details text.

I'll leave it to your judgement to include it (or not)

DUI3-DX/DUI3/Speckle.Connectors.DUI/Bridge/BrowserBridge.cs
oguzhankoral: Good point! will think on it also what/how show messages on dialog

JR-Morgan (approved):

PR #3443, opened 2024-05-28
CNX-9501 agis highlighting loaded elements

KatKatKateryna:
- Save LayerURI instead of speckleId to Receive ModelCard, so selection/highlight method can find the layers in the host app
- Add Table of Content (TOC) selection as well, because non-feature layers and Mulltipatch feature layer selections will not be visible on the canvas. Also, there might be multiple layers with the same shapes but different properties, just canvas selection might not be helpful


AlanRynne (approved):
Can't say i know enough of ArcGIS api to fully understand it. But it looks good codewise

KatKatKateryna:
> @KatKatKateryna slowness on the selection fixed now?

it is

oguzhankoral (approved):

PR #3454, opened 2024-05-30
Promoted several analysers to warnings for DUI3

JR-Morgan:


AlanRynne (approved):
Same as #3453, lets merge these 2 and try to upate `dui3/alpha` so that any conflicts are fixed ASAP

PR #3459, opened 2024-05-31
Arcgis documentation kate jedd

KatKatKateryna:
Expanding context stack Document from Map to (Project, Map, SpeckleDatabasePath) and making it accessible throughout converters. 
Reasoning: ensuring and getting the SpeckleDatabasePath should only be called once at the start of Operation, and be accessible for Connector and Converters


JR-Morgan (approved):

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/ArcGISConversionContextStack.cs
```diff
@@ -1,18 +1,99 @@
...
+    ) { }
+
+  private const string FGDB_NAME = "Speckle.gdb";
+
+  public static Uri EnsureOrAddSpeckleDatabase()
+  {
+    return AddDatabaseToProject(GetDatabasePath());
+  }
+
+  public static Uri GetDatabasePath()
```
oguzhankoral, line 35: 1. Why we have it static? Can't we retrieve from instance?
2. Why we catch specific exception types and rethrow regardless?

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/ArcGISConversionContextStack.cs
```diff
@@ -1,18 +1,99 @@
...
+  }
+
+  public static Uri AddDatabaseToProject(Uri databasePath)
+  {
+    // Create a FileGeodatabaseConnectionPath with the name of the file geodatabase you wish to create
+    FileGeodatabaseConnectionPath fileGeodatabaseConnectionPath = new(databasePath);
+    // Create actual database in the specified Path unless already exists
+    try
+    {
+      Geodatabase geodatabase = SchemaBuilder.CreateGeodatabase(fileGeodatabaseConnectionPath);
```
oguzhankoral, line 67: When we need to create it and why dispose it immediately if we won't use in this scope?

oguzhankoral (changes requested):

DUI3-DX/Converters/ArcGIS/Speckle.Converters.ArcGIS3/ArcGISConversionContextStack.cs
KatKatKateryna: It creates a database on a hard drive, here we just need to make sure it exists and ArcGIS project is linked to its path

oguzhankoral (approved):

PR #3460, opened 2024-05-31
CNX-9443 Rhino Compute units set as mm

AlanRynne:
> [!IMPORTANT]
> This PR targets `main` and should be released as part of the next hotfix `2.19.1`

https://spockle.atlassian.net/browse/CNX-9443

- Adds RHINO7_OR_GREATER to grasshopper
- Fixes GetVersionedAppName to use Loader equivalent. Did not delete it but flagged it as obsolete
- Fixed exception on initial open event in headless document run.


oguzhankoral (approved):
Makes sense! LGTM

PR #3461, opened 2024-05-31
CNX-9586 Text input not working in Rhino 8

AlanRynne:
https://spockle.atlassian.net/browse/CNX-9586
Fixes text input not working by applying the same fix as in #1362


AlanRynne:
CI says it's failing, but only bc I accidentaly re-ran it adn then cancelled it.

All jobs are ✅ so we should be good to go.

adamhathcock (approved):

oguzhankoral (approved):
image
Seems OK

PR #3469, opened 2024-06-04
Copy important changes to core from DUI3/alpha to dev

JR-Morgan:


AlanRynne:
If we wanted this for users of the hackathon, this should actually go into `main` to release along side 2.19.1 stable

JR-Morgan:
I'll rebase for main

PR #3476, opened 2024-06-05
Revit file sharing

adamhathcock:
Overrides https://github.com/specklesystems/speckle-sharp/pull/3468

Uses info from https://github.com/dotnet/sdk/issues/2511

Basically, shared projects are deprecated and only made because of UWP.  Using the new build system features makes us more future proof


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Speckle.Connectors.RevitShared.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.Build.NoTargets/3.3.0">
```
BovineOx, line 1: This is what makes it shared, I guess?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Raw/Geometry/SolidConversionToSpeckle.cs
```diff
@@ -31,23 +31,23 @@ public class SolidConversionToSpeckle : IRawConversion<DB.Solid, List<SOG.Mesh>>
...
+    return _meshByMaterialConverter.Convert(meshesByMaterial);
   }
 
   private Dictionary<DB.ElementId, List<DB.Mesh>> GetTriangulatedMeshesFromSolidByMaterial(DB.Solid solid)
   {
     var result = new Dictionary<DB.ElementId, List<DB.Mesh>>();
     foreach (DB.Face face in solid.Faces)
     {
-      if (!result.ContainsKey(face.MaterialElementId))
+      if (!result.TryGetValue(face.MaterialElementId, out var materials))
```
BovineOx, line 45: Yeah.... much better

BovineOx (approved):
I think it's fine but @AlanRynne should review also before merging and also we must eb sure it's all good on the CI

DUI3-DX/Connectors/Revit/Speckle.Connectors.Revit2023/Speckle.Connectors.Revit2023.csproj
```diff
@@ -34,6 +32,16 @@
     <PackageReference Include="Speckle.Revit.API" Version="2023.0.0" IncludeAssets="compile;build" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\Speckle.Connectors.RevitShared\**\*.cs"  Exclude="..\Speckle.Connectors.RevitShared\obj\**\*.*"/>
+  </ItemGroup>
```
AlanRynne, line 37: I've never had to run this Exclude attribute for the `obj` folder. 🤔

It kinda feels wrong

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Speckle.Connectors.RevitShared.csproj
```diff
@@ -0,0 +1,17 @@
...
+  
+
+  <ItemGroup>
+    <Compile Include="**\*.cs" Exclude="obj\**\*.*"/>
+  </ItemGroup>
+  <ItemGroup>
+    <Page Include="**\*.xaml"  Exclude="obj\**\*.*">
+      <SubType>Designer</SubType>
+      <Generator>MSBuild:Compile</Generator>
+    </Page>
```
AlanRynne, line 15: If the .xaml and .cs files live in this project, why do we need to also add them to the specific 2023 project as well?

Also, that exclude for the `obj` folder feels wrong 😅

AlanRynne (commented):
Got some questions about this before we merge. My understanding is that `NoTarget` SDK projects are intended for projects that have no code but do have project dependencies (i.e. building it results in no DLL being built, just it's dependencies being copied)
The way we're doing this is somewhat opposite to that 🤔

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Speckle.Connectors.RevitShared.csproj
adamhathcock: This csproj isn't used at all so we don't want msbuild to do anything but can be seen in the sln

DUI3-DX/Connectors/Revit/Speckle.Connectors.Revit2023/Speckle.Connectors.Revit2023.csproj
adamhathcock: making a csproj just for shared makes an obj folder.

we don't strictly need the csproj for shared (since each individual "real" csproj includes the files separately) but it's making the shared stuff viewable in the sln

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Speckle.Connectors.RevitShared.csproj
adamhathcock: they don't really "live" here....it's not used by other csprojs if you look

adamhathcock:
> Got some questions about this before we merge. My understanding is that `NoTarget` SDK projects are intended for projects that have no code but do have project dependencies (i.e. building it results in no DLL being built, just it's dependencies being copied) The way we're doing this is somewhat opposite to that 🤔

You're right....the shared csprojs could be deleted and we wouldn't need to exclude the OBJ folder as it wouldn't exist.  The shared csproj is just for viewing in the SLN

files are directly added to each csproj via the compile include

adamhathcock:
went through this again with @AlanRynne in some detail and actually gives a slightly worse DX even though shprojs aren't supported going forward.  Taking this on the chin for now.

PR #3477, opened 2024-06-05
CNX-9760 [Rhino]: add gis attributes as user strings

clairekuang:
## Description & motivation

Adds a direct conversion for `PolygonElement`, the new `GisFeature` object, as well as a general fallback conversion storing `attributes` on the top level object.

Also converts gis classes as groups of objects and fixes layer placement

## Changes:

- rhino connector and converter

image


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/UI/ConnectorBindingsRhino.Receive.cs
```diff
@@ -650,6 +640,42 @@ void FlattenConvertedObject(object item)
           bakedCount++;
           break;
 
+        case Group o: // this is a GIS object
```
AlanRynne, line 643: Are we sure this is the only case this would happen?

Would this explode if the converter returns a group that has no attributes for example?

AlanRynne (commented):
Just a couple questions around edge-cases or future work.

Codewise it looks right but I have not tested this yet locally.

Doing this sometime this morning

adamhathcock (approved):
Smells fine

PR #3479, opened 2024-06-06
Agis feature highlights with report

KatKatKateryna:
current limitations:
- some report items are not reaching the BakeToMap step and are lost
- need to implement per-feature selection, but everything is ready for it


DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
```diff
@@ -88,35 +88,63 @@ private async void HighlightObjectsOnView(List<string> objectIds)
...
+  private Dictionary<string, (MapMember member, List<int>? rows)> GetMapMembers(List<string> objectIds, MapView mapView)
   {
-    List<MapMember> mapMembers = new();
+    Dictionary<string, (MapMember member, List<int>? rows)> mapMembersRows = new();
 
     foreach (string objectId in objectIds)
     {
-      MapMember mapMember = mapView.Map.FindLayer(objectId);
+      List<string> uriRows = objectId.Split("_speckleRowIndex_").ToList();
+      string mapURI = uriRows[0];
```
oguzhankoral, line 108: Once we split `objectId` with magical keyword, is the first one mapURI? Is this mean we combined them before with `_speckleRowIndex_`? We should make a class to handle this if we do some magic here.

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
```diff
@@ -88,35 +88,63 @@ private async void HighlightObjectsOnView(List<string> objectIds)
...
+      int? index = null;
+      if (uriRows.Count > 1)
```
oguzhankoral, line 110: These checks and assignments from list indexes are a bit unclear. We should have a class to handle this. This function has too much responsibility.

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Operations/Receive/HostObjectBuilder.cs
```diff
@@ -125,63 +231,75 @@ CancellationToken cancellationToken
 
     int allCount = objectsToConvert.Count;
     int count = 0;
-    Dictionary<string, (string path, Geometry converted, string? parentId)> convertedGeometries = new();
+    Dictionary<string, (string path, Geometry converted, string? parentId, Base baseObj)> convertedGeometries = new();
```
oguzhankoral, line 234: what is the key string here and the value is too much. As far as understand `ConvertNonNativeGeometries` return 4 different value as tuple. We should start to get rid of this tuples.

oguzhankoral (changes requested):
I wouldn't have it immediately to alpha. Overall a bit hard to follow because of many tuples, we need to convert them into structs to have code more readable. Once we merge the `jrm/rootObjectBuilder` to the `dui3/alpha` -> lets rebase this into alpha and continue

PR #3480, opened 2024-06-06
CNX-9295 Speckle Connector doesn't initialize on Mac

AlanRynne:
https://spockle.atlassian.net/browse/CNX-9295
Fixes error initialising Rhino 7 on mac due to loggerConfiguration not being passed down to the Initialize call.

Additionally, adds `RhinoPluginType` to the csproj as `rhp`, which turns out is absolutely necessary for Debug on Mac.

### DIFF
--- a/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
+++ b/ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
@@ -184,7 +184,7 @@ public class SpeckleRhinoConnectorPlugin : PlugIn
 
       var hostAppName = HostApplications.Rhino.Slug;
       var hostAppVersion = Utils.GetRhinoHostAppVersion();
-      SpeckleLog.Initialize(HostApplications.Rhino.Slug, Utils.GetRhinoHostAppVersion());
+      SpeckleLog.Initialize(HostApplications.Rhino.Slug, Utils.GetRhinoHostAppVersion(), logConfig);
       SpeckleLog.Logger.Information(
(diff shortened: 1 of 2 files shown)


ConnectorRhino/ConnectorRhino/ConnectorRhinoShared/Entry/Plugin.cs
```diff
@@ -184,7 +184,7 @@ protected override LoadReturnCode OnLoad(ref string errorMessage)
 
       var hostAppName = HostApplications.Rhino.Slug;
       var hostAppVersion = Utils.GetRhinoHostAppVersion();
-      SpeckleLog.Initialize(HostApplications.Rhino.Slug, Utils.GetRhinoHostAppVersion());
+      SpeckleLog.Initialize(HostApplications.Rhino.Slug, Utils.GetRhinoHostAppVersion(), logConfig);
```
BovineOx, line 187: wait, does this completely fix it?

AlanRynne: Answered on Discord but copying it over here:

we obviously had this serilog.exceptions issue in the past, so we added a property enhancedLogContext to the loggerConfiguration that we could flip to have that non initialised.

This "fix" was reverted accidentaly when we fixed the issue with logs being mixed between Rhino/Grasshopper, and the logConfig was no longer being passed down

This caused the config to be the default, which sets enhancedLogContext as true, and runs that chunk of code that calls EnrichWithExceptions()

JR-Morgan (approved):

BovineOx (approved):

PR #3484, opened 2024-06-06
Use Map on GetDocumentInfo instead Project

oguzhankoral:


KatKatKateryna (approved):
It solves the problem! 
For the future, when we figure out exact use of the DocumentInfo properties, we might consider changing to (Project.Current.URI, Project.Current.Name, MapView.Active.Map.URI) or something more specific (URI objects for maps and layers are basically unique IDs, and for Project it's the location on the disk) 
image

PR #3487, opened 2024-06-06
fetching branchName for FE2 URLs

KatKatKateryna:
This solves the Grasshopper Outdated Receive Component notification issue in FE2, because the branch.name of the commit created on the server should correspond to the StreamWrapper subscription. And so far it was "hacked" to store the branch id as a BranchName for FE2 URLs.

While this is a "proper" way to set the branchName, it might affect other methods using StreamWrapper constructor. Send/Receive was tested and works fine. But if this is too risky, we can switch to another hack checking for the branch.id here instead:
image


AlanRynne (commented):
This looks great @KatKatKateryna, I'm going to give it a good test on Grasshopper and merge it in if all is good :)

This still has the issue of branch 101+ won't work (@jsdbroughton I know you love this limitation), but I think we can live with this?

jsdbroughton:
> This still has the issue of branch 101+ won't work (@jsdbroughton I know you love this limitation), but I think we can live with this?

fixing the branch limits would be out of scope for this anyways..

AlanRynne (approved):

PR #3488, opened 2024-06-06
Do not catch speckle exception at top

oguzhankoral:
Also do not throw speckle exception in UnitOfWorkFactory. Registration issues in resolving service will be red flag on UI.


BovineOx (approved):
LGTM

PR #3491, opened 2024-06-07
Dim/dui3/send caching di (DUI3-79)

didimitrie:
DI-ifies the send conversion cache, and removes a bunch of wiggle code that was no longer needed as a result. 

Points of focus for the review is arcgis: it builds, but i haven't tested it locally to see that it does what it's supposed to. Do note the last paragraph on the cache class: 

>  Caching is optional in the send ops; an instance of this should be injected only in applications <b>where we know we can rely on change tracking!

If we do not trust change detection in arcgis (ie, difficult re object selection, layers, etc.), it can be skipped entirely! 

https://spockle.atlassian.net/browse/DUI3-29


DUI3-DX/Sdk/Speckle.Connectors.Utils/Caching/SendConversionCache.cs
```diff
@@ -0,0 +1,48 @@
...
+  /// </summary>
+  /// <param name="objectIds"></param>
+  public void EvictObjects(IEnumerable<string> objectIds);
+  bool TryGetValue(string projectId, string applicationId, out ObjectReference objectReference);
+}
+
+///<inheritdoc/>
+public class SendConversionCache : ISendConversionCache
+{
+  public SendConversionCache() { }
```
oguzhankoral, line 28: not needed

DUI3-DX/DUI3/Speckle.Connectors.DUI/Models/Card/SenderModelCard.cs
```diff
@@ -1,12 +1,11 @@
...
 
 namespace Speckle.Connectors.DUI.Models.Card;
 
 public class SenderModelCard : ModelCard
 {
   public ISendFilter? SendFilter { get; set; }
 
-  [JsonIgnore]
-  public HashSet<string> ChangedObjectIds { get; set; } = new();
+  // [JsonIgnore]
```
oguzhankoral, line 9: happy to get rid of from it. You can remove it I believe. God bless `EvictObjects`

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RhinoRootObjectBuilder.cs
```diff
@@ -17,10 +19,12 @@ namespace Speckle.Connectors.Rhino7.Operations.Send;
 public class RhinoRootObjectBuilder : IRootObjectBuilder<RhinoObject>
 {
   private readonly IUnitOfWorkFactory _unitOfWorkFactory;
+  private readonly ISendConversionCache? _sendConversionCache;
 
-  public RhinoRootObjectBuilder(IUnitOfWorkFactory unitOfWorkFactory)
+  public RhinoRootObjectBuilder(IUnitOfWorkFactory unitOfWorkFactory, ISendConversionCache? sendConversionCache = null)
```
oguzhankoral, line 24: If the concrete type of this interface not registered at all in Connector, this might throw but not sure. @BovineOx can put more input on this. Do we need nullability here since this is connector specific builder?

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -57,25 +55,16 @@ public List<ISendFilter> GetSendFilters()
...
 
   public SendBindingUICommands Commands { get; }
 
-  private async Task HandleSend(string modelCardId)
+  public async Task Send(string modelCardId)
   {
-    // POC: this try catch pattern is coming from Rhino. I see there is a semi implemented "SpeckleTopLevelExceptionHandler"
-    // above that wraps the call of this HandleSend, but it currently is not doing anything - resulting in all errors from here
-    // bubbling up to the bridge.
+    // Note: removed top level handling thing as it was confusing me
```
oguzhankoral, line 67: yes, thanks for it! we do it in bridge already

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/BasicConnectorBindingRevit.cs
```diff
@@ -88,7 +88,7 @@ public void HighlightModel(string modelCardId)
     SenderModelCard model = (SenderModelCard)_store.GetModelById(modelCardId);
 
     var elementIds = model.SendFilter.NotNull().GetObjectIds().Select(ElementId.Parse).ToList();
-    if (elementIds.Count != 0)
+    if (elementIds.Count == 0)
```
oguzhankoral, line 91: back to square one, I don't even know how/why it transformed that much

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/Extensions/DocumentExtensions.cs
```diff
@@ -17,13 +18,26 @@ public static List<AutocadRootObject> GetObjects(this Document doc, IEnumerable<
...
-              objects.Add(new(dbObject, objectIdHandle));
+              if (tr.GetObject(myObjectId, OpenMode.ForRead) is DBObject dbObject)
+              {
+                objects.Add(new(dbObject, objectIdHandle));
+              }
             }
           }
+          catch (Autodesk.AutoCAD.Runtime.Exception e)
+          {
+            if (e.ErrorStatus == ErrorStatus.WasErased) // Note: TBD if we want to catch more things in here. For now maybe not, but it does seem like this function gets into "crashes the host app territory"
```
oguzhankoral, line 34: Since it is in foreach loop we can maybe aggregate them and throw as global error in UI. yes, TBD

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadSendBinding.cs
```diff
@@ -162,14 +158,14 @@ private async Task SendInternal(string modelCardId)
...
 
-      // Store the converted references in memory for future send operations, overwriting the existing values for the given application id.
-      foreach (var kvp in sendResult.ConvertedReferences)
-      {
-        _convertedObjectReferences[kvp.Key + modelCard.ProjectId] = kvp.Value;
-      }
-
-      // It's important to reset the model card's list of changed obj ids so as to ensure we accurately keep track of changes between send operations.
-      modelCard.ChangedObjectIds = new();
+      // // Store the converted references in memory for future send operations, overwriting the existing values for the given application id.
```
oguzhankoral, line 161: I would delete as others

oguzhankoral (changes requested):
Overall I'm happy it is sorted out nicely 🙌
My main question on having caching nullable, how it can be null?
As far as I understand caching is doable in ArcGIS and it is quite important since we might have some layouts that user don't touch often. Good to cache them on send.

didimitrie:
> My main question on having caching nullable, how it can be null?

Basically I am not sure if all connectors will have this cache, so hence it's nullable (optinal) everywhere. Send should work without it - for example, if i comment out in rhino this line, 

``` 
    // register send conversion cache
    builder.AddSingleton<ISendConversionCache, SendConversionCache>();
```

in the DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/RhinoConnectorModule.cs, it will send as always, but just not use the cache. 

Does this make sense @oguzhankoral?

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RhinoRootObjectBuilder.cs
didimitrie: Good point actually. If we know Rhino's going to have it, it doesn't make sense here for it to be nullable 🤔

KatKatKateryna:
As a general comment, the app is running and changedObjectIds are detected as before. 
Re change detection: ATM, we track it on layer level, and it works well. More granularity (and more complexity) can be added for report purpose, but for Selecting/Updating selection looks fine to me

BovineOx:
> > My main question on having caching nullable, how it can be null?
> 
> Basically I am not sure if all connectors will have this cache, so hence it's nullable (optinal) everywhere. Send should work without it - for example, if i comment out in rhino this line,
> 
> ```
>     // register send conversion cache
>     builder.AddSingleton<ISendConversionCache, SendConversionCache>();
> ```
> 
> in the DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/DependencyInjection/RhinoConnectorModule.cs, it will send as always, but just not use the cache.
> 
> Does this make sense @oguzhankoral?
> 
> I think I get it now (after i replied here [#3491 (comment)](https://github.com/specklesystems/speckle-sharp/pull/3491#discussion_r1631180249)) : pretty much the only place it actually needs to be nullable is in the shared `RootObjectSender` class, which is shared by any connector, including ones that will potentially not have this cache.
> 
> That okay ?

As I mentioned above we should probably have a NoCacheSendConversionCache() I am unsure if the Singleton() is the right scope also, maybe, probably it should be as the cache will persist across sends.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -57,25 +55,16 @@ public List<ISendFilter> GetSendFilters()
     return new List<ISendFilter> { new RevitSelectionFilter() { IsDefault = true } };
   }
 
-  public async Task Send(string modelCardId)
-  {
-    await SpeckleTopLevelExceptionHandler
```
JR-Morgan, line 62: It looks like `SpeckleTopLevelExceptionHandler` is no longer used. Shall we delete the class also

JR-Morgan (changes requested):

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
didimitrie: done!

DUI3-DX/Connectors/Rhino/Speckle.Connectors.Rhino7/Operations/Send/RhinoRootObjectBuilder.cs
didimitrie: done, we now have a null send conversion cache for future use, and it's not optional anywhere now

didimitrie:
Oki doki, ready for review again: 
- made the cache non-optional anywhere, and added a do nothing one for future use in cnx's that can't support it (cc @BovineOx @oguzhankoral )
- removed stale class and comments (cc @JR-Morgan @oguzhankoral) 

Also thx @KatKatKateryna for testing arcgis!

oguzhankoral (approved):
LGTM 🚀

PR #3495, opened 2024-06-08
CNX-9775 sending revit elements via dynamo

oguzhankoral:
We have assumption on this user has installed Revit connector, otherwise we will be missing `RevitSharedResources`


Objects/Converters/ConverterRevit/ConverterRevitShared/ConversionUtils.cs
```diff
@@ -390,39 +390,48 @@ public static T GetParamValue<T>(DB.Element elem, BuiltInParameter bip, string u
 #else
     ForgeTypeId unitTypeId = null;
 #endif
+    
+    ParameterToSpeckleData paramData;
+    
+    // Local function to create ParameterToSpeckleData
+    ParameterToSpeckleData CreateParamData()
+    {
```
AlanRynne, line 398: Why local function? 🙂

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
```diff
@@ -126,32 +145,45 @@ public IRevitCategoryInfo GetRevitCategoryInfo(string categoryName)
...
     {
-      return categoryInfo;
-    }
+      revitCategoryInfoCache = revitDocumentAggregateCache.GetOrInitializeWithDefaultFactory<IRevitCategoryInfo>();
 
-    foreach (var info in revitCategoryInfoCache.GetAllObjects())
-    {
-      if (categoryName.IndexOf(info.CategoryName, StringComparison.OrdinalIgnoreCase) >= 0)
+      categoryInfo = revitCategoryInfoCache.TryGet(categoryName);
+      if (categoryInfo != null)
```
AlanRynne, line 157: This makes me a bit supicious... we were previously calling `GetAllObjects` instead of this and I'm not sure the reason why we did this.

Your change seems legit though... just wondering if we're missing something here.

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
```diff
@@ -126,32 +145,45 @@ public IRevitCategoryInfo GetRevitCategoryInfo(string categoryName)
...
-        }
+        return SHC.Undefined;
       }
     }
-
-    return SHC.Undefined;
+    else
+    {
+      return SHC.Undefined;
+    }
```
AlanRynne, line 169: could be simpler if we flip the condition? could somewhat reduce nesting and make the logic a bit more linear.

The 2 return statements above this one could also be collapsed into a `return categoryInfo ?? SHC.Undefined` which I consider neater (super personal opinion here...)

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
```diff
@@ -126,32 +145,45 @@ public IRevitCategoryInfo GetRevitCategoryInfo(string categoryName)
...
+    //   foreach (var alias in info.CategoryAliases)
+    //   {
+    //     if (categoryName.IndexOf(alias, StringComparison.OrdinalIgnoreCase) >= 0)
+    //     {
+    //       return info;
+    //     }
+    //   }
+    // }
+    //
+    // return SHC.Undefined;
```
AlanRynne, line 186: If this is no more, let's just kill it. We can use git to see the prior state

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitElementTypeUtils.cs
```diff
@@ -140,6 +140,12 @@ public void SetElementFamily(Base @base, string family)
       appObj.Update(logItem: $"Could not find valid incoming type for element of type \"{element.speckle_type}\"");
     }
     var typeInfo = AllCategories.GetRevitCategoryInfo<T>(element);
+
+    if (revitDocumentAggregateCache is null)
+    {
+      return default;
+    }
```
AlanRynne, line 147: Is this safe? This means that it will always return null if the cache is not set, which is not immediately obvious.

AlanRynne (commented):
Just a couple comments 🙌🏼

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
AlanRynne: If this turns into not a 3 second refactor, disregard it.

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
oguzhankoral: This is also my most suspicious place, regardless I believe we should test this with real version too, on both Revit and Dynamo to understand we are having side effect on anything. 🤞

Objects/Converters/ConverterRevit/ConverterRevitShared/AllRevitCategories.cs
oguzhankoral: I guess you were referencing to this body (from 156-163) right? Confused because your comment targeting to 166-169

```c#
if (categoryInfo != null)
      {
        return categoryInfo;
      }
      else
      {
        return SHC.Undefined;
      }
```

Objects/Converters/ConverterRevit/ConverterRevitShared/RevitElementTypeUtils.cs
oguzhankoral: The logic was around this as below so I thought it is safe to return immediately if even cache is null....

```c#
var types = revitDocumentAggregateCache
      .GetOrInitializeWithDefaultFactory<List<ElementType>>()
      .GetOrAddGroupOfTypes(typeInfo);

    if (!types.Any())
    {
      var name = typeof(T).Name;
      if (element["category"] is string category && !string.IsNullOrWhiteSpace(category))
      {
        name = category;
      }

      appObj.Update(
        status: ApplicationObject.State.Failed,
        logItem: $"Could not find any loaded family to use for category {name}."
      );

      return default;
    }
```

AlanRynne (approved):

PR #3496, opened 2024-06-08
Logger will not try and resolve path if not logging to disk

JR-Morgan:
https://speckle.community/t/how-to-override-path-or-disable-specklelog/11705/3


JR-Morgan:
@AlanRynne since this is no longer a blocker for [the user](https://speckle.community/t/how-to-override-path-or-disable-specklelog/11705/3). I don't think we should rush this one in, I'd like to test some more.

AlanRynne (approved):
This looks good but I agree about testing it heavily before going forward :) Let's chat later

PR #3498, opened 2024-06-09
Fix revit threading

oguzhankoral:
Revit was syncing before with its current thread. This behavior changed when syncing with the Main thread became the default. I overridden it to the current thread.


JR-Morgan:
Should we be implementing our own Revit sync to thread, implemented using `APIContext.Run`. I foresee we'll need to do this for receive if not also for sendd.

JR-Morgan (approved):

adamhathcock (approved):

oguzhankoral:
> Should we be implementing our own Revit sync to thread, implemented using `APIContext.Run`? I foresee we'll need to do this for receive if not also for send.

Hmm, don't know yet the difference using the current thread or APIContext.Run. We can consider this when we start to implement revit receive

PR #3500, opened 2024-06-11
Feature [Revit]: Excel parameter updater support for shared type parameters

efdiloreto:
## Description & motivation

I opened a thread in the community some days ago because I couldn't update some parameters in the Revit schedules using the Excel Connector. Here is the original post:

https://speckle.community/t/revit-excel-parameter-updater-graying-out-custom-parameters-in-excel/11613

The problem was that Revit considers shared parameters as ScheduleField.Instance instead of ScheduleField.ElementType. Therefore, if the parameter was a "Type Parameter" and was obtained from the element, it would always be considered as "ReadOnly."

The solution was to obtain the parameter from the "Type" rather than from the element. Because of this, I added some logic in the code to handle Shared Parameters.

Reference: https://www.revitapidocs.com/2024/9888db7d-00d0-4fd7-a1a9-cdd1fb5fce16.htm

## Changes:

`ConvertView.Schedule -> AddColumnMetadataToDataTable`

## Screenshots:

Using the sample Architecture Project add a shared type parameter to the planting category:
image

Add the type and the shared parameters to the existing schedule and sort by type (not itemize every instance)
image

Then:
 1. Send to Speckle
 2. Receive from Excel
 3. Edit in Excel
 4. Send To Speckle
 5. Receive in Revit

## Checklist:

- [x ] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x ] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x ] My commits are related to the pull request and do not amend unrelated code or documentation.
- [ x] My code follows a similar style to existing code.

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertView.Schedule.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertView.Schedule.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using Autodesk.Revit.DB;
 using Objects.Organization;
 using RevitSharedResources.Models;
+using Speckle.Core.Api.SubscriptionModels;
 using Speckle.Core.Logging;
 using Speckle.Core.Models;
 using DB = Autodesk.Revit.DB;
@@ -244,11 +245,12 @@ public partial class ConverterRevit
 
     var columnMetadata = new Base();
     columnMetadata["BuiltInParameterInteger"] = info.field.ParameterId.IntegerValue;
-    columnMetadata["FieldType"] = info.field.FieldType.ToString();
-
+    string fieldType = info.field.FieldType.ToString();
+    
     Parameter param;
     if (info.field.FieldType == ScheduleFieldType.ElementType)
     {
+      
       if (firstType != null)
       {
         param = firstType.get_Parameter(builtInParameter);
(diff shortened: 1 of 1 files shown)


AlanRynne:
Hi @efdiloreto!

Thanks for this! We'll try to get someone to review this as soon as possible.

Meanwhile, I see the CI is failing because you're not following our formatting rules (which we recently started enforcing)

You can ensure the code is fully formatted by running:

`dotnet tool install` in the root of the repository and then `dotnet csharpier ./`

The rest of the repo is already compliant so you should only see your changes be formatted.

Let us know if you have any further issues getting this done :)

efdiloreto:
Thanks Alan, will try to format it ASAP

efdiloreto:
@AlanRynne I think is passing the formatting test, but not sure why is failing now.

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertView.Schedule.cs
```diff
@@ -257,10 +258,27 @@ Element firstElement
...
+
+          // If the parameter is readonly in the element but not in the type, is a type parameter
+          if (typeParam != null && !typeParam.IsReadOnly && param != null && param.IsReadOnly)
+          {
+            fieldType = ScheduleFieldType.ElementType.ToString();
+          }
+        }
+
+        // Check IsReadOnly for both
+        columnMetadata["IsReadOnly"] = (param?.IsReadOnly ?? false) && (typeParam?.IsReadOnly ?? false);
```
connorivy, line 281: I have some concerns about this line here. Let's say that the "firstElement" in question is a regular instance parameter (not shared). Param would have a value and typeParam would (at least sometimes) be null. In this situation, this statement would always evaluate to false, regardless of the value of "param.IsReadOnly" which doesn't seem correct to me.

What are your thoughts?

connorivy: I'm happy to merge this pr in if we can resolve this. To me it looks like we need split this line and move the type param check to the if statement on line 276. Again, please let me know what you think.

efdiloreto: Sorry for the delay in the response. I think you are right, but I haven't had the time to change it yet.

connorivy:
I made the discussed change and created a new PR https://github.com/specklesystems/speckle-sharp/pull/3585. If that one gets merged then we can close this.

PR #3501, opened 2024-06-12
Convert revit converters to use proxies for unit testing

adamhathcock:
https://spockle.atlassian.net/browse/DUI3-149

Revit types and proxies exist in new repo: https://github.com/specklesystems/speckle-sharp-host-apis
types and proxies are consumed via Nuget (preview at the moment)

DI registration occurs with new facility to scan repo and add to Autofac as transient objects

Converter assembly only references interfaces
New test assembly made which only references interfaces
Connector assembly references APIs and Interfaces to glue together with DI

Future work:
some clean up in the connector assembly
image

Need to be careful to ensure no breaking.  Probably around some casting?


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Speckle.Converters.RevitShared.projitems
AlanRynne, line 1: Same! Is this what is breaking your IDE experience?

AlanRynne (approved):
Looks good! just a couple comments but I'm approving meanwhile

PR #3504, opened 2024-06-12
DUI3-174 Cancel ops on doc swap

oguzhankoral:
UI implementation is ready to merge -> https://github.com/specklesystems/speckle-server/pull/2363

- We have command for each binding to set any message to UI as global toast notification
- I removed all old `ICancelable` interface from send and receive bindings. Which was something talking with @JR-Morgan
- I aligned naming of `_cancellationManager` in bindings. Some of still was having setter on `CancellationManager`
- I have dup code now on RevitSendBinding, AutocadSendBinding and AutocadReceiveBinding for document switch logic to cancel out all active operations and then tell UI to show message.
- Feel free to suggest better "message" on doc switch 🙂

Also would like to sync @clairekuang before merging in!


clairekuang:
Note: removed the autocad portion, as operations can't be safely cancelled from the document store. Autocad op cancellation will be addressed in a separate pr, where the event is subscribed to from within the op instead of the binding

clairekuang (approved):

PR #3506, opened 2024-06-13
Nesting arc gis layers on receive

KatKatKateryna:
On Receive, the created layers will be nested in groups. 1st group: name of Project and Model. Following group levels: the replicated structure of the received Commit object (e.g. Rhino Model > Layer1 > Polylines).

### DIFF
--- a/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
+++ b/DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
@@ -163,7 +163,10 @@ public class BasicConnectorBinding : IBasicConnectorBinding
     {
       if (member is Layer layer)
       {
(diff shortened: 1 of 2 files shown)


oguzhankoral (approved):
Feature wise it is a good improvement. But could you please create ticket to handle URL/Path's elegantly?

```c#
// sample 1
new($"{_contextStack.Current.Document.SpeckleDatabasePath.AbsolutePath.Replace('/', '\\')}\\{datasetId}");

// sample 2
string shortName = nestedLayerName.Split("\\")[^1];

// sample 3
string nestedLayerPath = string.Join("\\", nestedLayerName.Split("\\").SkipLast(1));

// sample 4
var allPathElements = nestedLayerPath.Split("\\").Where(x => !string.IsNullOrEmpty(x));
```

I believe there are more logics around these paths which deserves to be handled nicely. We were chatting on this before with @JR-Morgan too IIRC.

KatKatKateryna:
@oguzhankoral https://spockle.atlassian.net/browse/DUI3-317

PR #3509, opened 2024-06-14
DUI3-149 Level conversions should be optional?  Not all!

adamhathcock:
fixes related to DUI3-149


DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Helpers/ParameterValueExtractor.cs
```diff
@@ -164,7 +164,7 @@ public IRevitElementId GetValueAsElementId(IRevitElement element, RevitBuiltInPa
   {
     if (!TryGetValueAsElementId(element, builtInParameter, out var elementId))
     {
-      throw new SpeckleConversionException();
```
AlanRynne, line 167: Was this the original code? or a change you made when doing the testing changes?

AlanRynne (approved):

PR #3510, opened 2024-06-14
enable lock files for dui3

adamhathcock:
enable nuget lock files for dui3

- this will allow us to better monitor change in dependencies
- use lock files for restoration for repeatable builds
- cache nuget and restore for faster builds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
     outputs:
       version: ${{ steps.set-version.outputs.version }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,15 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 8.0.2xx # Align with global.json (including roll forward rules)
+      
+      - name: Cache Nuget     
+        uses: actions/cache@v4
+        with:
+          path: ~/.nuget/packages
+          key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json') }}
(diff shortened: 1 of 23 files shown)


AlanRynne (changes requested):
I think we should also modify the CI to run in locked mode since we're at it

.github/workflows/ci.yml
```diff
@@ -12,32 +12,24 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-    outputs:
-      version: ${{ steps.set-version.outputs.version }}
+    runs-on: windows-latest
```
AlanRynne, line 15: 💔

.github/workflows/main.yml
```diff
@@ -9,10 +9,47 @@ on:
...
+      - name: ⬆️ Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: output-${{ env.GitVersion_FullSemVer }}
+          path: output/*.*
+          compression-level: 0 # no compression
 
+      - id: set-version
+        name: Set version to output
+        run: echo "version=${{ env.GitVersion_FullSemVer }}" >> "$GITHUB_OUTPUT"
```
AlanRynne, line 49: Why the copy/paste? reusable workflows are ♥️
If we need some variation we can add inputs to it.

.github/workflows/main.yml
```diff
@@ -9,10 +9,47 @@ on:
...
+        
   deploy-installers:
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
```
AlanRynne, line 52: This doesn't need to be windows at all 🙂

AlanRynne (approved):

adamhathcock:
For https://spockle.atlassian.net/browse/DUI3-311

adamhathcock:
DUI3-311

PR #3514, opened 2024-06-17
Update Model.cs

didimitrie:
Removes incorrect usage of chunkable in structural analysis model class. This is causing generating large data chunks with embedded objects inside of them. 

see: https://latest.speckle.systems/projects/ae810b79c9/models/88ebf44a96 - and a specific bad chunk looks like this: https://latest.speckle.systems/projects/ae810b79c9/models/932fc3a0bb07f34d82b850b8a97d3668


didimitrie:
closing as it's just a band-aid - deeper thunking (sic) and refactoring is needed on the struct connectors

PR #3517, opened 2024-06-17
extracting some non-controversial changes to be merged first for clea…

JR-Morgan:
some quick and easy to review changes picked from https://github.com/specklesystems/speckle-sharp/pull/3515 that I want to merge in first to make the diff cleaner


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitBaseBinding.cs
```diff
@@ -8,13 +8,13 @@ namespace Speckle.Connectors.Revit.Bindings;
 internal abstract class RevitBaseBinding : IBinding
```
oguzhankoral, line 8: I don't know why we have it at all but it's ok.

oguzhankoral (approved):

PR #3522, opened 2024-06-17
better fallback reporting

KatKatKateryna:
If we don't throw exception here, later in the hostApp converter we won't be able to tell which object was it initially and report properly.

### DIFF
--- a/DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ConverterWithFallback.cs
+++ b/DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ConverterWithFallback.cs
@@ -1,4 +1,5 @@
-﻿using Speckle.Converters.Common.Objects;
+using System.Collections;
+using Speckle.Converters.Common.Objects;
 using Speckle.Core.Models;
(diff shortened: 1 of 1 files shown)


DUI3-DX/Sdk/Speckle.Converters.Common.DependencyInjection/ToHost/ConverterWithFallback.cs
```diff
@@ -49,6 +50,10 @@ public object Convert(Base target)
     var displayValue = target.TryGetDisplayValue<Base>();
     if (displayValue != null)
     {
+      if (displayValue is IList && !displayValue.Any())
```
oguzhankoral, line 53: This is ideally should be prevented by sender but I don't see why not to report it

AlanRynne: This is good, there is no guarantee an object will be correctly generated as they can come from anywhere. Sending a wall without a displayValue is _doable_ but not something our connectors would do (except Grasshopper/Dynamo which technically could)

AlanRynne (approved):

PR #3526, opened 2024-06-19
D ui3 123 receiving non native attributes revit params gh attributes etc

KatKatKateryna:
UPD: receiving of non-native properties has been turned into a placeholder until we find better strategy for sending props from host apps. 

Purpose is to preserve objects properties when placing them into GIS layers(tables).
Current approach is to iterate through all objects that will be placed under the same layer and collect their property names and their Type (e.g. string, int) while defaulting to string if nothing better fits).
The opinionated choice of non-native properties to receive is:
- all properties that are not Base: Dynamic for "Objects.Geometry" and Typed for everything else
- Base properties if its Revit "properties" or Rhino "userStrings"

P.S. atm the following exception is being generated inconsistently, needs better understanding. Thrown during "Conversion" step, so likely not a problem with the current PR: 
image

image


adamhathcock (approved):

KatKatKateryna:
@adamhathcock I tested with the new function, and "default" value doesn't work very well (e.g. it assigns a random date like "{01/01/0001 00:00:00}" ). So I changed it to null and removed a redundant null check from the method calling this one.

PR #3529, opened 2024-06-21
D ui3 305 agis implement highlight for individual features e.g. from receive report

KatKatKateryna:
1. Introduce struct holding LayerURI (represented as string because this is how native API is treating LayersURI), featureId, MapMember (actual layer object on the map)
2. Adapt Report building and Highlight functionality to use this instead of flat string with LayerURI
3. Add functionality to select individual features if FeatureID is available


DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/ObjectID.cs
```diff
@@ -0,0 +1,41 @@
+using ArcGIS.Desktop.Mapping;
+
+namespace Speckle.Connectors.ArcGIS.Utils;
+
+public struct ObjectID
```
oguzhankoral, line 5: I would omit MapMember from this struct, it is confusing why objectID checks for MapMember as below. 

```c#
if (objectId.MapMember == null)
{
   continue;
}
``` 

What I expect from `ObjectID` is just id, if you include MapMember it becomes something else. In this case, we might need another struct that has MapMember too instead considering its nullability on `ObjectID` struct.

oguzhankoral (changes requested):
Overall, I liked the extraction on object id which was always confusing for me, maybe I'm greedy but I believe we have an opportunity to have it more clean way. Then good to go 🚀

KatKatKateryna:
converting to draft to update documentation

oguzhankoral (approved):
Tested and LGTM 🚀

PR #3531, opened 2024-06-24
DUI3-325 Add Civil3d 2024 connector

clairekuang:
## Description & motivation
 Adds the Civil3d 2024 connector, with just a send binding, to the Autocad connector folder. Also adds a corresponding Autocad2024 converter so that civil3d connector can send out of the box.

## Changes:

- Noteworthy changes are a separate `Civil3dConnectorModule` in the connector to allow for Civil3d bindings to differ from Autocad, and preprocessor directives in `AutocadCommand` to create an `AutocadSettings` with the correct host app and version

## Screenshots:
Sent a point from Civil3d: [https://app.speckle.systems/projects/b53a53697a/models/719612f9da](https://app.speckle.systems/projects/b53a53697a/models/719612f9da)


Core/Core/Kits/Applications.cs
```diff
@@ -64,6 +64,7 @@ public static class HostApplications
     Unity = new("Unity", "unity"),
     GSA = new("GSA", "gsa"),
     Civil = new("Civil 3D", "civil3d"),
```
oguzhankoral, line 66: do we need this one if we added Civil3D?

Core/Core/Kits/Applications.cs
```diff
@@ -124,6 +125,11 @@ public static HostApplication GetHostAppFromString(string? appname)
       return Civil;
     }
 
+    if (appname.Contains("civil3d"))
```
oguzhankoral, line 128: what is the difference with "civil"? because this will always return the previous one I believe because it will contain already "civil"

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/AutocadConnectorModule.cs
```diff
@@ -1,3 +1,4 @@
+#if AUTOCAD
```
oguzhankoral, line 1: It might make sense to put note here why we do this

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/Civil3dConnectorModule.cs
oguzhankoral, line 1: We can have shared module for autocad verticals, seems we are having many dups

oguzhankoral (changes requested):
Overall looks good and good job!
Few comments maybe small changes needed.

Core/Core/Kits/Applications.cs
clairekuang: Yes, to prevent breaking the DUI2 connectors

Core/Core/Kits/Applications.cs
clairekuang: Ah good point, will fix this

Core/Core/Kits/Applications.cs
oguzhankoral: I believe we should start to consider have new Applications(?) class for DUI3 only instead populating old.

DUI3-DX/Connectors/Autocad/Speckle.Connectors.AutocadShared/DependencyInjection/SharedConnectorModule.cs
```diff
@@ -0,0 +1,40 @@
...
+    builder.AddSingleton(new AutocadDocumentManager()); // TODO: Dependent to TransactionContext, can be moved to AutocadContext
+    builder.AddSingleton<DocumentModelStore, AutocadDocumentStore>();
+    builder.AddSingleton<AutocadContext>();
+    builder.AddSingleton<AutocadLayerManager>();
+    builder.AddSingleton<AutocadIdleManager>();
+
+    // Register bindings
+    builder.AddSingleton<IBinding, TestBinding>();
+    builder.AddSingleton<IBinding, AccountBinding>();
+    builder.AddSingleton<IBinding, AutocadBasicConnectorBinding>();
```
oguzhankoral, line 36: I believe this is unnecessary registration since we register it on next line as `IBasicConnectorBinding`

clairekuang: This was preexisting, so I'm leaving it here for now

oguzhankoral: fair enough

Core/Core/Kits/Applications.cs
oguzhankoral: FYI: @BovineOx added as subtask under the other relevant story (assigned to Ian) https://spockle.atlassian.net/browse/DUI3-431

oguzhankoral (approved):

PR #3533, opened 2024-06-25
V3 core sdk via nuget

adamhathcock:
https://spockle.atlassian.net/browse/DUI3-35

takes the intermediate step of referencing core from the new repo and removing core from the solution filter.

Some changes to logging help with dependencies


DUI3-DX/Connectors/Autocad/Speckle.Connectors.Civil3d2024/Speckle.Connectors.Civil3d2024.csproj
```diff
@@ -19,7 +19,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="7.0.0" />
-    <PackageReference Include="Speckle.AutoCAD.API" Version="2024.0.0" />
```
adamhathcock, line 22: @clairekuang is this okay?

DUI3-DX/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/packages.lock.json
```diff
@@ -404,7 +434,7 @@
         "dependencies": {
           "Serilog.Extensions.Logging": "[7.0.0, )",
           "Speckle.Autofac": "[2.0.999-local, )",
-          "Speckle.Core": "[2.0.999-local, )"
+          "Speckle.Core": "[3.0.1-alpha.11, )"
```
BovineOx, line 437: Should we have some rules around what we should pick up? i.e. accept minor versions and bug fixes? Maybe this isn't the right place and we perhaps don't need to do it right away

BovineOx (approved):

BovineOx:
I mean it's all a bit hard to read but updating reference to the nuget essentially, mainly just:
- Should we be taking minor and bug fix improvements automatically?

I did have another but it's gone...

PR #3536, opened 2024-06-26
Manage V3 versions centrally (exception AutoCAD)

adamhathcock:
Manage V3 versions centrally (exception AutoCAD as there are multiple versions and not wrapped yet)

requires https://github.com/specklesystems/speckle-sharp/pull/3533

also includes sourcelink and gubbins for more nuget checkboxes


BovineOx (commented):
Same question about taking minor versions etc...  I think @clairekuang needs to approve ask questions to go ahead

BovineOx (approved):

Issue #3530, opened 2024-06-24
Revit: Area of cut openings such as windows and walls in host elements is not available when converted.


ks-cph:
## Prerequisites

- [x] I read the [contribution guidelines](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)
- [x] I checked the [documentation](https://speckle.guide/) and found no answer.
- [x] I checked [existing issues](../issues?q=is%3Aissue) and found no similar issue. 
- [x] I checked the [community forum](https://speckle.community/) for related discussions and found no answer.
- [x] I'm reporting the issue to the correct repository (see also [speckle-server](https://github.com/specklesystems/speckle-server), [speckle-sharp](https://github.com/specklesystems/speckle-sharp), [specklepy](https://github.com/specklesystems/specklepy), [speckle-docs](https://github.com/specklesystems/speckle-docs), and [others](https://github.com/orgs/specklesystems/repositories))

## What package are you referring to?
server (backend)

## Describe the bug
When converting walls with hosted elements, e.g. windows and doors, the area of these openings is relevant for different use cases. However, no parameter converted represents the area of the hosted element within the host. While we expected that the parameter `HOST_AREA_COMPUTED` would correspond to this desired value, we found out that this is not always the case (see https://thebuildingcoder.typepad.com/blog/2015/03/ifcexportutils-methods-determine-door-and-window-area.html). 

## To Reproduce
Export from Revit to Speckle.

## Expected behavior
We would expect a parameter representing the area of the hosted object within the host as illustrated in the following image.
## Screenshots

![Description](https://github.com/specklesystems/speckle-sharp/assets/86473259/bd7f96f7-7ce7-46d7-ac2e-a4df0b9ada08)

PR #3539, opened 2024-06-27
Add hosted element area to parameters in Revit.

ks-cph:
## Description & motivation

Refers to Issue #3530 
This change adds the area of hosted elements within a host to a parameter. So far, there is no such parameter available. This becomes especially important for windows and doors where the area of removed elements can be essential. For instance, prefabricated timber elements are normally produced as a whole and then these parts are removed, thus, are somewhat waste (even though they may be re-used).

Others will benefit from this change as such essential information is not yet available in Speckle.

## Changes:

- Add a method to get the Area of a Hosted element in a Host
- Add the area as a new Parameter to the Base object

## To-do before merge:

- [ ] The code requires the RevitAPIIFC.dll

## Validation of changes:
The contribution has been verified with several Revit models where windows and doors are added to walls.

## Checklist:

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevit2023/ConverterRevit2023.csproj
+++ b/Objects/Converters/ConverterRevit/ConverterRevit2023/ConverterRevit2023.csproj
@@ -24,12 +24,6 @@
     <PackageReference Include="Speckle.Revit.API" Version="2023.0.0" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Reference Include="RevitAPIIFC">
-      <HintPath>..\..\..\..\..\..\..\..\..\..\Program Files\Autodesk\Revit 2023\RevitAPIIFC.dll</HintPath>
-    </Reference>
-  </ItemGroup>
-
   <Target Name="CopyTemplatesToKitfolder" AfterTargets="CopyToKitFolder" Condition="'$(IsDesktopBuild)' == true">
     <Message Text="Copying templates to kit folder" />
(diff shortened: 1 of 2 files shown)


teocomi:
Hey @ks-cph @martinromby thanks for the ping, sorry this slipped through!
@bimgeek is looking to add a review for this in the coming dev cycles and we'll keep you posted 🙏
Thanks again for your contributions!

martinromby:
Any feedback for us in the review @bimgeek ?

didimitrie:
From a quick glance: my hesitation centres around performance as we're introducing a new calculation to this. I appreciate the need for precision in your case, but how strong is it?

didimitrie:
note: this breaks sending of hosted elements that are not on a wall, will fix

didimitrie:
Fixed the last few bits; and added a poc comment around the units - cutout area is always extracted out in meters, but that's all i have bandwidth for now. Maybe @clairekuang can help set them correctly? 

We can consider merging this if @bimgeek discovers no large performance implications. Regarding our next gen connectors, how important is this number for y'all @martinromby? Since it's doing an extra calculation, we would extract this out as an opt-in extra setting, but ideally we'd not extract it out at all!

ks-cph:
Having this knowledge, especially on windows and doors, is essential for reliable quantity-takeoff. We would therefore highly appreciate it if you could merge this in - potentially with the opt-in option.

ks-cph:
Dear @didimitrie,

Have you made any decisions on how to progress with this PR, yet? It is a very relevant topic for us, and we have many users who would highly appreciate it if this feature could be supported. The opt-in solution works just fine for us!

Could you please let us know if we can expect this to be included at some point? Otherwise, we will look for alternative solutions.

ks-cph:
Any update with this issue? The latest connector version still reports the Host_Area_Compute as area.

PR #3543, opened 2024-06-28
Host app non-central managed

adamhathcock:
Now we're locally moving host app versions to converters/connectors.  Everything else is centrally managed


AlanRynne (dismissed):
Is this really required? I fear we're going to end with +20 Package.props files, which is just going to make everything more confusing.

adamhathcock:
> Is this really required? I fear we're going to end with +20 Package.props files, which is just going to make everything more confusing.

Well, it's either this or get rid of Central management altogether.  It's already doing this, btw it's just now reflection in the solution.

AlanRynne:
I'm not sure we need or want to centrally manage every package in a way that's not centralized.

My take on this would be:
- We centrally manage everything that **has to have a consistent version** across the entire monorepo
- We manually manage dependencies that are "per-project" (host API versions specifically)

I'm not sure what's the benefit of having different `Directory.Packages.props` files just for 2 projects at a time. Specially considering the converter also needs the same version as the connector and it lives elsewhere.

We seem to already have 3 different package.props files just for autocad alone.

That being said, I'm happy to be convinced otherwise 😉

AlanRynne (approved):
Live reviewed with Adam with some added changes 🙂

PR #3544, opened 2024-06-28
DUI3-71 receive bindings and receive geometry

connorivy:
#### Connectors / Revit
- Added receive binding (pretty much copied from Rhino)
- Modified service registrations for new types
  - Added default traversal function
- Registered a run on thread method that uses the Revit Async package
  - Running on main thread alone is not sufficient
- Added simple host object "builder" that just opens a transaction, traverses, receives, and then commits the transaction
- Copied transaction manager from dui2 connector

#### Converter / Revit
- added import method next to export method on referencePointConverter (and interface)
- Changed name of 'RevitRootToHostConverter' which implements 'IRootToSpeckleConverter' to accurately reflect the implemented interface (aka replaced ToHost with ToSpeckle)
- Added toHost scaling service to mirror toSpeckle scaling service

#### Converter / Revit / ToHost
- implemented ITypedConverter interface for several types of geometries
- added a modelCurve topLevel converter to be able to actually use the typedConverters

#### Converter / Revit / ToSpeckle
- I moved a bunch of files into this 'toSpeckle' folder to mirror the folder structure of Rhino

## Validation of changes:

receiving works for model lines of various types

## Checklist:


DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
```diff
@@ -0,0 +1,88 @@
...
+      return;
+    }
+  }
+
+  private void OnSendOperationProgress(string modelCardId, string status, double? progress)
+  {
+    Commands.SetModelProgress(modelCardId, new ModelCardProgress(modelCardId, status, progress));
+  }
+
+  public void CancelSend(string modelCardId) => _cancellationManager.CancelOperation(modelCardId);
```
oguzhankoral, line 87: `CancelSend` not needed in receive binding

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+    // Catch here specific exceptions if they related to model card.
+    catch (OperationCanceledException)
+    {
+      // SWALLOW -> UI handles it immediately, so we do not need to handle anything
+      return;
+    }
+  }
+
+  private void OnSendOperationProgress(string modelCardId, string status, double? progress)
```
oguzhankoral, line 82: pls update your PR with recent merge and this function should go away and its usage will be updated on line 63

oguzhankoral (dismissed):
Only comments on binding for now, will test converters later and review again.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/RevitConnectorModule.cs
```diff
@@ -49,20 +50,24 @@ public void Load(SpeckleContainerBuilder builder)
...
 
     // send operation and dependencies
     builder.AddScoped<SendOperation<ElementId>>();
     builder.AddScoped<IRootObjectBuilder<ElementId>, RevitRootObjectBuilder>();
-
-    // register send conversion cache
     builder.AddSingleton<ISendConversionCache, SendConversionCache>();
+
+    // receive operation and dependencies
+    builder.AddScoped<ReceiveOperation>();
```
oguzhankoral, line 68: This is not needed because defined in `builder.AddConnectorUtils();` https://github.com/specklesystems/speckle-sharp/blob/cd39d867065b53c111ff02adbe66642780cc863f/DUI3-DX/Sdk/Speckle.Connectors.Utils/ContainerRegistration.cs#L17

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+  private HostObjectBuilderResult BakeObjects(IEnumerable<TraversalContext> objectsGraph)
+  {
+    var conversionResults = new List<ReceiveConversionResult>();
+    var bakedObjectIds = new List<string>();
+
+    foreach (TraversalContext tc in objectsGraph)
+    {
+      try
+      {
+        YieldToUiThread();
```
oguzhankoral, line 63: Do we really need this since we run already Dispacther.Background on the CefSharpPanel?

https://github.com/specklesystems/speckle-sharp/blob/7e7db96a1b5c23e527766dc6d2cf6cc87f399cb5/DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Plugin/CefSharpPanel.xaml.cs#L15-L16

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitRootToSpeckleConverter.cs
```diff
@@ -6,12 +6,12 @@
 namespace Speckle.Converters.RevitShared;
 
 // POC: maybe possible to restrict the access so this cannot be created directly?
-public class RevitRootToHostConverter : IRootToSpeckleConverter
```
oguzhankoral, line 9: Hmm, confused why it is renamed? And do we still have `RevitRootToHostConverter` ? 🤔

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/ICurveConverterToHost.cs
```diff
@@ -0,0 +1,128 @@
...
+        curveArray.Append(_ellipseConverter.Convert(ellipse));
+        return curveArray;
+
+      case SOG.Spiral spiral:
+        return _polylineConverter.Convert(spiral.displayValue);
+
+      case SOG.Curve nurbs:
+        var n = _curveConverter.Convert(nurbs);
+
+        // poc : in original converter, we were passing a bool into this method 'splitIfClosed'.
```
oguzhankoral, line 66: IIRC @AlanRynne was explaining something similar to this before.. by saying smth like "there is no closed curve definition.. (maybe I'm wrong)"

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
connorivy: hmm that is a good question. I'll leave it out for now. Once we can send larger models, we will find out if we need it

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/RevitRootToSpeckleConverter.cs
connorivy: I renamed it because the class name, RevitRootToHostConverter, didn't match the implemented interface, RootToSpeckleConverter.

The IRootToHostConverter is actually added in the `AddConverterCommon` method

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/ICurveConverterToHost.cs
AlanRynne: Revit does not support closed nurbs curves/splines. So the split logic needs to come back one way or another.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ScalingServiceToHost.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+      return value;
+    }
+
+    return UnitUtils.ConvertToInternalUnits(value, UnitsToNative(units));
+  }
+
+  public ForgeTypeId UnitsToNative(string units)
+  {
+    var u = Core.Kits.Units.GetUnitsFromString(units);
```
BovineOx, line 20: I thought we had other ways of doing this conversion? I think in other places we're using a dictionary which feels a bit more terse?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/ArcConverterToHost.cs
```diff
@@ -0,0 +1,57 @@
...
+    _pointToXyzConverter = pointToXyzConverter;
+    _scalingService = scalingService;
+    _planeConverter = planeConverter;
+  }
+
+  public DB.Arc Convert(SOG.Arc target)
+  {
+    double startAngle,
+      endAngle;
+    if (target.startAngle > target.endAngle)
```
BovineOx, line 28: whitespace between engAngle and if plz, can those doubles be one line or two statements?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/ArcConverterToHost.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+      startAngle = (double)target.endAngle;
+      endAngle = (double)target.startAngle;
+    }
+    else
+    {
+      startAngle = (double)target.startAngle.NotNull();
+      endAngle = (double)target.endAngle.NotNull();
+    }
+    var plane = _planeConverter.Convert(target.plane);
```
BovineOx, line 38: whitespace, this is just me asking for us to ensure we have a line space between the start and end of any scope

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/ICurveConverterToHost.cs
BovineOx: If it's complex I am okay this being broken into a separate ticket (DUI3 project) and the link dropped here

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
```diff
@@ -0,0 +1,96 @@
...
+  {
+    _lineConverter = lineConverter;
+    _scalingService = scalingService;
+    _contextStack = contextStack;
+  }
+
+  public CurveArray Convert(Polyline target)
+  {
+    var curveArray = new CurveArray();
+    if (target.value.Count == 6)
```
BovineOx, line 30: guessing this was here before but could we comment why 6 is the magic number and this is correct just for future readers, i.e. if the Count (and that count is of) is 6 this means that blah blah...

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
```diff
@@ -0,0 +1,96 @@
...
+  /// It will also convert the line to Revit before appending it to the <see cref="CurveArray"/>.
+  /// </summary>
+  /// <param name="curveArray">The revit <see cref="CurveArray"/> to add the line to.</param>
+  /// <param name="line">The <see cref="SOG.Line"/> to be added.</param>
+  /// <returns>True if the line was added, false otherwise.</returns>
+  public bool TryAppendLineSafely(CurveArray curveArray, SOG.Line line)
+  {
+    if (IsLineTooShort(line))
+    {
+      // poc : logging "Some lines in the CurveArray where ignored due to being smaller than the allowed curve length."
```
BovineOx, line 82: ahh hmmmm this is one of those grey areas. I presume we just brought this over but we should make some solid decision about whether this is valid as it's an effective mutation. Please discuss with @clairekuang and we can probably make a ticket and drop it here, even if that ticket is. Let's think about this later and either do something or close the ticket and remove the POC

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
```diff
@@ -0,0 +1,96 @@
...
+      return false;
+    }
+    try
+    {
+      curveArray.Append(_lineConverter.Convert(line));
+      return true;
+    }
+    catch (Exception ex) when (!ex.IsFatal())
+    {
+      // poc : logging
```
BovineOx, line 92: @JR-Morgan not sure we could get anything meaningful into logs or the report from here??

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
```diff
@@ -0,0 +1,96 @@
...
+  /// <param name="line">The <see cref="SOG.Line"/> to be added.</param>
+  /// <returns>True if the line was added, false otherwise.</returns>
+  public bool TryAppendLineSafely(CurveArray curveArray, SOG.Line line)
+  {
+    if (IsLineTooShort(line))
+    {
+      // poc : logging "Some lines in the CurveArray where ignored due to being smaller than the allowed curve length."
+      return false;
+    }
+    try
```
BovineOx, line 85: can we have a whitespace here also, more importantly...
- Why the try catch?
- Why are we swallowing siently?

I know we're bring this stuff across but I think we should review some of these slightly bogus choices. If the convert() files, why are we continuing?

@JR-Morgan has strong feelings on this and @clairekuang I think you and @didimitrie need to establish the whole fail policy, as this is succeed but some things went bad and oh we didn't tell you. I don't think that's the right approach tbh...

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ModelCurveToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+      else
+      {
+        revitCurve = _contextStack.Current.Document.Create.NewModelCurve(
+          curve,
+          NewSketchPlaneFromCurve(curve, _contextStack.Current.Document)
+        );
+      }
+    }
+
+    return revitCurve.NotNull();
```
BovineOx, line 55: So we're saying, we expected to make a curve but this didn't happen. I think that's probably but I see we might make many curves, is it the case that we expect to make at least one curve?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ModelCurveToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+  /// <summary>
+  /// Credits: Grevit
+  /// Creates a new Sketch Plane from a Curve
+  /// https://github.com/grevit-dev/Grevit/blob/3c7a5cc198e00dfa4cc1e892edba7c7afd1a3f84/Grevit.Revit/Utilities.cs#L402
+  /// </summary>
+  /// <param name="curve">Curve to get plane from</param>
+  /// <returns>Plane of the curve</returns>
+  private DB.SketchPlane NewSketchPlaneFromCurve(DB.Curve curve, DB.Document doc)
+  {
+    DB.XYZ startPoint = curve.GetEndPoint(0);
```
BovineOx, line 67: start and end are always at indexes 0 and 1 ?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ModelCurveToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+      endPoint = curve.GetEndPoint(i);
+      i++;
+    }
+
+    // Plane to return
+    DB.Plane plane;
+
+    // If Z Values are equal the Plane is XY
+    if (startPoint.Z == endPoint.NotNull().Z)
```
BovineOx, line 82: only Z may be null?

BovineOx (changes requested):
Seems pretty good. I have some concern around the correct lifetime for the transaction (but i could be fine). I've also some points where I'd like @clairekuang to be involved and perhaps for between you, you raise any necessary tickets for gaps.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
```diff
@@ -0,0 +1,82 @@
...
+using Speckle.Connectors.Utils.Cancellation;
+using Speckle.Connectors.Utils.Operations;
+using Speckle.Autofac.DependencyInjection;
+using Speckle.Connectors.Utils;
+
+namespace Speckle.Connectors.Revit.Bindings;
+
+internal class RevitReceiveBinding : IReceiveBinding
+{
+  public string Name => "receiveBinding";
```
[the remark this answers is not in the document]
oguzhankoral: All same, I believe we will have some abstract binding classes to have commons instead implementing for each

BovineOx:
Oh yeah and the Dispose()/lack of using/disabling the warning

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/TransactionManager.cs
```diff
@@ -0,0 +1,163 @@
+using Autodesk.Revit.DB;
+using Speckle.Converters.RevitShared.Helpers;
+
+namespace Speckle.Connectors.Revit.Operations.Receive;
+
+/// <summary>
+/// Is responsible for all functionality regarding subtransactions, transactions, and transaction groups.
+/// This includes starting, pausing, committing, and rolling back transactions
+/// </summary>
+public sealed class TransactionManager : IDisposable
```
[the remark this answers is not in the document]
connorivy: this is scoped. Certain converters sometimes need to pause an ongoing transaction or startup a subtransaction, so they need access to the same instance.

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+    string modelName,
+    Action<string, double?>? onOperationProgressed,
+    CancellationToken cancellationToken
+  )
+  {
+    var objectsToConvert = _traverseFunction
+      .TraverseWithProgress(rootObject, onOperationProgressed, cancellationToken)
+      .Where(obj => obj.Current is not Collection);
+
+    _transactionManager.StartTransactionGroup($"Received data from {projectName}");
```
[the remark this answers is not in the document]
connorivy: this object gets created by the generic "ReceiveOperation" as an IHostObjectBuilder which is not a disposable interface. 
I've wrapped this method in a try finally to dispose the transaction manager instead. I'm happy to hear other suggestions if that isn't satisfactory.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
connorivy: The long-term intention is that we wouldn't swallow silently, as indicated by the poc : logging comment. Have we decided on a method of logging that we are happy with while I was gone? If so, then I'll change these lines.

My assumption is that we added these guard clauses and try-catch statements for a reason, so I left them here. I could just remove them if we think that is a better idea.

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
JR-Morgan: I would like us to discuss if we think a "partial conversion" with missing curves is a good UX here.

If we agree that there is a need for it, @didimitrie and I have already discussed how we would bring a "succeeded with warnings" system to converters, and this is where logging should exist, not directly in converters.

However I strongly object to having any indescriminate `catch(Exception ex) when (!ex.IsFatal())` involved inside specific conversions. We should decide the expected cases that we're expecting, and let unexpected ones fail the entire object. It gets very messy if we are were to allow unpredictable failures at many levels of our application

JR-Morgan (changes requested):

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
JR-Morgan: If @didimitrie and @BovineOx agree, I'm not sure we're ready yet do "succeeded with warnings", so for the sake of this pr, I would suggest we err on the side of failing the entire object, then we can always being a more granular conversion when we have the appropriate tooling for informing the user

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ModelCurveToSpeckleTopLevelConverter.cs
connorivy: correct

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ModelCurveToSpeckleTopLevelConverter.cs
connorivy: not the z value, the notNull check is on the entire point object

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
connorivy: okay, try catch removed 👍

DUI3-DX/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/TransactionManager.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  public void StartTransaction()
+  {
+    if (_transaction == null || !_transaction.IsValidObject || _transaction.GetStatus() != TransactionStatus.Started)
+    {
+      _transaction = new Transaction(Document, "Speckle Transaction");
+      var failOpts = _transaction.GetFailureHandlingOptions();
+      // POC: make sure to implement and add the failure preprocessor
```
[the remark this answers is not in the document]
BovineOx: Can you drop he DUI ticket made in the DUI3 project (will have DUI3 pre-fix) here in the comment please

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/Services/ScalingServiceToHost.cs
BovineOx: I was referring to this class? does the achieve the same thing?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
BovineOx: Can we have the comment?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
BovineOx: I guess Claire was away today but can you pick it afterwards?

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ModelCurveToSpeckleTopLevelConverter.cs
BovineOx: Ahh of course and I guess it only needs doing once

BovineOx (changes requested):
There's still a couple of points

DUI3-DX/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/PolylineConverterToHost.cs
connorivy: I'm not sure when this was first implemented, but clearly someone ran into the issue of trying to import a polyline with one segment that was so short that it couldn't be modeled in Revit. What we are doing here makes a lot of sense to me, just ignore the really small segment. The tolerance is .8mm so it seems unlikely that a user will even miss this segment of their polyline

connorivy: I can definitely talk to claire about it though

BovineOx (approved):

PR #3546, opened 2024-06-29
AGIS: put methods under the native struct

KatKatKateryna:


adamhathcock:
moved https://github.com/specklesystems/speckle-sharp-connectors/pull/5

PR #3547, opened 2024-06-30
DUI3 289 investigate and create versioning poc

BovineOx:
This is related to the PR draft here: https://github.com/specklesystems/speckle-sharp-sdk/pull/7

The first step with object upgrading had to be the (unexpected) act of excising Kits from the SDK. Most of Kits has now gone, what remains appears to be largely unrelated to kits.

Type Caching is now done with a new ITypeCache, an AbstractTypeCache and and ObjectsTypeCache. In theory, because this is injected down to Receive, it is possible to continue doing a kits like thing, by passing in a different TypeCache. By default the ObjectsTypeCache is caching all types that are derived from Base and found in the assembly where ObjectsTypeCache exists and the base, i.e. looking in Core and Objects. This might be something we can do a bit more elegantly but it's definitely no worse than existing.

Note that some changes may relate to working with the SDK and the connectors at the same time.


BovineOx:
I'd like to discuss the wisdom of checking the lock files in - I believe they are auto-generated and it's adding some unnecessary noise

PR #3557, opened 2024-07-03
revert version recording to be powershell

adamhathcock:


adamhathcock:
Don't need this anymore as old repo doesn't do this CI, only new repo will: https://github.com/specklesystems/speckle-sharp-connectors

PR #3566, opened 2024-07-09
Disable signed builds on non-main/dev branches

adamhathcock:
https://spockle.atlassian.net/browse/CNX-9818

### DIFF
--- a/.circleci/scripts/config-template.yml
+++ b/.circleci/scripts/config-template.yml
@@ -218,6 +218,7 @@ jobs: # Each project will have individual jobs for each specific task it has to 
                 environment:
                   SLUG: << parameters.slug >>
       - when: # Setup certificates and build installers signed for tagged builds that are in main or dev
+          condition:
             and:
               -  << pipeline.git.tag >>
               - or:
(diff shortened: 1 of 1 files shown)


.circleci/scripts/config-template.yml
```diff
@@ -203,17 +203,27 @@ jobs: # Each project will have individual jobs for each specific task it has to
...
           command: if [ "$CIRCLE_PR_REPONAME" ]; then circleci-agent step halt; fi
-      - unless: # Build installers unsigned on non-tagged builds
-          condition: << pipeline.git.tag >>
+      - unless: # Build installers unsigned on non-tagged builds that aren't in main or dev
+          condition: 
+            and:
+              -  << pipeline.git.tag >>
+              - or:
+                -  equal: [ main, << pipeline.git.branch >> ]
+                -  equal: [ dev, << pipeline.git.branch >> ]
```
AlanRynne, line 212: Maybe just signing from `main` will do, any releases from `dev` should be pre-releases anyway.

AlanRynne (commented):
This is a nice compromise :) Maybe we just sign tags from main with DigiCert though?

WDYT

AlanRynne:
Archicad CI is failing but it's unrelated to this ticket.

Installers in `dev` and `main` work fine, but our branch selection for the `ci-tools` repo falls back onto `main` when a matching branch does not exist, and the changes that this needs to pass are in `dev`.

Not a huge issue, we can ensure this passes by creating a matching branch on the `ci-tools` repo and re-running the workflow.

PR #3572, opened 2024-07-15
fix(Revit): DirectShape Instances to Speckle Conversion support

jsdbroughton:
DirectShape Geometry instances are nested Mesh and Solid within a GeometryInstance objetc. Previous conversion had switch case handlers only for the geometry types, this adds a recursion to cater for nested instances.


AlanRynne (commented):
This looks good, but it opens the door to non-meshes to end up being part of the directShape, which will be ignored when creating the Speckle DirectShape as we have the following validation call 👇🏼

Not sure if every GeometryInstance will fit in this 👇🏼

Screenshot 2024-09-13 at 16 27 24

jsdbroughton:
> Not sure if every GeometryInstance will fit in this 👇🏼

Ultimately, as this the nested GeometryInstance handling is recursive if a GeometryInstance, the method will still only return a Mesh or a Solid, never a GeometryInstance itself

jsdbroughton:
This fix will never make it to v2 i suppose. nevertheless, it might be of interest to whoever owns the v3 Revit @clairekuang @oguzhankoral

JR-Morgan (approved):

PR #3588, opened 2024-07-25
Cnx 9910 v2 needs better open uri handling

adamhathcock:
https://spockle.atlassian.net/browse/CNX-9910

new Open functions taken from: https://blog.reeset.net/archives/2923

### DIFF
--- a/Core/Core/Helpers/Open.cs
+++ b/Core/Core/Helpers/Open.cs
@@ -6,14 +6,13 @@ public static class Open
 {
   public static void Url(string url)
   {
-    var psi = new ProcessStartInfo
+    var psi = new ProcessStartInfo { FileName = url, UseShellExecute = true };
+    try
     {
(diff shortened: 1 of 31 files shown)


Core/Core/Helpers/Open.cs
```diff
@@ -0,0 +1,62 @@
...
+    try
+    {
+      Process.Start(psi);
+    }
+#pragma warning disable CA1031
+    catch
+#pragma warning restore CA1031
+    {
+      psi.UseShellExecute = false;
+      Process.Start(psi);
```
JR-Morgan, line 19: What this for?

adamhathcock: In case we can't shell execute for backwards compatibility and cross platformness.  The false for .NET Core was because not all platforms like it.  Like Mac/Linux.

PR #3592, opened 2024-07-25
Retry Fixes

JR-Morgan:
- Fixed reporting of retries: 
   - We were never incrementing the retry number given to the logger, so it was allways reporting as 0 even if it had retried.
   - Looks like a regression introduced by https://github.com/specklesystems/speckle-sharp/pull/2305 when we commented out some logging.
   - Also fixed the headers which previously were appending new correlation ids with every retry.
- Removed ancient flawed retries loop in ServerAPI
    - Clients already retry with proper backoff policy, this naive loop was only causes problems for the server.

### DIFF
--- a/Core/Core/Helpers/Http.cs
+++ b/Core/Core/Helpers/Http.cs
@@ -30,16 +30,10 @@ public static class Http
       .HandleTransientHttpError()
       .WaitAndRetryAsync(
         delay ?? DefaultDelay(),
-        (ex, timeSpan, retryAttempt, context) => {
-          //context.Remove("retryCount");
-          //context.Add("retryCount", retryAttempt);
-          //Log.Information(
-          //  ex.Exception,
-          //  "The http request failed with {exceptionType} exception retrying after {cooldown} milliseconds. This is retry attempt {retryAttempt}",
-          //  ex.GetType().Name,
-          //  timeSpan.TotalSeconds * 1000,
-          //  retryAttempt
-          //);
+        (ex, timeSpan, retryAttempt, context) =>
+        {
+          context.Remove("retryCount");
+          context.Add("retryCount", retryAttempt);
         }
       );
   }
@@ -209,11 +203,13 @@ public sealed class SpeckleHttpClientHandler : HttpClientHandler
       var timer = new Stopwatch();
       timer.Start();
       context.Add("retryCount", 0);
+
+      request.Headers.Add("x-request-id", context.CorrelationId.ToString());
+
       var policyResult = await Http.HttpAsyncPolicy(_delay)
         .ExecuteAndCaptureAsync(
           ctx =>
           {
-            request.Headers.Add("x-request-id", ctx.CorrelationId.ToString());
             return base.SendAsync(request, cancellationToken);
           },
           context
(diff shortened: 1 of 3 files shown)


Core/Core/Helpers/Http.cs
```diff
@@ -30,16 +30,10 @@ public static AsyncRetryPolicy<HttpResponseMessage> HttpAsyncPolicy(IEnumerable<
...
-          //  ex.Exception,
-          //  "The http request failed with {exceptionType} exception retrying after {cooldown} milliseconds. This is retry attempt {retryAttempt}",
-          //  ex.GetType().Name,
-          //  timeSpan.TotalSeconds * 1000,
-          //  retryAttempt
-          //);
+        (ex, timeSpan, retryAttempt, context) =>
+        {
+          context.Remove("retryCount");
+          context.Add("retryCount", retryAttempt);
```
BovineOx, line 36: this is to update?

Core/Core/Transports/ServerUtils/ServerAPI.cs
```diff
@@ -77,13 +75,9 @@ public async Task<string> DownloadSingleObject(string streamId, string objectId)
...
     };
 
-    HttpResponseMessage rootHttpResponse;
-    do
-    {
-      rootHttpResponse = await _client
-        .SendAsync(rootHttpMessage, HttpCompletionOption.ResponseContentRead, CancellationToken)
-        .ConfigureAwait(false);
-    } while (ShouldRetry(rootHttpResponse));
+    HttpResponseMessage rootHttpResponse = await _client
```
BovineOx, line 78: we don't call this directly? I think you were saying this is being retried at a higher level, are we 100% sure of this?

Core/Core/Transports/ServerUtils/ServerAPI.cs
```diff
@@ -335,13 +325,9 @@ CbObjectDownloaded onObjectCallback
...
     childrenHttpMessage.Headers.Add("Accept", "text/plain");
 
-    HttpResponseMessage childrenHttpResponse;
-    do
-    {
-      childrenHttpResponse = await _client
-        .SendAsync(childrenHttpMessage, HttpCompletionOption.ResponseHeadersRead, CancellationToken)
-        .ConfigureAwait(false);
-    } while (ShouldRetry(childrenHttpResponse));
+    HttpResponseMessage childrenHttpResponse = await _client
```
BovineOx, line 328: 1. Same question about retrying a a higher level here... and at the other places

BovineOx (approved):
Approved on basis of understanding in comments

Core/Core/Helpers/Http.cs
JR-Morgan: yep, set the new retry count to be the latest number of retry attempts

Core/Core/Transports/ServerUtils/ServerAPI.cs
JR-Morgan: it's being rertied at a lower level by the http client.
This old code here predates our polly implementation, and we just forgot to remove this nieve do while loop

JR-Morgan: not sure what you mean by "we don't call this directly"

PR #3593, opened 2024-07-25
bug(civil3d) : fixes pipe base curve conversion bc civil3d api sucks

clairekuang:
tl;dr: `basecurve` property on pipes is fake news and the line/arc needs to be retrieved directly from the pipe class

bonus: I believe we were misconverting arc pipes previously, and this pr fixes it.
Before:
image
After:
image

### DIFF
--- a/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
+++ b/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
@@ -1065,8 +1065,23 @@ public partial class ConverterAutocadCivil
   // TODO: add pressure fittings
   public Pipe PipeToSpeckle(CivilDB.Pipe pipe)
   {
-    ICurve curve = CurveToSpeckle(pipe.BaseCurve);
-
+    // get the pipe curve
+    // rant: if this is a straight or curved pipe, the BaseCurve prop is fake news && will return a DB.line with start and endpoints set to [0,0,0] & [0,0,1]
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
```diff
@@ -1065,8 +1065,23 @@ private List<CivilDataField> PartDataRecordToSpeckle(PressureNetworkPartData par
...
+    {
+      case PipeSubEntityType.Straight:
+        var line = new Acad.LineSegment3d(pipe.StartPoint, pipe.EndPoint);
+        curve = LineToSpeckle(line);
+        break;
+      case PipeSubEntityType.Curved:
+        curve = ArcToSpeckle(pipe.Curve2d);
+        break;
+      default:
+        curve = CurveToSpeckle(pipe.BaseCurve); // basecurve is fake news, but we're still sending the other types with props for now
```
BovineOx, line 1082: Should we send it?

BovineOx (approved):
Looks good, just a Q about whether we should send if default and also I would provide some specific instructions for anyone testing to rest these code paths fully

Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Civil.cs
clairekuang: I'm saying yes since previously we were: the meshes and other props on these pipes should still be correct, just not the basecurve

PR #3594, opened 2024-07-26
CNX-9912 - Implement a display fallback for Revit Areas equivalent to the outline.

jsdbroughton:
Implements `displayValue` to be the Area outline as there is no method to retrieve or generate a `Mesh`


JR-Morgan (changes requested):
Looks to be some build failures in other connectors (tekla) introduced by this PR. I'd assume there's a couple places where we're consuming this displayValue list directly.

Let me know if you need a hand @jsdbroughton

jsdbroughton:
> I'd assume there's a couple places where we're consuming this displayValue list directly.

Tekla ContourPlates were also inherited from Area - this could be where AdvanceSteel was inspired

JR-Morgan (approved):

PR #3595, opened 2024-07-26
QGIS attributes in Revit

KatKatKateryna:
duplicating attribute receive functionality from GisPolygon to GisFeature

### DIFF
--- a/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertPolygonElement.cs
+++ b/Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertPolygonElement.cs
@@ -11,6 +11,34 @@ namespace Objects.Converter.Revit;
 
 public partial class ConverterRevit
 {
+  public ApplicationObject GisFeatureToNative(GisFeature gisFeature)
+  {
+    // targeting Polygon features with List of Meshes as dispayValue
+    var speckleDirectShape = new Objects.BuiltElements.Revit.DirectShape()
+    {
+      applicationId = gisFeature.applicationId ??= Guid.NewGuid().ToString(),
+      baseGeometries = new List<Base>(),
+      parameters = new Base(),
+      name = "",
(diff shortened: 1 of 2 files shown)


Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertPolygonElement.cs
```diff
@@ -11,6 +11,34 @@ namespace Objects.Converter.Revit;
...
+      baseGeometries = new List<Base>(),
+      parameters = new Base(),
+      name = "",
+      category = RevitCategory.GenericModel
+    };
+
+    var traversal = new GraphTraversal(DefaultTraversal.DefaultRule);
+    var meshes = traversal.Traverse(gisFeature).Select(tc => tc.current).Where(b => b is Mesh);
+
+    speckleDirectShape.baseGeometries.AddRange(meshes);
```
BovineOx, line 29: Is it okay if meshes is empty? could meshes be empty?

Objects/Converters/ConverterRevit/ConverterRevitShared/PartialClasses/ConvertPolygonElement.cs
```diff
@@ -11,6 +11,34 @@ namespace Objects.Converter.Revit;
 
 public partial class ConverterRevit
 {
+  public ApplicationObject GisFeatureToNative(GisFeature gisFeature)
```
BovineOx, line 14: This is creating a tight coupling between Gis and Revit, are we happy for this? We did discuss that we need to be careful when picking support for these workflows. I guess we've done it for other things but might be worth discussing with @clairekuang as well.

BovineOx (approved):
It feels broadly okay but it does introduce tight coupling between GIS and Revit. I suggest you can @clairekuang agree this makes sense and socailise with @didimitrie also but happy to approve. If it's an a deeper issue then we can always back this out.

PR #3598, opened 2024-07-29
fix(civil3d): prop sets & changes name to definition name

clairekuang:
Fixes community reported bug on receiving property sets due to change of type to `Base`. 
sample commit: [https://app.speckle.systems/projects/b53a53697a/models/899753a050](https://app.speckle.systems/projects/b53a53697a/models/899753a050)
Also changes the name of property sets to use the definition name instead of the applied set name.
image


oguzhankoral (approved):
LGTM

PR #3599, opened 2024-07-29
Fix: Throw SpeckleException on unhandled elements

r-wessel:
Fix: Change handling of non-polyline openings to throw SpeckleException rather than nulling object refs (which throws an exception terminating the receive)


JR-Morgan (changes requested):
CI is failing because of formatting violations: https://app.circleci.com/pipelines/github/specklesystems/speckle-sharp/16133/workflows/83fb3eb9-0431-4634-ab54-3916cadb82ab/jobs/150626

Please run `dotnet csharpier .` in the command line, it should reformat the offending file

JR-Morgan (approved):

PR #3606, opened 2024-07-31
add null check around assembly & location

adamhathcock:
As found here: https://speckle.community/t/dllconflictmanager-throwing-error-in-revit/12589

### DIFF
--- a/ConnectorCore/DllConflictManagement/DllConflictManager.cs
+++ b/ConnectorCore/DllConflictManagement/DllConflictManager.cs
@@ -101,11 +101,16 @@ public sealed class DllConflictManager
     }
   }
 
-  private bool ShouldSkipCheckingConflictBecauseOfAssemblyLocation(Assembly loadedAssembly)
+  private bool ShouldSkipCheckingConflictBecauseOfAssemblyLocation(Assembly? loadedAssembly)
   {
+    if (string.IsNullOrWhiteSpace(loadedAssembly?.Location))
+    {
+      return false;
+    }
+    string location = loadedAssembly!.Location;
     foreach (var exactPath in _exactAssemblyPathsToIgnore)
     {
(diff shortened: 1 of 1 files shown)


ConnectorCore/DllConflictManagement/DllConflictManager.cs
```diff
@@ -101,19 +101,37 @@ HashSet<string> visitedAssemblies
...
+  {
+    if (string.IsNullOrWhiteSpace(loadedAssembly?.Location))
+    {
+      return false;
+    }
+    string location = loadedAssembly!.Location;
+    if (!PathIsValid(location))
+    {
+      return false;
+    }
```
jsdbroughton, line 123: Do we want to log either of these exits?

adamhathcock: I don't want to log anything to our servers that doesn't matter.  We're logging too much as it is.

BovineOx: If there's an unhandled exception it should surface - we're not expecting one I guess

BovineOx: Might we want to check for NotSupportedException? https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=net-8.0

BovineOx (changes requested):
I wonder if the NotSupportedException is a possibility here?

https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=net-8.0

ConnectorCore/DllConflictManagement/DllConflictManager.cs
adamhathcock: maybe but it's likely already handled somewhere and it's not the current problem this is solving

JR-Morgan (approved):

PR #3637, opened 2024-10-07
Added workspace Id tracking

JR-Morgan:
This PR adds `workspace_id` tracking for Receive and Send events.

---

I was going to add support for fetching the workspace id to all of the `ProjectResource` queries, like I've done for the v3 sdk https://github.com/specklesystems/speckle-sharp-sdk/pull/134

But, since its important that the v2 SDK is compatible with older self hosted servers, I thought it would be easiest simply to
add a helper function just for this query. This avoids having to add  a version check in each of the Project resources (which, at least right now in their current form don't have direct access to the underlying `GqlClient`.

---

Needs through testing:

- Test with app.speckle.systems:
    - [x]  project on a workspace, does it track send and receives in mixpanel
    - [x] project not on a workspace, should track `null`
  
- Test with testing1
    - [x] project not on a workspace, should track `null`, nothing should throw!

### DIFF
--- a/Core/Core/Api/GraphQL/Client.cs
+++ b/Core/Core/Api/GraphQL/Client.cs
@@ -17,6 +17,7 @@ using Serilog.Core;
 using Serilog.Core.Enrichers;
 using Serilog.Events;
 using Speckle.Core.Api.GraphQL;
+using Speckle.Core.Api.GraphQL.Models.Responses;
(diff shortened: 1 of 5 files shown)


didimitrie (approved):

didimitrie:
note: didn't test locally, just sanity check glanced at the code!

AlanRynne (approved):

PR #3647, opened 2024-10-30
Fix: Potential fix for godot issue

JR-Morgan:
see https://speckle.community/t/godot-parameter-path-is-empty-exception/14629/8

### DIFF
--- a/Core/Core/Logging/SpeckleLog.cs
+++ b/Core/Core/Logging/SpeckleLog.cs
@@ -266,12 +266,9 @@ public static class SpeckleLog
 
   private static FileVersionInfo? GetFileVersionInfo()
   {
-    // For some reason, Location will return null inside GoDot runtimes,
-    // Despite `Location` being decorated as non-nullable
-    string? assembly = Assembly.GetExecutingAssembly().Location;
+    string assembly = Assembly.GetExecutingAssembly().Location;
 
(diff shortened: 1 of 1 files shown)


Core/Core/Logging/SpeckleLog.cs
```diff
@@ -264,9 +264,18 @@ private static string GetUserIdFromDefaultAccount()
...
-  private static FileVersionInfo GetFileVersionInfo()
+  private static FileVersionInfo? GetFileVersionInfo()
   {
-    var assembly = Assembly.GetExecutingAssembly().Location;
+    // For some reason, Location will return null inside GoDot runtimes,
+    // Despite `Location` being decorated as non-nullable
+    string? assembly = Assembly.GetExecutingAssembly().Location;
+
+    // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
+    if (assembly == null)
```
Gytaco, line 274: Hi @JR-Morgan 

I tracked down the issue, the problem is the assembly isn't null its empty.
Using a Null Or Empty check correctly returns a null value.
if (string.IsNullOrEmpty(assembly))

JR-Morgan: ah yep ok, that makes more sense now. I've updated it to do a `IsNullOrEmpty`

Thanks a lot!

AlanRynne (approved):

adamhathcock (approved):

PR #3672, opened 2025-01-10
Fix ping in SDK and add user agent headers

adamhathcock:


didimitrie (approved):

Core/Core/Helpers/Http.cs
```diff
@@ -143,7 +143,7 @@ public static async Task<HttpResponseMessage> HttpPing(Uri uri)
     try
     {
       using var httpClient = GetHttpProxyClient();
-      HttpResponseMessage response = await httpClient.GetAsync(uri).ConfigureAwait(false);
+      HttpResponseMessage response = await httpClient.GetAsync(GetPingUrl(uri)).ConfigureAwait(false);
       response.EnsureSuccessStatusCode();
       SpeckleLog.Logger.Information("Successfully pinged {uri}", uri);
       return response;
```
JR-Morgan, line 149: This function is used by `IsFrontend2` to determine if the server supports FE2.
This PR breaks this, because `/favicon.ico` doesn't have the FE headers its expecting.

I've checked with @gjedlicska, and FE1 is now deprecated, we can safely assume now that all servers are FE2. (I.e. always set `Fe2 = true` on accounts, rather than fetch the headers.

JR-Morgan (changes requested):

JR-Morgan (approved):

PR #3673, opened 2025-01-13
Removed frontend headers check for FE1 servers

JR-Morgan:
#3672 is failing on integration tests since it changes, since we change the ping URL to  `/favicon.ico`, and this doesn't return the `x-speckle-frontend-2` header.

With approval from @gjedlicska, I'm now assuming that all accounts are FE2. Since FE1 is deprecated. This only thing that this affects is:
 - DUI2 UI terminology
 - StreamWrapper `ToString()` (i.e. we'll be sending people to FE2 formatted urls even if they're on FE1)

### DIFF
--- a/Core/Core/Api/GraphQL/Models/ServerInfo.cs
+++ b/Core/Core/Api/GraphQL/Models/ServerInfo.cs
@@ -13,10 +13,9 @@ public sealed class ServerInfo
 
   /// <remarks>
   /// This field is not returned from the GQL API,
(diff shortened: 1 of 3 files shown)


adamhathcock (approved):
This removes a common ping as well

Any more ping usage to remove?

gjedlicska (approved):
ok from my end

JR-Morgan:
None that need the headers.
The rest of pings are required to check the server is online/user has internet access.

PR #3679, opened 2025-01-21
fix: Disappearing Grasshopper components after save

AlanRynne:
SyncReceive would disappear upon opening a document (or while saving) due to exceptions being thrown in the Read/Write logic, which shouldn't ever throw.


JR-Morgan (approved):
lgtm

PR #3687, opened 2025-02-04
Jrm/branch get nullability

JR-Morgan:
Fixed regression with branch get not handling null return when branch does not exist


jsdbroughton (approved):
Perfect. Next time more tests 👯‍♀️

PR #3692, opened 2025-02-19
Back merge main -> dev

JR-Morgan:


JR-Morgan:
On second thoughts, I will kill dev, since we don't want a 2 channel speckle-sharp

Issue #3615, opened 2024-08-14
Revit Connector: Gutters not converted when sending "everything"


ks-cph:
## Prerequisites

- [x] I read the [contribution guidelines](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)
- [x] I checked the [documentation](https://speckle.guide/) and found no answer.
- [x] I checked [existing issues](../issues?q=is%3Aissue) and found no similar issue. 
- [x] I checked the [community forum](https://speckle.community/) for related discussions and found no answer.
- [x] I'm reporting the issue to the correct repository (see also [speckle-server](https://github.com/specklesystems/speckle-server), [speckle-sharp](https://github.com/specklesystems/speckle-sharp), [specklepy](https://github.com/specklesystems/specklepy), [speckle-docs](https://github.com/specklesystems/speckle-docs), and [others](https://github.com/orgs/specklesystems/repositories))

## What package are you referring to?
Backend

## Describe the bug
I am trying to send a model in Revit 2023 with roofs and gutters. When sending "everything", the gutters are not converted. However, when selecting them and sending "selection", they are converted correctly.

## To Reproduce

1. Create a Revit model with gutters and roofs
2. Click on Send everything
3. Open the stream on Speckle
4. See that the gutters are missing

## Expected behavior
We would expect the gutters also to be streamed when sending "everything"

## Proposed Solution

The error is somewhere in `ConnectorRevitUtils`:
https://github.com/specklesystems/speckle-sharp/blob/149d35e7600e09c13ea9ceaa33db15a441ff9104/ConnectorRevit/ConnectorRevit/ConnectorRevitUtils.cs#L70-L75

The category `OST_Gutter` has not been added. When manually adding OST_Gutter with the id `-2001391`, the Gutters are correctly converted even when sending "everything".

I do not understand what exactly is going on in `GetOrInitializeWithDefaultFactory`, so I can neither propose a solution.

ks-cph:
I noticed that the same behavior occurs for elements of the category `Soffit` and `Fascia`.

martinromby:
Any guidance to how we better can understand `GetOrInitializeWithDefaultFactory` and propose a solution here @bimgeek ?

clairekuang:
Do these element types not fall under the types of categories listed in this method? ->
https://github.com/specklesystems/speckle-sharp/blob/main/ConnectorRevit/RevitSharedResources/Helpers/Extensions.cs#L64

[...]

PR #3695, opened 2025-03-14
fix(revit) #3615: include gutters and other subcategories in Revit connector sending "everything"

jcesarpolo:
## Description & motivation
- Linked to: #3615 
- This PR fixes an issue where **gutters, roof soffits, fascias, wall sweeps and slab edges** were missing when using the "Export Everything" option in the Revit connector.
- The fix ensures that these subcategories are correctly included in the exported data.

## Changes:
- **Modified `IRevitDocumentCacheExtensions.cs`**:
  - Previously `CacheInitializer()` looped only through the categories in `Autodesk.Revit.DB.Document.Settings.Categories` which returns only the top-level categories, ommiting some supported model categories.
  - Updated `CacheInitializer()` method to correctly include the mentioned **subcategories**. 

## Screenshots:
**Model exported from Revit sending "everything"**
image

**Result on Speckle. No Gutters**
image

**Result on Speckle after the fix (Roof Soffits also tested)**
image

## Validation of changes:
- Tested on **all supported versions of Revit** to confirm the fix works across different versions.
- **Avoided looping through all built-in subcategories** to prevent unexpected results for unknown subcategories.
- Below is an example of artifacts that appeared when including all built-in subcategories supported:
image

## Checklist:

### DIFF
--- a/All.sln
+++ b/All.sln
@@ -427,6 +427,9 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ConnectorCore", "ConnectorCore", "{DA9DFC36-C53F-4B19-8911-BF7605230BA7}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ConnectorRhino8", "ConnectorRhino\ConnectorRhino8\ConnectorRhino8.csproj", "{D22A887D-976C-4DBF-AE5B-9039F169E61C}"
+	ProjectSection(ProjectDependencies) = postProject
+		{89996067-3233-410A-A6A1-39E2F11F0626} = {89996067-3233-410A-A6A1-39E2F11F0626}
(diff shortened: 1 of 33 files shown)


didimitrie:
Thank you @jcesarpolo, this looks simple enough to review and merge!

JR-Morgan (approved):
Tested in 2022 and 2025. lgtm

JR-Morgan:
yep, so this adds extra categories (see diff) to both the `Categories` filter and `Everything` filter (since they use the same internal cache.

This is good, I've checked it works correctly for both filters, and that it doesn't lead to duplicate elements, we simply were not sending categories like wall sweep and gutters from either the `Categories` or `Everything` filters before.

I just want to do a final check in Revit 2021, since there may be differences in categories behave in the older versions

JR-Morgan:
2021 looks ok

PR #3717, opened 2025-05-07
chore: Renamed command in Rhino and Autocad toolbars

JR-Morgan:
Rhino and AutoCAD toolbars also need updating to the changed command name

image

### DIFF
--- a/All.sln
+++ b/All.sln
@@ -100,6 +100,9 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Grasshopper", "Grasshopper", "{9461B162-AD2F-4F60-BCE4-E72DCEFD4608}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Rhino", "Rhino", "{E94E7327-5A9B-48EE-93CC-E9E9A5B980F1}"
+	ProjectSection(SolutionItems) = preProject
+		ConnectorRhino\ConnectorRhino\Toolbars\SpeckleConnectorRhino.rui = ConnectorRhino\ConnectorRhino\Toolbars\SpeckleConnectorRhino.rui
+	EndProjectSection
 EndProject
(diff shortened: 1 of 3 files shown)


ConnectorAutocadCivil/ConnectorAutocadCivil/Entry/App.cs
```diff
@@ -166,9 +173,30 @@ public void Create()
       "Check out our tutorials! Opens a page in your web browser"
     );
     RibbonToolTip docsTip = CreateToolTip("Docs", "Check out our documentation! Opens a page in your web browser");
-    RibbonButton community = CreateButton("Community", "SpeckleCommunity", null, helpButton, communityTip, "forum");
-    RibbonButton tutorials = CreateButton("Tutorials", "SpeckleTutorials", null, helpButton, tutorialsTip, "tutorials");
-    RibbonButton docs = CreateButton("Docs", "SpeckleDocs", null, helpButton, docsTip, "docs");
+    RibbonButton community = CreateButton(
```
clairekuang, line 176: Actually I would remove the tutorials and docs links and buttons entirely - we are moving towards deprecating all v2 support. And we don't have updated links for these in V3 yet

Can leave community for help requests

All.sln
```diff
@@ -100,6 +100,9 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Grasshopper", "Grasshopper", "{9461B162-AD2F-4F60-BCE4-E72DCEFD4608}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Rhino", "Rhino", "{E94E7327-5A9B-48EE-93CC-E9E9A5B980F1}"
+	ProjectSection(SolutionItems) = preProject
```
clairekuang, line 103: can you explain rq why this was added to the solution?

clairekuang (changes requested):

clairekuang (approved):

ConnectorAutocadCivil/ConnectorAutocadCivil/Entry/App.cs
JR-Morgan: I think we should wait until we know what's happening with v2 docs post v3 docs site

PR #3721, opened 2025-05-22
chore (tekla): legacy terminology

dogukankaratas:
Renames button and the form to legacy and updates the thumbnail.

image

PS: 2021 and 2020 versions are not supported anymore. It's not even possible to download them. Therefore they've been not updated.


ConnectorTeklaStructures/ConnectorTeklaStructuresShared/MainPlugin.cs
```diff
@@ -5,7 +5,7 @@ namespace Speckle.ConnectorTeklaStructures;
 
 public class StructuresData { }
 
-[Plugin("Speckle.ConnectorTeklaStructures")]
```
JR-Morgan, line 8: This looks like a namespace... are we sure this is a safe change?

JR-Morgan (approved):

PR #3723, opened 2025-05-29
feat(dui2): Use permission check for receive

JR-Morgan:
The logic in DUI2 to check the user can receive a model was implemented using a simple check for the user's project role.
This was implemented before the server implemented the permission checks into the graphql api.

This update replaces the main checks that the DUI2 connectors do with the server permission check, and should allow workspace admins to paste by url and receive.

However, non-explicit projects still won't appear in the DUI2 UI, since there is no include implicit role filter option for the `Streams` query like there is with `activeUser.projects`


clairekuang (approved):
Tested in Autocad 2023:

- [x] workspace editor (not admin) receiving via url to project that I'm not on : no access message
- [x] workspace admint receiving via url to project that I'm not on: correctly received

